PR #127, opened 2020-10-22
multiple gpu optuna

zzhaozhe-profolio:
A SleepTune function that allows us to train:

1. On single or multiple GPUs
2. Select available device ids
3. Easy set up for search space
4. Return a optuna.trial.FrozenTrial so that we could access the best parameters


zzhaozhe-profolio:
Hi @ismael-mendoza , I think it's a good time to merge the branch. 

I implement the function SleepTune in hyperparameter.py so that we can easily set up single or multiple gpu training. No need to go through set up backend or optuna.

bliss/hyperparameter.py
```diff
@@ -0,0 +1,315 @@
...
+    :param pruner: optuna.pruner
+    :param num_gpu: numer of GPUs to use. Could be an integer or a list of ids.
+                    CPU only: set this arugument to be zero
+                    Single GPU: either pass in 1 ot a list single id, e.g. [1]
+                    Multiple GPUs: either pass in an integer > 1 and SleepTune pick
+                                    available GPUS, or a list of ids.
+    :param storage: If using multiple GPUs, storage must be defined as optuna.RDBStorage
+
+    :return: an optuna.trial.FrozenTrial object
+    """
```
ismael-mendoza, line 227: Could you change the format of the docstring to the google format? Thanks! (there is an option in pycharm settings)

bliss/hyperparameter.py
```diff
@@ -0,0 +1,315 @@
...
+    batch_size,
+    direction: str = "minimize",
+    data_seed: int = 10,
+    n_trials=100,
+    time_out=600,
+    gc_after_trial: bool = False,
+    sampler=None,
+    pruner=None,
+    num_gpu: int or list = 2,
+    storage=None,
```
ismael-mendoza, line 203: since this class has many keyword arguments, it would be great if you created a config file for optuna/hyper-parameter tuning that contains these parameters. I'm also ok doing this in a future PR but I just thought I mentioned it as an idea.

bliss/hyperparameter.py
```diff
@@ -0,0 +1,315 @@
...
+        # check if available GPUs are enough
+        if len(deviceID) < num_gpu:
+            raise Warning("Available GPU is less than gpu number specified")
+        gpu_queue = mp.Manager().Queue()
+        for i in deviceID:
+            gpu_queue.put(i)
+    elif isinstance(num_gpu, list) and len(num_gpu) > 1:
+        gpu_queue = mp.Manager().Queue()
+        for i in num_gpu:
+            gpu_queue.put(i)
```
ismael-mendoza, line 249: what about the `else` case? maybe this is just the last case?

bin/optuna_multiple_gpu.py
```diff
@@ -0,0 +1,56 @@
...
+
+    cfg.optimizer.params.update({"lr": (1e-4, 1e-2), "weight_decay": (1e-6, 1e-4)})
+
+    SleepTune(
+        cfg,
+        max_epochs=100,
+        model_dir=model_dir,
+        monitor=monitor,
+        n_batches=4,
+        batch_size=32,
```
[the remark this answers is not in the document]
ismael-mendoza: remember you can change things from the command line with hydra like `python optuna_multiple_gpu dataset.params.n_batches=32`

bin/optuna_multiple_gpu.py
```diff
@@ -0,0 +1,56 @@
...
+        direction="minimize",
+        data_seed=10,
+        n_trials=100,
+        time_out=600,
+        gc_after_trial=True,
+        sampler=None,  # use the default sampler of the optuna
+        pruner=pruner,
+        num_gpu=n_gpu,
+        storage="sqlite:///zz.db",
```
jeff-regier, line 49: @ismael-mendoza should all these parameters be stored in hydra config files? i’m not sure how it is typically done

bin/optuna_multiple_gpu.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
```
jeff-regier, line 51: tunning->tuning

ismael-mendoza (changes requested):
Hi @zzhaozhe-profolio , thanks for doing this PR. Your hyper-parameter tuning script is shaping up to be very useful for bliss. I have a couple of suggestions to integrate it more smoothly, thanks in advance for your help with these.

bin/optuna_multiple_gpu.py
ismael-mendoza: Yeah that would be a good improvement, I was thinking we can have a config file for hyper-parameter tuning stuff. Not sure if @zzhaozhe-profolio wants to get into this in this PR but that would be nice

bliss/hyperparameter.py
```diff
@@ -0,0 +1,315 @@
...
+                timeout=time_out,
+                gc_after_trial=gc_after_trial,
+            )
+
+    # get the best trial
+    trial = study.best_trial
+
+    logger.info("Number of finished trials: {}".format(len(study.trials)))
+    logger.info(f"Best trial: {trial.number} with value {trial.value}")
+    print("  The params for the best trial: ")
```
jeff-regier, line 311: Should we use the logger here too?

tests/test_optuna.py
```diff
@@ -18,44 +18,61 @@ def on_validation_end(self, trainer, pl_module):
...
+
+                SleepTune(
+                    cfg,
+                    max_epochs=n_epochs,
+                    model_dir=os.getcwd(),
+                    monitor="val_loss",
+                    n_batches=4,
+                    batch_size=32,
+                    n_trials=1,
+                    num_gpu=0,
```
jeff-regier, line 77: The if block and the else block contain duplicated code. Can you merge them and just do `num_gpu=int(use_gpy)`? That will help with unit test coverage

bin/optuna_multiple_gpu.py
zzhaozhe-profolio: @ismael-mendoza sampler and pruner need to be constructed by using optuna object. It's not just a number or string. I'm not sure if the config file could do that?

tests/test_optuna.py
```diff
@@ -18,44 +18,61 @@ def on_validation_end(self, trainer, pl_module):
...
+                        "enc_kern": 3,
+                        "enc_hidden": 64,
+                    }
+                )
+                n_epochs = 1
+
+                SleepTune(
+                    cfg,
+                    max_epochs=n_epochs,
+                    model_dir=os.getcwd(),
```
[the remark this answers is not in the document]
zzhaozhe-profolio: Yes, it's where the folder is created. Will delete them.

bin/optuna_multiple_gpu.py
ismael-mendoza: I guess it depends how flexible you want the options to be. If you always want to use the median pruner you can just add the keyword arguments of the median pruner to the config file. If you want to allow for multiple types pruners then it gets more complicated. This might be helpful: https://hydra.cc/docs/patterns/instantiate_objects/overview

bliss/hyperparameter.py
ismael-mendoza: sounds good to me

PR #128, opened 2020-10-26
Im/improve testing1

ismael-mendoza:
## Summary

Per our discussion last week I will use an ensemble of images to test inside the unit tests instead of having a single one. 

## Tasks

- [x] Add hparams to tensorboard (finally!) 

- [x] simplify validation step in sleep phase (averaging is taking care by tensorboard)

- [x] Only one flag for plotting in sleep phase `self.plotting`

- [x] update with the correct __new__ way of adding logs (`self.log` inside lightning module, much cleaner)

- [x] unit tests now use functions inside the lightning module (no need to copy map_estimate boilerplate every time we want to test)


ismael-mendoza:
All tests based on ensemble averages of locations and counts works on galaxies and stars using images produced on the fly. Next PR will add data for tests that needs to be read from hard disk.

PR #138, opened 2020-11-11
Bl starnet case study

Runjing-Liu120:
The M2 test now seems to be working (in a jupyter notebook). I'll convert it to a formal unittest before I merge. 

The key issue that ended up being inconsistencies in loading the hubble catalog (the locations weren't scaled appropriately). To avoid this in the future, I now have scripts to download directly from source the raw Hubble and SDSS data (for example, see `./get_m2_data.sh` in the data folder). Before I was loading in a npz file with data saved from my old repo, and inconsistencies were really hard to track. 

Two other substantive changes: 
1) the decoder is now allowed to generate more stars per tile than the encoder can detect. In other words, the `max_sources` argument in the decoder can be larger than the `max_detections` argument in the encoder. The encoder, of course, will with probability 1 miss those extra sources. But this is now consistent with the pre-tiling code where locations were generated on the full image (instead of tiles) -- in that case, the prior used to have nonzero probability of a tile having more sources than the decoder can detect. 

2) There is an option to not pad the encoder background with a constant. Its easier to explain in picture. Suppose we have an 8x8 image (`slen = 8`), and the encoder requires 4x4 tiles (`tile_slen = 2`), with 2-pixels of padding around each tile (so `ptile_slen=8`). 

The image returned by the decoder: 
Screenshot from 2020-11-11 11-16-52

The tiles of the encoder: 
Screenshot from 2020-11-11 11-17-28

where black are padded values because some padded tiles go off the image. 

However, for real data, you can always give the encoder extra padding -- in this example, instead of feeding in an encoder an 8x8 image, I could really just give it a 12x12 image and let it return everything only in the central 8x8. 

So now, there is an argument in the decoder where you can set `border_padding = 2`, and the decoder now returns a 12x12 image: 
Screenshot from 2020-11-11 11-21-16

and you can set `pad_border_w_constant = False`, so tiles of the encoder become: 
Screenshot from 2020-11-11 11-21-49

I've set it so that the defaults are the same as before, so old code should not be broken. If you want to play around with this, you just have to make sure the tiles still line up in the decoder and the encoder: so `border_padding` should be set to be the same as the edge padding (`(ptile_slen - tile_slen) / 2`) in the encoder.

In particular, as a special case, this will allow @ismael-mendoza to run those one-tile galaxy experiments where the galaxy is constrained to be in the center of the image.


data/sdss/Makefile
```diff
@@ -13,17 +10,17 @@ FIELD4 := $(shell printf '%04d' $(FIELD_STRIPPED))
...
-all : $(SUBDIR3)/fpM-$(RUN6)-u$(CAMCOL)-$(FIELD4).fits \
-      $(SUBDIR3)/fpM-$(RUN6)-g$(CAMCOL)-$(FIELD4).fits \
-      $(SUBDIR3)/fpM-$(RUN6)-r$(CAMCOL)-$(FIELD4).fits \
-      $(SUBDIR3)/fpM-$(RUN6)-i$(CAMCOL)-$(FIELD4).fits \
-      $(SUBDIR3)/fpM-$(RUN6)-z$(CAMCOL)-$(FIELD4).fits \
+all : $(SUBDIR3)/fpM-$(RUN6)-u$(CAMCOL)-$(FIELD4).fit \
+      $(SUBDIR3)/fpM-$(RUN6)-g$(CAMCOL)-$(FIELD4).fit \
+      $(SUBDIR3)/fpM-$(RUN6)-r$(CAMCOL)-$(FIELD4).fit \
+      $(SUBDIR3)/fpM-$(RUN6)-i$(CAMCOL)-$(FIELD4).fit \
+      $(SUBDIR3)/fpM-$(RUN6)-z$(CAMCOL)-$(FIELD4).fit \
```
jeff-regier, line 17: I think @ismael-mendoza wanted to rename the `.fit` files to `.fits`

jeff-regier:
Great work! Really exciting to see this fixed.

ismael-mendoza:
Thanks @Runjing-Liu120 ! When you have a chance can you rebase with master? This will make for an easier review.

data/sdss/Makefile
ismael-mendoza: Yes that would be great

data/sdss/Makefile
```diff
@@ -1,7 +1,4 @@
 URLBASE := https://data.sdss.org/sas/dr12/boss
-RUN := 2583
-CAMCOL := 2
-FIELD := 136
```
jeff-regier, line 4: Is there a problem with these default settings?

config/model/m2.yaml
```diff
@@ -25,13 +25,15 @@ decoder:
...
-    psf_params_file: ${paths.root}/data/psField-002583-2-0136.fits
+    psf_params_file: ${paths.root}/data/sdss/2583/2/136/psField-002583-2-0136.fits
     background_values: ${model.image.background}
+    border_padding: 3
 encoder:
   params:
     n_bands: ${model.image.n_bands}
     tile_slen: ${model.decoder.params.tile_slen}
-    ptile_slen: 8
+    ptile_slen:  8
```
jeff-regier, line 35: Extra space?

case_studies/starnet/data/sdss/sample_rcfs.txt
```diff
@@ -0,0 +1,309 @@
+RUN=4153 CAMCOL=4 FIELD=100
```
jeff-regier, line 1: Please use the version of the Makefile in the data/sdss directory. It's not necessary to include all these RCFs after all. See `https://github.com/prob-ml/bliss/blob/master/data/sdss/fetch_sample_rcfs.sh`

bliss/datasets/sdss.py
```diff
@@ -140,3 +140,164 @@ def get_from_disk(self, idx):
...
+    # the full SDSS image, ~1500 x 2000 pixels
+    sdss_image_full = torch.Tensor(sdss_data[0]["image"])
+    sdss_background_full = torch.Tensor(sdss_data[0]["background"])
+
+    #####################
+    # load hubble catalog
+    #####################
+    hubble_cat_file = (
+        hubble_dir + "hlsp_acsggct_hst_acs-wfc_ngc7089_r.rdviq.cal.adj.zpt"
+    )
```
jeff-regier, line 197: Could put the stuff related to Hubble in it's own file, e.g. `datasets/hubble.py`?

bliss/models/decoder.py
```diff
@@ -145,8 +156,8 @@ def __init__(
         if ext == ".npy":
             psf_params = torch.from_numpy(np.load(psf_params_file)).to(device)
             psf_params = psf_params[list(range(n_bands))]
-        elif ext == ".fits":
-            assert n_bands == 2, "only 2 band fits files are supported."
+        elif (ext == ".fits") or (ext == ".fit"):
```
jeff-regier, line 159: Better just to support ".fits"

bliss/models/decoder.py
```diff
@@ -707,9 +724,11 @@ def _construct_full_image_from_ptiles(image_ptiles, tile_slen):
...
         # trim to original image size
+        x0 = ptile_padding * tile_slen - border_padding
+        x1 = (n_tiles1 + ptile_padding) * tile_slen + border_padding
         return canvas[
             :,
             :,
-            (ptile_padding * tile_slen) : ((n_tiles1 + ptile_padding) * tile_slen),
-            (ptile_padding * tile_slen) : ((n_tiles1 + ptile_padding) * tile_slen),
+            x0:x1,
+            x0:x1,
```
jeff-regier, line 733: Can we fit this all on one line now?

case_studies/starnet/m2_utils/m2_data.py
```diff
@@ -0,0 +1,288 @@
...
+
+        return ret
+
+
+def convert_mag_to_nmgy(mag):
+    return 10 ** ((22.5 - mag) / 2.5)
+
+
+def convert_nmgy_to_mag(nmgy):
+    return 22.5 - 2.5 * torch.log10(nmgy)
```
jeff-regier, line 145: Better not to duplicate these functions

Runjing-Liu120:
OK! It looks like all the tests are up and running. There's still a few hacky things to take care of, namely: 

1) I gave up on the .fit vs .fits distinction, and just saved what we needed for the m2 test as a npz file ... I'd still like at some point to have a clear pointer to where the data comes from and how it was processed, but I can't seem to understand how the Makefile works and why it seems to work with .fit files and not .fits files. 

2) I put the functions for Hubble comparisons in the sdss.py file, though down the line I should implement more generic versions of these and add them as metrics for the sleep phase. 

But, if it is OK, we can go ahead and merge this, and open new pull requests to take care of these changes.

jeff-regier:
@ismael-mendoza Please merge this one you've review it

case_studies/starnet/m2_results.ipynb
```diff
@@ -0,0 +1,1005 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
```
jeff-regier, line 5: Do we want to strip the output of the case studies notebooks? I'm not sure. All the output makes for large diffs and can make the repository large. On the other hand, it's nice for people to see the results without having to run it themselves.

jeff-regier (approved):

PR #139, opened 2020-11-11
Im/kl qp sleep encoder

ismael-mendoza:
# Assumptions

- Size of padded tiles and tiles is shared between ImageEncoder and GalaxyEncoder. (could potentially relax if useful)
- Every tile has at most one galaxy in it. 

# Summary of changes
- Removed flags of `conftest.py` that are now deprecated (`--runslow`, `--repeat`)
- decoder `render_images` now returns galaxy variance images 
- correct `super()` syntax
- `use_galaxy_encoder: False` flag in `sleep.py` and config files whether to turn on optimization of new galaxy encoder. 
- `encoder.ImageEncoder` has been simplify and now doesn't do any `galaxy_param` optimization or calculations. (removed `n_galaxy_params`  keyword)
- pinned `jedi= "0.17.2"` since update breaks Ipython completion, see: https://github.com/ipython/ipython/issues/12740. Updated `pyproject.toml`
- `test_tune.py` had repeated keys in overrides dict. 
- simplified `test_kl_pq.py`/ `test_encoder.forward.py` to account for exclusion of galaxy_params in encoder. 
- added `optimizer_indx` to `sleep.py` to decide which optimization you are doing in sleep phase. 
- Uses `galaxy_net.CenteredGalaxyEncoder` as the galaxy_encoder in `sleep.py`
- divided the loss in `sleep.py` into `get_detection_loss` and `get_galaxy_loss`. Simplifying `get_detection_loss` since it does not have to do calculations on galaxy_params.
- added `map_estimate` to `SleepPhase` that will use both encoders to calculate a 'complete' MAP estimate. 
- simplified `encoder.py` in several ways. 
- `sample_encoder` now works with multiple images


ismael-mendoza:
Closing this PR soon since it's getting big, next PR will center each tile on the corresponding galaxy and add evaluation of the galaxy images.

PR #140, opened 2020-11-13
Im/simpler dataset

ismael-mendoza:


bliss/sleep.py
```diff
@@ -508,15 +505,15 @@ def _get_params_logprob_all_combs(true_params, param_mean, param_logvar):
 class SleepObjective(object):
```
zzhaozhe-profolio, line 505: You can delete this. Ray-tune seems to work pretty good. Probably not gonna use Optuna anymore

PR #142, opened 2020-11-18
fix in galaxy scaling

Runjing-Liu120:
I believe this will address the problem with galaxies looking different depending on the size of the ptile padding. It was the same issues with stars -- which I fixed earlier -- but I guess that fix didn't go through to galaxies. 

For future reference, the "rule of thumb" to keep things scale-invariant to image size is to always pass in the a source (either a psf or a rendered galaxy) to `._render_one_source` that is the same size as the padded tile (or whatever you're rendering the source on). If the original PSF or galaxy is too large, we should trim it; if it is too small, we pad it with zeros. 

Before merging, maybe @ismael-mendoza can checkout this branch and see if it indeed fixes those example images that you showed before? And of course double check if the rendering of galaxies look reasonable to you. 

Also, one of the galaxy sleep tests fail on this branch (I believe it is the assert on line 60 of test_sleep_galaxy). We get an accuracy on counts of 0.72, while the threshold is 0.75. Is this a major problem?

### DIFF
--- a/bliss/models/decoder.py
+++ b/bliss/models/decoder.py
@@ -519,8 +519,7 @@ class ImageDecoder(nn.Module):
         # forward only galaxies that are on!
         gal_on, _ = self.galaxy_decoder.forward(z[b == 1])
 
-        # size the galaxy (either trims or crops to the
-        # size of ptile)
+        # size the galaxy (either trims or crops to the size of ptile)
         gal_on = self._size_galaxy(gal_on)
 
         # set galaxies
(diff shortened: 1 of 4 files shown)


ismael-mendoza (approved):

ismael-mendoza:
Hey @Runjing-Liu120 , thank you so much for this change, I produced some images in galaxies where you can clearly see your code is cropping and doing the right thing. Here are some examples: 

**(ptile_padding=3, tile_slen=4)**
Screen Shot 2020-11-19 at 17 23 19


**(ptile_padding=5, tile_slen=4)**
Screen Shot 2020-11-19 at 17 25 16

As you can see galaxies appear the same size but appear more cropped in the first case.

ismael-mendoza:
> For future reference, the "rule of thumb" to keep things scale-invariant to image size is to always pass in the a source (either a psf or a rendered galaxy) to ._render_one_source that is the same size as the padded tile (or whatever you're rendering the source on). If the original PSF or galaxy is too large, we should trim it; if it is too small, we pad it with zeros.

That makes sense, I'll keep this in mind going forward

PR #149, opened 2020-11-20
Tuning for m2

zzhaozhe-profolio:
Use this PR to run the tuning process for M2. 

Search Space:
**enc_conv_c**: tune.grid_search([10, 15, 20, 25]),
**enc_kern**: tune.grid_search([3]),
**enc_hidden**: tune.grid_search([64, 128, 192, 256]),
**lr**: tune.loguniform(1e-5, 1e-2),
**weight_decay:** tune.loguniform(1e-6, 1e-3)

Config: m2

Number of epochs: 200


zzhaozhe-profolio:
Current best space is  
```
Best config: {'enc_conv_c': 25, 'enc_kern': 3, 'enc_hidden': 256, 'lr': 0.0023728538449271077, 'weight_decay': 4.8380692551037007e-05}
```

bin/SleepRayTune.py
```diff
@@ -57,20 +57,20 @@ def SleepRayTune(config, cfg: DictConfig, num_epochs, num_gpu):
 
 @hydra.main(config_path="../config", config_name="config")
 # model=m2, dataset=m2, training=m2 in terminal
-def main(cfg: DictConfig, num_epochs=100, gpus_per_trial=1):
+def main(cfg: DictConfig, num_epochs=200, gpus_per_trial=1):
```
jeff-regier, line 60: Would you name this file using just lower case letters an underscores? That's the convention

jeff-regier:
> Current best space is
> 
> ```
> Best config: {'enc_conv_c': 25, 'enc_kern': 3, 'enc_hidden': 256, 'lr': 0.0023728538449271077, 'weight_decay': 4.8380692551037007e-05}
> ```

Really interesting! Would you report the true positive rate (TPR) and positive predictive value (PPV) with the best settings? There's code for computing those two summary statistics here:
https://github.com/prob-ml/bliss/blob/849e5a1ffa1ce4cd0ea0ac44acbd926c6764f1c9/bliss/datasets/sdss.py#L176-L186

bin/SleepRayTune.py
```diff
@@ -57,20 +57,20 @@ def SleepRayTune(config, cfg: DictConfig, num_epochs, num_gpu):
...
 
     logger = logging.getLogger()
 
     # restrict the number for cuda
-    os.environ["CUDA_VISIBLE_DEVICES"] = "1,4,3"
+    os.environ["CUDA_VISIBLE_DEVICES"] = "1,4,3,2"
```
jeff-regier, line 65: It's better to not to hard code the values of environment variables. They can be set in the shell that you use to launch the script, or in your bashrc file

bin/SleepRayTune.py
jeff-regier: Can `num_epochs` be set in the config file?

bin/SleepRayTune.py
```diff
@@ -57,20 +57,20 @@ def SleepRayTune(config, cfg: DictConfig, num_epochs, num_gpu):
...
     config = {
         "enc_conv_c": tune.grid_search([10, 15, 20, 25]),
-        "enc_kern": tune.grid_search([3, 5]),
-        "enc_hidden": tune.grid_search([64, 128]),
-        "lr": tune.loguniform([1e-5, 1e-2]),
-        "weight_decay": tune.loguniform([1e-6, 1e-3]),
+        "enc_kern": tune.grid_search([3]),
+        "enc_hidden": tune.grid_search([64, 128, 192, 256]),
+        "lr": tune.loguniform(1e-5, 1e-2),
+        "weight_decay": tune.loguniform(1e-6, 1e-3),
```
jeff-regier, line 73: @ismael-mendoza Do you think these ranges should be set in the config files? Or would that be overly complicated?

ismael-mendoza: I think this can be added to a config file, since our config files accept lists this should be pretty straightforward to incorporate. @zzhaozhe-profolio let me know if you have questions about how to do this

bin/SleepRayTune.py
ismael-mendoza: I would agree with this , also can you name it `n_epochs` for consistency. Thanks!

bin/SleepRayTune.py
```diff
@@ -57,20 +57,20 @@ def SleepRayTune(config, cfg: DictConfig, num_epochs, num_gpu):
...
     # define the parameter space
     config = {
         "enc_conv_c": tune.grid_search([10, 15, 20, 25]),
-        "enc_kern": tune.grid_search([3, 5]),
-        "enc_hidden": tune.grid_search([64, 128]),
-        "lr": tune.loguniform([1e-5, 1e-2]),
-        "weight_decay": tune.loguniform([1e-6, 1e-3]),
+        "enc_kern": tune.grid_search([3]),
+        "enc_hidden": tune.grid_search([64, 128, 192, 256]),
+        "lr": tune.loguniform(1e-5, 1e-2),
```
jeff-regier, line 72: When I try to run the script, I get an error message from the line above: 
```
TypeError: loguniform() missing 1 required positional argument: 'upper'
```
Are we using different versions of Ray tune?

zzhaozhe-profolio: Did you run it on master branch?

I think I fixed it in my branch. Make sure it's `tune.loguniform(1e-5, 1e-2)` but not `tune.loguniform([1e-5, 1e-2])` in the master

jeff-regier: Ah that must be it. thanks

bin/tuning.py
```diff
@@ -56,28 +54,39 @@ def SleepRayTune(config, cfg: DictConfig, num_epochs, num_gpu):
 
 
 @hydra.main(config_path="../config", config_name="config")
-# model=m2, dataset=m2, training=m2 in terminal
-def main(cfg: DictConfig, num_epochs=100, gpus_per_trial=1):
+# model=tuning, dataset=m2, training=tuning optimizer=m2 in terminal
```
ZachariahPang, line 57: For now, this script has to be run with `python tuning.py model=tuning dataset=m2 training=tuning optimizer=m2`. Would it be better to override all or some of them in the script, since they're mostly default settings? So that it can be run with `python tuning.py model=some_model_to_tune` 
@zzhaozhe-profolio

config/model/tuning.yaml
```diff
@@ -0,0 +1,41 @@
...
+  params:
+    n_bands: ${model.image.n_bands}
+    tile_slen: ${model.decoder.params.tile_slen}
+    ptile_slen: 8
+    max_detections: 2
+    n_galaxy_params: ${model.galaxy.latent_dim}
+    enc_conv_c: 20,
+    enc_kern: 3,
+    enc_hidden: 256
+
```
ZachariahPang, line 41: This whole config file is very similar to the default config `/config/model/m2.yaml` and since these three lines are already managed by Raytune. Should we just use the default config in `bin/tuning.py`?

ZachariahPang:
Are we expecting `/bin/tuning.py` to be used with several different architectures with different parameter spaces at the same time, or is this only going to be used with the latest architecture? 

I think this is especially important for designing our new config layout for tuning. I imagine this can be done in at least two ways.
1. Include relevant search space for each model, dataset, optimizer in its own config file. For example, the `adam.yaml` would include a range for learning rate and weight decay. Add another stand-alone tuning config file that includes generic settings such num_gpu_for_tuning, max_epoch_to_search. This might be a good approach when used with several different architectures.
2. Have only one config file for tuning with all the settings included. Whenever there is an architecture change, we update the config content in parallel with the possible search space accepted by our model. This might be easier if we only need `/bin/tuning.py` to work with the latest architecture at hand. This might be easier to manage as our model evolves since I suppose part of the architecture design is based on the previous tuning run.

bin/SleepRayTune.py
ZachariahPang: Would it be better to do this with something like
```yaml
enc_hidden:
 - 64
 - 256
 - 2
```
to be called as `tune.grid_search(list(range(*enc_hidden)))`
instead of listing every value in config file.

Or maybe more generic as
```yaml
enc_hidden: "tune.grid_search(...)"
```
to be called as `eval(enc_hidden)`

jeff-regier:
> Are we expecting `/bin/tuning.py` to be used with several different architectures with different parameter spaces at the same time, or is this only going to be used with the latest architecture?
> 
> I think this is especially important for designing our new config layout for tuning. I imagine this can be done in at least two ways.
> 
> 1. Include relevant search space for each model, dataset, optimizer in its own config file. For example, the `adam.yaml` would include a range for learning rate and weight decay. Add another stand-alone tuning config file that includes generic settings such num_gpu_for_tuning, max_epoch_to_search. This might be a good approach when used with several different architectures.
> 2. Have only one config file for tuning with all the settings included. Whenever there is an architecture change, we update the config content in parallel with the possible search space accepted by our model. This might be easier if we only need `/bin/tuning.py` to work with the latest architecture at hand. This might be easier to manage as our model evolves since I suppose part of the architecture design is based on the previous tuning run.

Good question, I hadn't thought about that. If it's simpler to suppose there's just one architecture, I'd start with that, and make it more complicated later if it turns out to be necessary

bin/SleepRayTune.py
jeff-regier: I suppose it depends how many settings of enc_hidden we want to test. I'd suggest starting with the simpler way (ie listing the values) and make it more automated if it seems like that will be helpful

config/model/tuning.yaml
ismael-mendoza: yeah, is there a reason that you created this config file instead of using the m2 one? @zzhaozhe-profolio

bin/tuning.py
ismael-mendoza: I think it's fine to keep as is since we might want to tune other models that are not `m2` later on.

bin/tuning.py
```diff
@@ -47,37 +39,54 @@ def SleepRayTune(config, cfg: DictConfig, num_epochs, num_gpu):
...
     )
     trainer.fit(model, datamodule=dataset)
 
 
 @hydra.main(config_path="../config", config_name="config")
-# model=m2, dataset=m2, training=m2 in terminal
-def main(cfg: DictConfig, num_epochs=100, gpus_per_trial=1):
+# model=tuning, dataset=m2, training=tuning optimizer=m2 in terminal
+# TODO: Maybe expose the config for search space outside of main either as argument or with Hydra 
+def main(cfg: DictConfig, num_epochs=200, gpus_per_trial=1):
```
ismael-mendoza, line 59: it would be nice if these two parameters were part of the config as well. I think you are discussing this already in the comments above.

ismael-mendoza (commented):
I just took a quick look to catch up on where you guys are with this and added a couple of small comments.

config/model/tuning.yaml
ZachariahPang: I discussed this with @zzhaozhe-profolio yesterday over zoom. This will be cleaned out shortly.

bin/tuning.py
ZachariahPang: We are working on collating the above two parameters, the total number of GPU for tuning, and the search space in a single config file.

ismael-mendoza: excellent, thanks!

config/tuning/m2.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - 320
+  lr:
+    - 1e-5
+    - 1e-1
+  weight_decay:
+    - 1e-6
+    - 1e-2
+max_epochs : 200
+gpus_per_trial: 1
+allocated_gpus: "1, 2, 3, 4"
```
ZachariahPang, line 25: Is there any special reason we only want to use these 4 specific gpus for tuning?
@zzhaozhe-profolio

zzhaozhe-profolio: no. you can assign any gpus to tune.

bin/tuning.py
```diff
@@ -0,0 +1,127 @@
...
+from ray.tune.suggest import ConcurrencyLimiter
+from ray.tune import CLIReporter
+from ray.tune.schedulers import ASHAScheduler
+from ray.tune.integration.pytorch_lightning import TuneReportCallback
+
+from bliss import sleep
+from bliss.datasets.simulated import SimulatedDataset
+
+
+def SleepRayTune(search_space, cfg: DictConfig):
```
jeff-regier, line 21: function names should be in lower case (with underscores)

jeff-regier:
PR looks great, please merge when ready

ZachariahPang:
> PR looks great, please merge when ready

Other than #164, I have finished most of what I had planned. Ready to merge on my end.

@zzhaozhe-profolio

Issue #135, opened 2020-11-11
More flexibility in "global constrains" on images from decoder


ismael-mendoza:
Sometimes I want to train the encoder on images of galaxies/stars that are not close to the edges of the image or where in each image there is a limit on the total number of objects. Currently there is no way to do this and it's a bit clunky to have to think of `max/mean/min_sources` over tiles instead of the whole image. It would be really nice to have a way of imposing this 'global' constrains on the images from the decoder. 

I really like that the decoder currently outputs the parameters over tiles. But is there a way to alleviate this other issue while maintaining the new structure? Not sure if it's even feasible.

ismael-mendoza:
I think that #138 solved the issue of having things to close to border via `border_padding` - @Runjing-Liu120 do you have plans to make the number of sources global too ? I remember you said you had an easy way of doing this. Thanks!

PR #151, opened 2020-11-23
Im/remove constant padding

ismael-mendoza:
Small PR to remove constant padding and always assume padding is set correctly by the decoder. By default border padding in decoder now matches border padding in encoder - please let me know @Runjing-Liu120 if you have any suggestions! I will really especially appreciate you review of **sleep.py/encoder.py/decoder.py**. Thanks! (you can ignore the other diffs)

I also added `slen` as metadata for the batch (this is ultimately necessary to get `locs` to be consistent)

**Note:** Everything is passing in the GPU. I removed one test that will be easier to do once #135 is closed.


bliss/models/decoder.py
```diff
@@ -82,17 +83,21 @@ def __init__(
...
-        # the number of pixels that pads the full image
-        self.border_padding = border_padding
+        # The width of the padding is given by ptile_slen.
+        # border_padding is the amount of padding we leave in the final image. Useful for
+        # avoiding sources getting too close to the edges.
+        if border_padding is None:
+            # default value matches encoder default.
+            border_padding = (ptile_slen - tile_slen) / 2
+
+        ptile_padding = (ptile_slen / tile_slen - 1) / 2
```
Runjing-Liu120, line 93: This might be annoying (and I know I'm probably the one that started this variable name ...), but is it possible to change this variable name? I'm not sure what it should be, but something to designate that this quantity is in units of tiles, not pixels. 

Maybe ``n_tiles_of_padding`` or something. 

Or, could we even get rid of this quantity?

Edit: after a quick search, we don't use this anywhere else do we? If so, this is fine then

bliss/models/decoder.py
```diff
@@ -82,17 +83,21 @@ def __init__(
...
+        assert border_padding % 1 == 0, "amount of border padding must be an integer"
+        assert ptile_padding % 1 == 0, "ptile_padding must be an integer."
         assert border_padding <= (
             ptile_padding * tile_slen
         ), "Too much border padding. Make ptile_padding larger. "
```
Runjing-Liu120, line 98: The print statement should now say "Make ptile_slen larger"

bliss/models/encoder.py
```diff
@@ -177,61 +175,45 @@ def __init__(
...
-        self.background_pad_value = background_pad_value
 
         # padding
         self.tile_slen = tile_slen
         self.ptile_slen = ptile_slen
-        self.edge_padding = (ptile_slen - tile_slen) / 2
-        assert self.edge_padding % 1 == 0, "amount of padding should be an integer"
-        self.edge_padding = int(self.edge_padding)
-        self.pad_border_w_constant = pad_border_w_constant
+        border_padding = (ptile_slen - tile_slen) / 2
```
Runjing-Liu120, line 185: thanks! this variable name is now consistent :) 

should have thought of this before...

bliss/models/encoder.py
```diff
@@ -595,12 +572,11 @@ def tiled_map_estimate(self, image):
...
-        if not self.pad_border_w_constant:
-            slen = slen - 2 * self.edge_padding
-
-        tile_estimate = self.tiled_map_estimate(image)
+    def map_estimate(self, slen, images):
+        # slen is size of the image without border padding
+        border_padding = (images.shape[-1] - slen) / 2
+        assert slen % self.tile_slen == 0, "incompatible image"
+        assert border_padding == self.border_padding, "incompatible border"
+        tile_estimate = self.tiled_map_estimate(images)
```
Runjing-Liu120, line 580: Do you have to pass in the slen? Could we set

```
slen = images.shape[-1] - 2 * self.border_padding
```
and keep only the assert on line 578 (and remove the one on 579). 

I **think** everything that follows should go through.

Runjing-Liu120 (approved):

bliss/models/decoder.py
ismael-mendoza: Ah yes good point, I changed the name just for clarity but it's not stored anywhere so it doesn't really matter.

bliss/models/decoder.py
ismael-mendoza: Thanks!

bliss/models/encoder.py
ismael-mendoza: Yes, in theory we could do that like we discussed. But then I realized that if I use this same encoder for other images that were not the decoder ones, I would like to know that the border padding is consistent. Otherwise the locs prediction will be off because the slen is wrong

ismael-mendoza:
Thanks for the helpful review @Runjing-Liu120 !

PR #152, opened 2020-11-28
Im/fix catsim1

ismael-mendoza:


ismael-mendoza:
All tests running in GPU so merging, the new galaxy VAE train workflow is much simpler + rendering/saving CATSIM images is simpler now (although still pretty slow to do it on the fly - future PR can investigate this).

PR #154, opened 2020-11-29
install poetry

jeff-regier:
This PR adds a `pyproject.toml` file, which makes bliss compatible with [poetry](https://python-poetry.org/). These are the main advantages I see:
- gets us all using the same versions of dependencies (recorded in `poetry.lock`)
- makes it possible to checkout old commits of bliss and run them with the old dependencies (because `poetry.lock` is versioned too)
- makes it (relatively) straightforward for new developers to install all the dependencies
- makes it (relatively) easy to distribute `bliss` through PyPi/`pip`.
- replaces `setup.py`, `MANIFEST.in`, and `requirements.txt` with a single config file

To get the bliss unit tests running with the changes in this PR, these steps worked for me:
1. Install Python3.8 if it isn't already installed. (It's already installed system-wide on deeplearning-01 now.) Probably better not to use the conda because that adds another layer of commands.
1. Install Poetry like this:
    > `curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -`
1. Add poetry to your PATH (the poetry installer may prompt you to do this)
1. From within the bliss repo, type `poetry install`. This installs all the bliss dependencies.
1. Type `poetry shell` to get access to the environment that has the dependencies satisfied.

I think that's it. Lmk if I'm missing steps.


jeff-regier:
I'm not sure why "build 3.7" is hanging, may be something do with the tested python version on master (3.7) and the version in this branch (3.8). Anyway, I think it's fine to merge.

ismael-mendoza:
Thanks Jeff! I will try it myself and merge after I get it to work

ismael-mendoza:
Quick question @jeff-regier , do we still need the `tox.ini` file and should we change the `py37` to `py38`?

jeff-regier:
I think tox is largely unrelated to poetry. There are some guides out there for using them together anyway. I'm not sure if we still want tox, but perhaps better to address that in another pr at a later time

jeff-regier:
I see, though we may as well change py37 to py38

jeff-regier:
OK I'll just remove tox since tox was expecting too find a requirements.txt file. I wonder why tox functionality isn't just built in to poetry. There's this plugin we could use someday:
https://pypi.org/project/tox-poetry/

ismael-mendoza:
@jeff-regier I just installed everything on python3.8 and that seems to work, for some reason when I run the tests I see the dataloader warning again

```
  /Users/Ismael/Library/Caches/pypoetry/virtualenvs/bliss-VKA6tDND-py3.8/lib/python3.8/site-packages/pytorch_lightning/utilities/distributed.py:45: UserWarning: The dataloader, test dataloader 0, does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` (try 4 which is the number of cpus on this machine) in the `DataLoader` init to improve performance.
    warnings.warn(*args, **kwargs)
```

do you know why it could be coming back now?

jeff-regier:
I've been seeing those warnings all along. Did you PR just get rid of the warnings when bliss is run from the bin directory?

ismael-mendoza:
When I tried it last time after merging that PR, it got rid of the warning in every case (both when running tests and when using the bin directory) so I find this strange.

ismael-mendoza:
I added this line to `simulated.py` : 
```
warnings.filterwarnings(
    "ignore", ".*does not have many workers which may be a bottleneck.*", UserWarning
)
```
which I figured should work everytime because this error is caused by the dataloaders in `simulated.py`

ismael-mendoza:
Could it have to do with the python version? ( I was using python 3.7 before)

ismael-mendoza:
Surprisingly, I don't see this warning in the CI. Instead I see 

```
../../../.cache/pypoetry/virtualenvs/bliss-UE3m6-Ul-py3.8/lib/python3.8/site-packages/torch/cuda/__init__.py:52
  /home/runner/.cache/pypoetry/virtualenvs/bliss-UE3m6-Ul-py3.8/lib/python3.8/site-packages/torch/cuda/__init__.py:52: UserWarning: CUDA initialization: Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx (Triggered internally at  /pytorch/c10/cuda/CUDAFunctions.cpp:100.)
    return torch._C._cuda_getDeviceCount() > 
```
do you know why?

jeff-regier:
> Surprisingly, I don't see this warning in the CI. Instead in the CI I see
> 
> ```
> ../../../.cache/pypoetry/virtualenvs/bliss-UE3m6-Ul-py3.8/lib/python3.8/site-packages/torch/cuda/__init__.py:52
>   /home/runner/.cache/pypoetry/virtualenvs/bliss-UE3m6-Ul-py3.8/lib/python3.8/site-packages/torch/cuda/__init__.py:52: UserWarning: CUDA initialization: Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx (Triggered internally at  /pytorch/c10/cuda/CUDAFunctions.cpp:100.)
>     return torch._C._cuda_getDeviceCount() > 
> ```
> 
> do you know why?

I wonder if poetry is installing the version of pytorch with cuda support by default. Hmmm....not sure how to fix that.

jeff-regier:
Looks like the cuda warning will be resolved soon: https://github.com/pytorch/pytorch/issues/47038

The way we’re having poetry install pytorch appears to match how others do it:
https://github.com/ludvb/xfuse/blob/master/pyproject.toml

ismael-mendoza:
excellent! thanks for looking into that! Do you have any idea why the `warnings.filterwarnings` does not work inside the poetry environment? Or do you have other suggestions for suppressing warnings? Otherwise I think we can merge this

jeff-regier:
I'm not sure about the warnings. I was seeing them all along, before and after poetry. Should the call to filter be somewhere else? Is any of this discussion relevant?
https://github.com/PyTorchLightning/pytorch-lightning/issues/3332

ismael-mendoza:
@jeff-regier I added a `pytest.ini` with the filter warnings enabled. Could you check if you still see warnings and if not merge this PR? Thanks!

ismael-mendoza (approved):

jeff-regier:
That worked! Thanks for adding the pytest.ini file.

jeff-regier:
@Runjing-Liu120 @zzhaozhe-profolio Please see above for instructions on using poetry.

PR #157, opened 2020-12-03
added ray and pre-commit to poetry dev dependencies

jeff-regier:
As discussed


ismael-mendoza:
Looks good to me, thanks @jeff-regier

PR #162, opened 2020-12-30
Misc Changes

ismael-mendoza:
Some small errors I noticed while working on the EECS 545 project last semester but never got pushed until now. 

@jeff-regier , quick question - should the `poetry.lock` file be committed? I figured that as long as everyone has the correct `pyproject.toml` that should be sufficient (since if this file changes then the `.lock` file will be updated automatically after running `poetry lock` or `poetry update`). What do you think?


jeff-regier:
I'm not sure what the standard practice is.
If we commit `poetry.lock` to version control, then, say, six months from now, we'll be able to checkout a version of bliss from today and run it easily, even though many dependencies will have made breaking changes by then.
But there are probably some drawbacks too, e.g. large diffs, having to run `poetry install` every time some else changes poetry.lock.

jeff-regier:
After reading a bit about it, I think committing `poetry.lock` is better than not. But we should try not to update it all the time without reason. A good reason would be that we have a new dependency to add. Or we need a newer version. Or that it's been awhile and many dependencies are better now.

ismael-mendoza:
I see, in this case it seems two dependencies were updated to newer versions but both don't seem important (although not sure). In this case should we not commit then?

jeff-regier:
Either way seems fine

ismael-mendoza:
all tests pass and gpu and cpu now

PR #163, opened 2021-01-01
Move common config value to sleep_base.yaml

ZachariahPang:
Since all `/config/model/sleep_*.yaml` are very similar, instead of having some values duplicated in all config files, I think it might be better to set them in `sleep_base.yaml` and re-use common values whenever possible. 

For now, this is done by force loading `sleep_base.yaml` before other more specific configs. Duplicated values in `sleep_base.yaml` are overwritten. 

This would introduce some unnecessary values when `model=galaxy_net`. This is fine for now as there are no duplicated keys. In the next release of Hydra, there will be more support for inheriting base config and this can be done more properly. 

For now, settings in `sleep_base.yaml` are just common settings shared by `sleep_*.yaml`. It might make more sense to split these based on model design.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,5 +1,6 @@
 defaults:
   - optimizer: adam
+  - model/sleep_base # force-load common config 
   - model: sleep_star_basic
   - training: default
   - dataset: default
--- a/config/model/m2.yaml
+++ b/config/model/m2.yaml
@@ -1,37 +1,20 @@
 # @package _group_
-name: SleepPhase
 image:
   slen: 100
   n_bands: 2
   background:
     - 686.0
     - 1123.0
-galaxy:
-  slen: 51
-  latent_dim: 8
 decoder:
   params:
-    n_bands: ${model.image.n_bands}
-    slen: ${model.image.slen}
-    tile_slen: 2
     ptile_slen: 26
     border_padding: 3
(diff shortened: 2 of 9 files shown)


ZachariahPang:
After some consideration, now I think this probably doesn't add much value as this would make it hard to read when looking at each config file. Have to remember what has been set in the base config. I'll close this PR without merging if you think the same.

ismael-mendoza:
Hi @ZachariahPang thanks for suggesting this PR and for looking into new hydra features that might be useful (there are certainly some that I haven't explored). I agree with you that there is a good deal of redundancy between the sleep config files, but I actually appreciate the redundancy as it helps me remember the parameters that are being used for each model without having to go back to the keyword arguments for each class. (this is a problem i was facing before I introduced the config files) So I think we should keep the redundancy in these config files for now. 

Just to give you some background, part of the reason for the large number of redundant config files is that I wanted to be very careful about the different parts of the codes that are being tested and separating those, e.g. `sleep_1galaxy_1star.yaml` tests that our code can correctly distinguish stars from galaxies and `sleep_star_one_tile.yaml` checks that our code can find/measure stars in a single tile as opposed to multiple tiles. As the code matures we can probably consolidate some of these config files I think. 

If you have any other suggestions though feel free to let me know and I'm happy to discuss!

PR #166, opened 2021-01-04
Zp/poetry packaging

### LINKED ISSUE #165: Use tool.poetry.scripts rather than having a bin directory
Rather than having our executable scripts (e.g., `tuning.py`) in the bin directory, it'd be preferable to put them within the `bliss` directory, where they can tested by pytest. Poetry has a way to add them to the path:
https://python-poetry.org/docs/pyproject/#scripts

This way, the current working directory does not need to be `bliss/bin` in order for the scripts to work. We may avoid a lot of tedious path issues by doing it this way.

ZachariahPang:
Close #165 

- Remove `bliss/bin` directory
- Move executables to `bliss/bliss`
- Instead of cd to the correct directory to run executables, now we can run
`poetry install` only for once to update the scripts in `bliss/bliss`
and run scripts from anywhere using `poetry run`
- For now, only `tune.py`,  `generate.py`, `train.py` are managed

eg. 
```shell
poetry run generate
```
or
```shell
poetry run tune model=m2 dataset=m2 training=m2 optimizer=m2 tuning.n_epochs=20
```


ZachariahPang:
I renamed `tuning.py` to tune.py to be consistent with the other executable names, `train.py` and `generate.py`, but this seems to also remove @zzhaozhe-profolio's editing history on that file.  Is there any work around?

jeff-regier:
`tune.py` is a good name. I'm not sure about the editing history; it looks to me like git recorded it as name change rather than a delete+add.

ZachariahPang:
> `tune.py` is a good name. I'm not sure about the editing history; it looks to me like git recorded it as name change rather than a delete+add.

It only shows me in git blame. It seems the problem is that I accidentally split the deleting and adding into two commits.
I'll force push a new history to this PR. Sorry about the inconvenience.

jeff-regier (changes requested):
Looks great! Would you add simple unit tests that runs the code in `tune.py`? No need to check the output--just verify it runs. You might set n_epoch=1 so it finishes quickly even on a CPU.

jeff-regier (approved):
Lgtm!

tests/test_tune.py
```diff
@@ -0,0 +1,31 @@
...
+class TestTune:
+    @pytest.fixture(scope="class")
+    def overrides(self, devices):
+        overrides = {
+            "model": "m2",
+            "dataset": "m2" if devices.use_cuda else "cpu",
+            "dataset.params.n_batches": 2 if devices.use_cuda else 10,
+            "training": "m2" if devices.use_cuda else "cpu",
+            "optimizer": "m2",
+            "tuning.n_epochs": 2 if devices.use_cuda else 1,
```
zzhaozhe-profolio, line 15: I think both gpu and cpu only run 1 epoch would be fine

zzhaozhe-profolio (approved):

Issue #164, opened 2021-01-04
Sleep phase weird metric behavior


ZachariahPang:
## Metric values behave weirdly but loss seems to be normal on m2
<img width="1400" alt="Screen Shot 2021-01-03 at 18 53 44" src="https://user-images.githubusercontent.com/46454028/103496989-0324c600-4df5-11eb-8a8b-0cab75da597f.png">

The metric `val_acc_counts`(shown as star_count_acc) is almost always 0.
The metric `val_gal_counts`(shown as galaxy_counts_acc) is the constant 0.
The metric `val_locs_median_mse`(shown as locs_median_mse) is almost always 0.5.
The metric `val_fluxes_avg_err`(shown as fluxes_avg_err) is almost inf.

**To reproduce**
```shell
git checkout a58ffde
cd bin
python tuning.py model=m2 dataset=m2 training=m2 optimizer=m2
```
General logs are printed in the terminal, detailed logs are available in Tensorboard
```shell
tensorboard --logdir=outputs/hyperparameter_tuning
```
## Metric behave normally with default 

<img width="1405" alt="Screen Shot 2021-01-03 at 4 58 46 PM" src="https://user-images.githubusercontent.com/46454028/103497434-5f3c1a00-4df6-11eb-9e27-a8c07c1223bf.png">
This image is from @zzhaozhe-profolio.
 
**To reproduce**
```shell
git checkout a58ffde
cd bin
python tuning.py tuning=m2
```
```shell
tensorboard --logdir=outputs/hyperparameter_tuning
```

This issue seems to be more related to modeling itself rather than the raytune.
Any suggestions?

zzhaozhe-profolio:
Is this because the source counts for each batch dose not always match (see the code [here](https://github.com/prob-ml/bliss/blob/master/bliss/sleep.py#L377) and below)? Seems that if the counts does not match, `locs_mse_vec` and `fluxes_mse_vec` won't be appended. Two metrics: `locs_median_mse` and `fluxes_avg_err` will always be default.

I checked if `true_n_sources_i ` matched with `true_n_sources_i`, and they failed the condition every time. @ismael-mendoza @Runjing-Liu120 Is this expected? 

```
        locs_mse_vec = []
        fluxes_mse_vec = []
        for i in range(batch_size):
            true_n_sources_i = true_params["n_sources"][i]
            n_sources_i = estimates["n_sources"][i]

            # only compare locations for mse if counts match.
[...]

PR #175, opened 2021-01-12
Bl metrics library

Runjing-Liu120:
Hi everyone, 

I've now added a `metrics.py` file containing some functions to evaluate metrics. 

In the `get_metrics` method of `SleepPhase`, we now return the TPR and PPV. I've removed those functions from `sdss.py`.  

In addition, we now evaluate the location/flux error slightly differently than before. Before, we were returning the average error in location/fluxes **only if** the number of estimated sources equals the number of true sources. That way, we could find a one-to-one and onto matching between the estimated and true sources; then for each match, we compare locations/fluxes. 

This was throwing errors on M2 because we never **exactly** get the true number of stars correctly (there are ~1000 on the image; we might estimate, for example 1012 stars instead of 1023 stars. Close, but not exactly right). 

Now, the evaluation of location/flux error can handle the situation where the true number and estimated number of sources differ. We do an "inner join" (in the spirit of the Dplyr operation) on the two lists of sources. We find the minimizing permutation to minimize the error in locations, but allow some sources to not have a match (so if there are more true stars than estimated stars, some true stars won't have a match; if there are more estimated stars than true stars, some estimated stars won't have a match). The matching is done with the Hungarian algorithm in Scipy. We then only evaluate the location/flux error on sources that were matched.


ZachariahPang (approved):

bliss/metrics.py
```diff
@@ -0,0 +1,164 @@
...
+            est_mag = torch.log10(est_fluxes) * 2.5
+
+            # get TPR and PPV
+            tpr_i, ppv_i = get_tpr_ppv(true_locs, true_mag, est_locs, est_mag)
+
+            tpr_vec[i] = tpr_i
+            ppv_vec[i] = ppv_i
+
+            # get l1 error in locations and fluxes
+            locs_mae_i, fluxes_mae_i = get_l1_error(
```
ismael-mendoza, line 146: `fluxes_mae` could be a bit misleading since it's really magnitudes right? Maybe change the name to `mag_mae`?

ismael-mendoza (approved):
Looks really good @Runjing-Liu120 , thank you! I just had 1 small comment about a variable's name

ismael-mendoza:
It would be excellent if @ZachariahPang  or @zzhaozhe-profolio re-ran their tests on m2 with this branch to see if #164 was resolved with Bryan's changes.

ZachariahPang:
> It would be excellent if @ZachariahPang or @zzhaozhe-profolio re-ran their tests on m2 with this branch to see if #164 was resolved with Bryan's new metrics library (or in a separate PR, whatever is easier)

I run a quick `tune.py` and have the following result.
Screen Shot 2021-01-12 at 11 35 08 
I think the new metrics work well. We probably can close #164 upon merging this.

Something I noticed though, is that `tune.py` has a much less GPU load, around 30% most of the time comparing to high 90% previously. I think this makes sense since the Hungarian algorithm is on CPU which bottlenecks the tuning process. The `tune.py` running time is significantly longer. A quick fix for this might be returning metrics once a few epochs instead of every epoch. I can make those changes if you think that might be beneficial.

PR #176, opened 2021-01-12
Finalize galaxy measurement on one tile

### LINKED ISSUE #123: learn single-tile one-galaxy parameters

ismael-mendoza:
Closes #123 

## Remaining issues after this PR 
- training galaxy encoder is quite slow (too much padding and too many tiles) 
- using 4x4 tiles makes detection suffer slightly (but 2x2 would make problem above worst)
- it seems from examples below that flux is misestimated sometimes, it seems to be in the cases of more extreme blends though? 
- current unittest is a dummy, havent' figured out good image ensemble metrics to assess performance. Next PR should focus on this 
- not sure what is best way to ensure consistency of MAP estimate when galaxy_encoder is turned off, but this is not too important at the moment. 
- only works with max 1 galaxy per tile.


Runjing-Liu120 (approved):

ismael-mendoza:
Hi @jeff-regier , I got some encouraging results on galaxy measurement/deblending. I attached some examples to this post, as you can see it's completely perfect but I think it's a step in the right direction (the most drastic residuals are the ones where we missed one detection). There are some cases where flux is overestimated slightly but it's a lot better than with our previous approach. You will also see detection suffers a bit from the 4x4 vs 2x2 tiles (I think)
epoch199

Should I think of a small test and merge at this point?

ismael-mendoza:
@jeff-regier - Should I think of a small test and merge at this point?

jeff-regier:
Exciting! It works! Great job.

Yes, adding a small test and merging makes sense to me.

ismael-mendoza:
Almost done with this PR, @ZachariahPang it seems like coverage decreased in `tune.py` , are you able to spot why?

jeff-regier:
> * using 4x4 tiles makes detection suffer slightly (but 2x2 would make problem above worst)

I'm hopeful a proper attention mechanism will let us work with even larger tiles.

> * it seems from examples below that flux is misestimated sometimes, it seems to be in the cases of more extreme blends though?

Could this due to #126?

> * current unittest is a dummy, havent' figured out good image ensemble metrics to assess performance. Next PR should focus on this

Cataloging a full ~2 megapixel SDSS frame, containing both stars and galaxies, would be a good next step too

jeff-regier (approved):
It's really exciting to see this working!

ZachariahPang:
> @ZachariahPang it seems like coverage decreased in `tune.py` , are you able to spot why?
This was my fault all the way back to #149. I wasn't paying much attention on Codecov before. 
I think the problem is with how raytune manages its different workers. The whole `trainable()` was not correctly reported even though it's covered. It would be an easy fix. Do you want me to make a commit to the PR?

jeff-regier:
> I think the problem is with how raytune manages its different workers. The whole `trainable()` was not correctly reported even though it's covered.

Not sure if this is related, but I've noticed the tuning tests fail when there only one GPU available (e.g. my home computer) or when the CUDA_VISIBLE_DEVICES environment variable is set to only show one GPU (e.g. on deeplearning-01).  Is there a way to test the tuning code without spawning new processes? (Would that solve all these issues at once?)

ismael-mendoza:
@ZachariahPang if you think it's a small/easy change please commit to this PR, but not sure if that is still true following Jeff's comment

ismael-mendoza:
> Could this due to #126?

sorry I just the word 'fluxes' a bit ambiguously, the galaxy flux estimation is completely separate from star flux estimation

ZachariahPang:
> > I think the problem is with how raytune manages its different workers. The whole `trainable()` was not correctly reported even though it's covered.
> 
> Not sure if this is related, but I've noticed the tuning tests fail when there is only one GPU available (e.g. my home computer) or when the CUDA_VISIBLE_DEVICES environment variable is set to only show only one GPU (e.g. on deeplearning-01). Is there a way to test the tuning code without spawning new processes? (Would that solve all these issues at once?)

I'm not sure if it's related. The solution for Codecov is to add `ray.init(local_mode=True)` so that only one process is running. The reason testing failed for only one-gpu is that the testing is written as using two gpus. I mainly wanted to make sure the testing covers the more general multi-gpu tuning case, but that can be changed easily or add felxibity if needed.

ismael-mendoza:
Test are working on the CPU and GPU (stably, I think) so merging. Galaxy measurement still works quite well (but no direct unittest for this yet). 

# Summary of changes

- Incorporated galaxy measurement when at most 1 galaxy per tile (use config_file: `sleep_galaxy_measure_basic.yaml`)
- pytests can be run in the CPU even when GPU is available via `pytest --gpu ''` 
- added some potential metrics for galaxy/image reconstruction but need to be examined in a future PR
- centering and cropping padded tiles is the `encoder.py` job
- improve stability of `test_galaxy_detection.py`

PR #178, opened 2021-01-15
bliss command line interface

### LINKED ISSUE #169: "bliss" command line interface (and unit tests)
Bliss current installs three executable scripts: train, tune, and generate. I think it'd be better to combine all three of these scripts into a single script called "bliss". The bliss script would take a command as it's first argument: train, tune, or generate. The advantage of this approach is that it bundles together all the different bliss commands, and let's us easily add more.

It would also be good to add basic unit tests for the "train" and "generate" modes, as we're missing them currently.

ZachariahPang:
close #169 

This currently has some problems with the unit test when `test_run.py` is tested on CPU on our server when GPU are available. The problem is mainly from the generated batch which would be created on GPU in `decoder.py`. I would need more time or some help to handle this properly.

This maybe less of an issue if the unit test on github runs fine when no GPU is available.


jeff-regier:
Looks good to me. @ismael-mendoza any idea why the decoder might generate batches on the GPU when the rest of the model is in main memory?

ZachariahPang:
After some more tracing, I think the problem might be https://github.com/prob-ml/bliss/blob/32f1c72fbfca05dc565cbedb873641576d27cc5a/bliss/__init__.py#L1-L4 and `device` is explicitly called in `decoder.py` to move/construct tensors.

Reading config file and determining `device` in `__init__.py` might be the most direct fix but I'm not sure if this would be a good idea.

ismael-mendoza:
Thanks @ZachariahPang for pointing this out, I actually think that this is exactly the problem. The other day I tried to modify the unit tests so that we could run them in the CPU while GPU where available but because of how the `device` is determined in `__init__.py`  I couldn't find a good way to do this. 

> Reading config file and determining device in __init__.py might be the most direct fix but I'm not sure if this would be a good idea.

Hmm yeah that could work but it's not completely clear to me how to set it up because then wouldn't you need to run the `__init__.py` from the command line and have a main() function? The other way to fix this (which is more annoying) is to create the device from the cfg inside the relevant classes and pass it around but I feel that there must be an easier way to make a global configurable `device` available in the code

jeff-regier:
@ZachariahPang I'm glad you identified this issue. Are either of the possible solutions better suited for distributed training? If so, that one seems better; @zzhaozhe-profolio was planning to attempt some multiple-GPU runs. The guide [here](https://towardsdatascience.com/trivial-multi-node-training-with-pytorch-lightning-ff75dfb809bd) shows how to use pytorch-lightning on multiple GPUs without any changes to the code. It'd be great if we could do that. Define device in the `__init__.py` file seems like it might prevent us from running on an arbitrary number of GPUs without changes to the code.

ismael-mendoza:
> Are either of the possible solutions better suited for distributed training? If so, that one seems better

I'm actually not sure if any of these options are suitable for distributed training (I was only thinking of 1 GPU), if @ZachariahPang or @zzhaozhe-profolio could investigate this further and look at the guide that would be great. (I can also do that once I finish the galaxy issue if this should be a priority)

> Define device in the __init__.py file seems like it might prevent us from running on an arbitrary number of GPUs without changes to the code.

Yeah I  think the `__init__` thing is not good practice in general but was an easy thing to do at the time.

ZachariahPang:
> I'm actually not sure if any of these options are suitable for distributed training (I was only thinking of 1 GPU), if @ZachariahPang or @zzhaozhe-profolio could investigate this further and look at the guide that would be great. 

Yes, I can work on that. Comparing to trying other architectures for the encoder and this issue, which would you recommend me to work on first?

> @zzhaozhe-profolio was planning to attempt some multiple-GPU runs.

Are we planning to do single-node or multiple-node? I remember you mentioned we would have a cluster later on but I'm sure if I recall that correctly.
 
Since this issue would be a bit more involved than a few lines of change, should I merge this and address #179  in another PR?

ZachariahPang:
Naming the entry point file as `bliss.py` in `bliss/bliss` would create some problems with things like `from bliss import ...` especially during debugging other files. Would it make sense to rename `bliss/bliss/bliss.py` to something like `main.py` or `bliss_interface.py`? We can still keep the cli as `bliss train ...` though.

jeff-regier:
> Yes, I can work on that. Comparing to trying other architectures for the encoder and this issue, which would you recommend me to work on first?

It'd be really good to the device consistency issues resolved, so I'd recommend that first

> Are we planning to do single-node or multiple-node? I remember you mentioned we would have a cluster later on but I'm sure if I recall that correctly.

Multi-node eventually, but hopefully we don't have to do anything special for multi-node (rather than single-node, multiple GPU) at this point.

> Would it make sense to rename bliss/bliss/bliss.py to something like main.py or bliss_interface.py? We can still keep the cli as bliss train ... though.

Good point, `main.py` or `cli.py` both seem better than `bliss.py`

jeff-regier:
> Since this issue would be a bit more involved than a few lines of change, should I merge this and address #179 in another PR?

Yes, that's preferable

PR #180, opened 2021-01-18
Remove `cuda` or `to` call

### LINKED ISSUE #179: Delete .cuda() or .to() calls
There are a number of calls to `.to()` in decoder.py, e.g.,
https://github.com/prob-ml/bliss/blob/68657fb8f417c75c41b24f4a5f806b67368c3936/bliss/models/decoder.py#L149

These may be problematic, at least for multi-GPU training, but perhaps just for getting everything on the right device. See
https://pytorch-lightning.readthedocs.io/en/stable/multi_gpu.html#delete-cuda-or-to-calls

ZachariahPang:
close #179 

The `device` and `use_cuda` in `__init__.py` are completely removed to avoid dependencies on global `device` variable that is not managed by pytorch-lightning or config. To achieve this, the following changes are made (the following changes are tested both on gpu and cpu on our server.). 
- For tensors initialized in `__init__()`, they are now managed by pytorch itself to move together with the whole model using `register_buffer()`.
- For tensors created after `__init__()` in a downstream pass, their `device` is inferred from `self.device` which is updated whenever the model gets moved to a new device by pytorch-lightning.
- For tensors created for operating with "input", their `device` is inferred from the "input" tensor.
- [For tensors registered as parameters](https://github.com/prob-ml/bliss/blob/1802f0aefdc44d0bb482f1575f328954b4ddfe4c/bliss/models/decoder.py#L158-L170), they would automatically be managed by pytorch along with other parameters, so we don't need to manually assign device.
- [For loaded submodules](https://github.com/prob-ml/bliss/blob/1802f0aefdc44d0bb482f1575f328954b4ddfe4c/bliss/models/decoder.py#L151-L155),  they would automatically be managed by pytorch along with other submodules, so we don't need to manually assign device.

One implementation detail is that `encoder.indx_mats` was previously created as a dictionary of tensors in `__init__()`. Pytorch doesn't support register a dictionary buffer, so tensors in `indx_mats` are now stored directly under `encoder` with their keys in `indx_mats` as attribute names. Those can be accessed, for example, by `self.loc_mean` or `getatter(self, "loc_mean")` depending on use case.

For our current training scheme. There are two `decoder` objects created, `SimulatedDataset.image_decoder` and `SleepPhase.image_decoder`. 
`SleepPhase.image_decoder` is managed by pytorch-lightning since it's passed along with `SleepPhase` in pytorch `Trainer` and pl should correctly manage it for multi-gpu since there's no hard-coded `device` outside.
`SimulatedDataset.image_decoder` is the one that we have to manually control since pl doesn't really have support for preparing data on gpu and `SimlulatedDataset.image_decoder` is not part of the training model.

There are two solutions in preparing for multi-gpu.
1. (implemented in this PR) Manually assign a device for dataloader for hosting `SleepPhase.image_decoder`. It would be initialized on cpu just as normal and then move to the assigned device. This operation itself should be fine as `nn.module` are normally moved like this and this only happens once. `SleepPhase.image_decoder` is automatically managed by pl and should run without problems. So now our dataloader is on a `generate_gpu` and maybe different from the gpu devices hosting models. This itself and the transferring data should also be fine since normally this is done with dataloader on `cpu` and models on gpus. Comparing to the next resolution, this requires less code change. But I'm not sure in DDP, how pl would manage multiple processes and dataloaders (now use gpu as well).
2. [Move data generation to model forward pass](https://github.com/prob-ml/bliss/issues/179#issuecomment-761842518). This should work smoothly with DDP and would require less footprint since we can potentially use only one `decoder` instance for both training and data generation. We just need to be careful with our computational graph. We would also need to create a dummy dataset and I'm not sure how much impact this would create for other things. 

Sorry for the late writeup, some of the information above might be relevant to your comments. @jeff-regier @ismael-mendoza 

**New update**: For a sanity check, I set `generate_device: "cuda:7"`. It also works. This way, the model is on `cuda:0` and the data generation is on `cuda:7`. So the communication and the way the device being setup should at least doesn't cause any error, whether it's efficient or not. I think solution 1 might worth a try. I'm not 100% sure, but I think, for DDP, pl would create several processes and dataloaders on `generate_device` and several models on all other free gpus. So the setup up would be something like `generate_device` hosting all dataloaders and each other free gpu has one model for training. Giving that loading data on gpu uses around 1100mb memory and only around 3% workload. I think 1 gpu dedicated to feeding data is enough for the amount of gpus we have on our server for now. But solution 2 is still the better long term approach in my opinion. 

**New update2**: This PR works out of the box with the default `ddp_spawn` DDP scheme using all gpus (one of the gpus has all dataloaders and one model, other gpus each has one model) for training.

### DIFF
--- a/bliss/datasets/simulated.py
+++ b/bliss/datasets/simulated.py
@@ -20,8 +20,8 @@ class SimulatedDataset(pl.LightningDataModule, IterableDataset):
 
         self.n_batches = cfg.dataset.params.n_batches
         self.batch_size = cfg.dataset.params.batch_size
-        self.image_decoder = ImageDecoder(
-            generate_device = cfg.dataset.params.generate_device, **cfg.model.decoder.params
+        self.image_decoder = ImageDecoder(**cfg.model.decoder.params).to(
+            cfg.dataset.params.generate_device
         )
         self.image_decoder.requires_grad_(False)  # freeze decoder weights.
 
--- a/bliss/datasets/simulated.py
+++ b/bliss/datasets/simulated.py
@@ -20,7 +20,9 @@ class SimulatedDataset(pl.LightningDataModule, IterableDataset):
 
         self.n_batches = cfg.dataset.params.n_batches
         self.batch_size = cfg.dataset.params.batch_size
-        self.image_decoder = ImageDecoder(**cfg.model.decoder.params)
+        self.image_decoder = ImageDecoder(
+            generate_device = cfg.dataset.params.generate_device, **cfg.model.decoder.params
+        )
         self.image_decoder.requires_grad_(False)  # freeze decoder weights.
 
         # check sleep training will work.
--- a/bliss/main.py
+++ b/bliss/main.py
@@ -12,3 +12,6 @@ def main(cfg):
     else:
         raise KeyError
     task(cfg)
+
+if __name__ == "__main__":
+    main()
--- a/bliss/main.py
+++ b/bliss/main.py
@@ -13,5 +13,6 @@ def main(cfg):
         raise KeyError
     task(cfg)
 
+
 if __name__ == "__main__":
     main()
--- a/bliss/models/decoder.py
+++ b/bliss/models/decoder.py
@@ -8,7 +8,6 @@ import torch.nn as nn
 import torch.nn.functional as F
 from torch.distributions import Poisson, Normal
 
-from .. import device
 from .encoder import get_is_on_from_n_sources
 from . import galaxy_net
 import pytorch_lightning as pl
@@ -39,8 +38,14 @@ class ImageDecoder(pl.LightningModule):
         background_values=(686.0, 1123.0),
         loc_min=0.0,
         loc_max=1.0,
+        generate_device=None
     ):
         super(ImageDecoder, self).__init__()
+        
+        if generate_device is None:
+            self.generate_device = self.device
+        else:
+            self.generate_device = generate_device
 
         # side-length in pixels of an image (image is assumed to be square)
         assert slen % 1 == 0, "slen must be an integer."
@@ -100,7 +105,7 @@ class ImageDecoder(pl.LightningModule):
         self.cached_grid = self.get_mgrid(self.ptile_slen)
 
         # misc
-        self.swap = torch.tensor([1, 0], device=device)
+        self.swap = torch.tensor([1, 0], device=self.generate_device)
 
         # background
         assert len(background_values) == n_bands
@@ -109,7 +114,7 @@ class ImageDecoder(pl.LightningModule):
             self.slen + 2 * self.border_padding,
             self.slen + 2 * self.border_padding,
         )
-        self.background = torch.zeros(background_shape, device=device)
+        self.background = torch.zeros(background_shape, device=self.generate_device)
         for i in range(n_bands):
             self.background[i] = background_values[i]
 
@@ -120,20 +125,20 @@ class ImageDecoder(pl.LightningModule):
         assert self.prob_galaxy > 0.0 or decoder_file is None
         if decoder_file is not None:
             dec = galaxy_net.CenteredGalaxyDecoder(gal_slen, n_galaxy_params, n_bands)
-            dec = dec.to(device)
-            dec.load_state_dict(torch.load(decoder_file, map_location=device))
+            dec = dec.to(self.generate_device)
+            dec.load_state_dict(torch.load(decoder_file, map_location=self.generate_device))
             dec.eval().requires_grad_(False)
             self.galaxy_decoder = dec
 
         # load psf params + grid
         ext = Path(psf_params_file).suffix
         if ext == ".npy":
-            psf_params = torch.from_numpy(np.load(psf_params_file)).to(device)
+            psf_params = torch.from_numpy(np.load(psf_params_file)).to(self.generate_device)
             psf_params = psf_params[list(range(n_bands))]
         elif ext == ".fits":
             assert n_bands == 2, "only 2 band fit files are supported."
             bands = (2, 3)
-            psf_params = self.get_fit_file_psf_params(psf_params_file, bands).to(device)
+            psf_params = self.get_fit_file_psf_params(psf_params_file, bands).to(self.generate_device)
         else:
             raise NotImplementedError(
                 "Only .npy and .fits extensions are supported for PSF params files."
@@ -168,7 +173,7 @@ class ImageDecoder(pl.LightningModule):
         # mgrid is between -1 and 1
         # then scale slightly because of the way f.grid_sample
         # parameterizes the edges: (0, 0) is center of edge pixel
-        return mgrid.float().to(device) * (slen - 1) / slen
+        return mgrid.float().to(self.generate_device) * (slen - 1) / slen
 
 
     def get_fit_file_psf_params(self, psf_fit_file, bands=(2, 3)):
@@ -187,10 +192,10 @@ class ImageDecoder(pl.LightningModule):
 
             psf_params[i] = torch.log(torch.tensor([sigma1, sigma2, sigmap, beta, b, p0]))
 
-        return psf_params.to(device)
+        return psf_params.to(self.generate_device)
 
     def _get_psf(self):
-        psf = torch.tensor([], device=device)
+        psf = torch.tensor([], device=self.generate_device)
         for i in range(self.n_bands):
             _psf = self._get_psf_single_band(self.params[i])
             _psf *= self.normalization_constant[i]
@@ -223,7 +228,7 @@ class ImageDecoder(pl.LightningModule):
         # output dimension is batch_size x n_tiles_per_image
 
         # always poisson distributed.
-        p = torch.full((1,), self.mean_sources, device=device, dtype=torch.float)
+        p = torch.full((1,), self.mean_sources, device=self.generate_device, dtype=torch.float)
         m = Poisson(p)
         n_sources = m.sample([batch_size, self.n_tiles_per_image])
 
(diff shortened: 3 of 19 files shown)


bliss/models/decoder.py
```diff
@@ -349,7 +355,7 @@ def sample_prior(self, batch_size=1):
     @staticmethod
     def _get_log_fluxes(fluxes):
         log_fluxes = torch.where(
-            fluxes > 0, fluxes, torch.ones(*fluxes.shape).to(device)
+            fluxes > 0, fluxes, torch.ones(*fluxes.shape).to(fluxes.device)
```
jeff-regier, line 358: Would it work to replace `torch.ones(*fluxes.shape).to(fluxes.device)` with `torch.ones_like(fluxes)`?

ZachariahPang: Oh right, this is much cleaner. I changed this in the next commit.

bliss/models/decoder.py
```diff
@@ -188,7 +190,7 @@ def forward(self):
         return psf
 
     def _get_psf(self):
-        psf = torch.tensor([], device=device)
+        psf = torch.tensor([], device=self.device)
         for i in range(self.n_bands):
             _psf = self._get_psf_single_band(self.params[i])
```
ismael-mendoza, line 195: hmm do you need to make sure that `psf_params` are in the correct device for this line to work ? (maybe this is why you have the `.to()` outside when you create the decoder?

bliss/models/decoder.py
```diff
@@ -240,7 +242,7 @@ def _sample_locs(self, is_on_array, batch_size):
             self.max_sources,
             2,
         )
-        locs = torch.rand(*shape, device=device)
+        locs = torch.rand(*shape, device=is_on_array.device)
```
ismael-mendoza, line 245: good workaround

bliss/models/encoder.py
```diff
@@ -228,11 +225,13 @@ def __init__(
...
 
         # get index for prob_n_sources, assigned indices that were not used.
-        self.indx_mats, last_indx = self._get_hidden_indices()
-        self.prob_n_source_indx = torch.arange(
-            last_indx,
-            self.dim_out_all,
-            device=device,
+        indx_mats, last_indx = self._get_hidden_indices()
+        for k, v in indx_mats.items():
+            self.register_buffer(k, v, persistent=False)
```
ismael-mendoza, line 230: now that the matrices are an attribute I guess I would rename them something like `k+'_indx'` to avoid confusion like the one below

config/dataset/single_tile.yaml
```diff
@@ -4,3 +4,5 @@ params:
   n_batches: 10
   batch_size: 6400
   num_workers: 0
+  generate_device: "cuda:0"
```
ismael-mendoza, line 7: there is a `gpus` argument in the main config file, is there a way you can propagate that value to this one?  (to avoid specifying two devices in two different places)

tests/test_encoder_forward.py
```diff
@@ -78,10 +78,10 @@ def test_forward(self, devices):
...
 
             # get index matrices
-            locs_mean_indx_mat = star_encoder.indx_mats["loc_mean"]
-            locs_var_indx_mat = star_encoder.indx_mats["loc_logvar"]
-            log_flux_mean_indx_mat = star_encoder.indx_mats["log_flux_mean"]
-            log_flux_var_indx_mat = star_encoder.indx_mats["log_flux_logvar"]
+            locs_mean_indx_mat = star_encoder.loc_mean
+            locs_var_indx_mat = star_encoder.loc_logvar
+            log_flux_mean_indx_mat = star_encoder.log_flux_mean
+            log_flux_var_indx_mat = star_encoder.log_flux_logvar
```
ismael-mendoza, line 84: yeah it would be best if the attribute was e.g. `loc_mean_indx` to avoid confusion that this is the matrix of indices.

bliss/main.py
```diff
@@ -12,3 +12,7 @@ def main(cfg):
     else:
         raise KeyError
     task(cfg)
+
+
+if __name__ == "__main__":
+    main()
```
ismael-mendoza, line 18: hmm do you know why `codecov` now says that this line is not covered (I thought it would be used by both `test_generate.py` and `test_train.py`)

ismael-mendoza (approved):
Looks really good, thanks for these changes @ZachariahPang the code is much cleaner without these calls and there is much less dependence on having a `device` at hand. I just have a couple of questions/suggestions in my review below.

config/dataset/m2.yaml
```diff
@@ -4,3 +4,4 @@ params:
   n_batches: 10
   batch_size: 20
   num_workers: 0
+  generate_device: "cuda:0"
```
jeff-regier, line 7: Should we specify a particular GPU here? I'm not sure. If it's being committed to our repo, it would seem like the setting would need to be appropriate for anyone who might use these config files. Would an environment variable be more appropriate?

Also, if we're calling it `generate_device`, does the mean the rest of the model might be on a different device? I'm not sure we'd want that ever. If we do use multiple GPUs, it would be in DDP mode (distributed data parallel), so the model would be copied across GPUs.

tests/test_encoder_forward.py
jeff-regier, line 84: Much better!

config/dataset/m2.yaml
ismael-mendoza: > Should we specify a particular GPU here? I'm not sure. If it's being committed to our repo, it would seem like the setting would need to be appropriate for anyone who might use these config files. Would an environment variable be more appropriate?

Good point, what I've been doing is specify the `gpus` the user wants to use in the [main config file](https://github.com/prob-ml/bliss/blob/master/config/config.yaml), create a `device` variable from this information in the config file (or the command line if the user changes it), and doing something like `torch.cuda.set_device(device)` not sure if there is a better approach.

bliss/datasets/simulated.py
```diff
@@ -20,7 +20,9 @@ def __init__(self, cfg: DictConfig):
 
         self.n_batches = cfg.dataset.params.n_batches
         self.batch_size = cfg.dataset.params.batch_size
-        self.image_decoder = ImageDecoder(**cfg.model.decoder.params)
+        self.image_decoder = ImageDecoder(**cfg.model.decoder.params).to(
+            cfg.dataset.params.generate_device
+        )
```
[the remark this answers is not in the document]
ZachariahPang: Since this `image_decoder` is not part of the training model. Pytorch-lightning won't move it to gpu even if the model is trained on gpu. It would be initialized on cpu and stays on cpu unless we don move it manually. Both are tested and speed change is not significantly different but generating on cpu does create a bottleneck.

bliss/models/decoder.py
ZachariahPang: Sorry I'm not sure which one you meant. `psf` should be in the correct device since it's created on `self.device`.
`psf_params` here is automatically managed by pytorch along with other parameters. So it doesn't need manually assign. 
https://github.com/prob-ml/bliss/blob/1802f0aefdc44d0bb482f1575f328954b4ddfe4c/bliss/models/decoder.py#L158-L170

bliss/models/encoder.py
ZachariahPang: This was so that attribute names align with `self.variational_params` keys.
https://github.com/prob-ml/bliss/blob/1802f0aefdc44d0bb482f1575f328954b4ddfe4c/bliss/models/encoder.py#L362-L363
But I can make those changes.

config/dataset/single_tile.yaml
ZachariahPang: I did this because this is the device for loading data and can be different from the device for training model. We may or may not use Data Parallel, but in that case, this `generate_device` is paired with a cpu process to generate data for models. And we would have multiple models on multiple gpus and those gpus might be different from `generate_device`. I guess I did this because I was trying to avoid the confusion of two different `decoder` objects potentially on two different devices as mentioned in the descriptions I just add.

But I can remove this and use the one from the main config file.

bliss/main.py
ZachariahPang: Both tests import `main()` from`train.py` and `generate.py`. This file is not tested. I can remove these lines as this is not needed for our code to run. I add this just to debug everything conveniently.

config/dataset/m2.yaml
ZachariahPang: > Should we specify a particular GPU here? I'm not sure. If it's being committed to our repo, it would seem like the setting would need to be appropriate for anyone who might use these config files. Would an environment variable be more appropriate?

I add this in alignment with the `gpus` in the [main config file](https://github.com/prob-ml/bliss/blob/master/config/config.yaml). So I guess the user would need to specify `generate_device` and also `gpus` for the training. But I can make those changes if needed.

> Also, if we're calling it generate_device, does the mean the rest of the model might be on a different device? I'm not sure we'd want that ever. If we do use multiple GPUs, it would be in DDP mode (distributed data parallel), so the model would be copied across GPUs.

Yes, models can be on different devices. This `generate_device` is for hosting the `image_decoder` object in the dataset which is different from the `image_decoder` in our model. This shouldn't break anything as this is similar to how data is moved from cpu to the model gpus before each batch. Here the data is moved from the generating gpu to model gpus. As for whether this would be a good choice for DDP, I'm not sure. Please see the description I just add.

ZachariahPang: I just double-checked and set `generate_device: "cuda:7"`. So the data would be generated on `cuda:7` and the model and training would be on `cuda:0`. It works. `cuda:7` has around 3% workload and around 1100mb memory used, `cuda:0` has around 70% workload and 3300mb memory used. This probably doesn't make much sense, just a sanity check at least that the devices and communication works. I tested "generate_device: cpu" earlier on our server and it also worked.

ZachariahPang: How about we leave the `generate_device` in the main config as optional. When not specified, we just use the first gpu in `gpus`, or cpu if `gpus` is `None`. But it can also be overridden from cli or in config to select a dedicated device for data generation in the event maybe our model saturates gpu memory.

jeff-regier (approved):
Looks great! Thanks for thinking through all these issues. Please merge whenever you're ready.

PR #182, opened 2021-01-20
Documentation

zzhaozhe-profolio:


jeff-regier:
We might consider posting our documentation on GitHub pages rather than readthedocs -- could be nice if it let's us avoid having another account to manage. But I haven't used github pages before. @zzhaozhe-profolio @ismael-mendoza What do you think?

zzhaozhe-profolio:
> We might consider posting our documentation on GitHub pages rather than readthedocs -- could be nice if it let's us avoid having another account to manage. But I haven't used github pages before. @zzhaozhe-profolio @ismael-mendoza What do you think?

I haven't used github pages neither, but I also prefer avoiding another account to manage. I can take a look at it.

Issue #186, opened 2021-01-21
Document dependency on BLISS_HOME environment variable


dereklhansen:
I tried running the tests on the Linux server and I noticed there needs to be "BLISS_HOME" present in the environment, but I don't see this documented anywhere. Is there a place we could document this?

Alternatively, Hydra comes with some helper functions that can return the absolute path of the original working directory (hydra.utils.get_original_cwd()). This may be useful to avoid hard-coding an absolute path. I don't know how much work this would be to change, however.

jeff-regier:
Good point. I'm not attached to the idea of having a BLISS_HOME environment variable. If there's a way to configure it so that documentation isn't even necessary, that would be ideal

jeff-regier:
Maybe something with the hydra config path?
https://hydra.cc/docs/advanced/search_path

dereklhansen:
I was thinking more along these lines: https://hydra.cc/docs/tutorials/basic/running_your_app/working_directory/

But now that I think about it, ```hydra.utils.get_original_cwd()``` would return the directory from which the script was called, which means we'd have to call the scripts from the same working directory. This is something else we'd have to document.

Since poetry seems to automatically register the top-level project in sys.path, the simplest solution might be just to use the `__file__` variable from the bliss module. This will be persistent regardless of the working directory (as long as poetry shell is activated).
```
(bliss-FSmgvuKM-py3.8) [dereklh@deeplearning-01:~/projects/bliss]
$ python
Python 3.8.5 (default, Jul 28 2020, 12:59:40) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import bliss
>>> bliss.__file__
'/home/dereklh/projects/bliss/bliss/__init__.py'
>>> 
(bliss-FSmgvuKM-py3.8) [dereklh@deeplearning-01:~/projects/bliss]
$ cd case_studies/
(bliss-FSmgvuKM-py3.8) [dereklh@deeplearning-01:~/projects/bliss/case_studies]
$ python
Python 3.8.5 (default, Jul 28 2020, 12:59:40) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import bliss
>>> bliss.__file__
'/home/dereklh/projects/bliss/bliss/__init__.py'
```

PR #189, opened 2021-01-24
Set BLISS_HOME on module load if not already set

dereklhansen:
This is what I came up with to solve #186.  We check if the user has BLISS_HOME in their environment. If they don't, we set it to a reasonable default (the directory of the loaded bliss module). This should work without changing code elsewhere.

### DIFF
--- a/bliss/__init__.py
+++ b/bliss/__init__.py
@@ -1,0 +1,6 @@
+from os import getenv, environ
+from pathlib import Path
+
+if getenv("BLISS_HOME") is None:
+    bliss_path = Path(__file__).resolve()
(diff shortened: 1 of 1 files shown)


jeff-regier:
Makes sense to me. @ismael-mendoza , what do you think?

jeff-regier:
I take it there wasn't a good way to do this with hydra? e.g. with a user-defined hydra file that doesn't get committed to the repo

ismael-mendoza (approved):

ismael-mendoza:
This solution makes sense to me as well. Something like what you suggest could be possible @jeff-regier but then it would require additional documentation for the user to set this up ?

jeff-regier:
Thanks @dereklhansen

PR #194, opened 2021-01-27
sdss-like galaxies

ismael-mendoza:
This PR enables drawing galaxies from a catsim catalog with sdss-like fluxes and sdss psf using galsim.


ismael-mendoza:
# Examples

SDSS galaxies: 
Screen Shot 2021-01-27 at 13 35 41
Screen Shot 2021-01-27 at 13 36 45

Simulated noisy galaxies: 
image

bliss/datasets/catsim.py
```diff
@@ -82,151 +41,186 @@ def __init__(
...
+            ab_magnitude(float): AB magnitude of source.
+        Returns:
+            float: Flux in detected electrons.
+        """
 
-        galaxy_builder = descwl.model.GalaxyBuilder(
-            single_obs, no_disk=False, no_bulge=False, no_agn=False, verbose_model=False
-        )
+        # TODO: other bands that are not 'r'
+        return 3000 * 10 ** (-0.4 * (ab_magnitude - 24))
```
ismael-mendoza, line 118: right now I just played with the numbers to get something like the fluxes in SDSS, is there a more principled way ?

.github/workflows/config.yml
```diff
@@ -6,55 +6,54 @@ name: tests
...
+        run: echo "BLISS_HOME=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+
+      - name: Install Poetry
+        run: |
+          pip install --upgrade pip
+          pip install "poetry >=1.1.4, <2.0.0"
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get install libfftw3-dev libeigen3-dev
```
ismael-mendoza, line 44: It's a bit ugly, but galsim needs these packages.

.github/workflows/config.yml
```diff
@@ -6,55 +6,54 @@ name: tests
...
+        run: echo "BLISS_HOME=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+
+      - name: Install Poetry
+        run: |
+          pip install --upgrade pip
+          pip install "poetry >=1.1.4, <2.0.0"
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get install libfftw3-dev
```
ismael-mendoza, line 44: add wheel to pyproject toml

bliss/datasets/catsim.py
jeff-regier: That seems fine. The results shouldn’t be too sensitive to differences in the distribution of fluxes between the simulated data and the real data. Probably better to error on the side of making the simulated too diverse

ismael-mendoza:
All tests pass in the GPU and CPU, so when CI passes I'll merge. Next PR will focus on predict mode on SDSS and evaluate performance. 

## Example galaxy deblending with SDSS-like conditions.

image

PR #197, opened 2021-01-27
Removes FITSIO dependency from SloanDigitalSkySurvey 

dereklhansen:
This removes the FITSIO dependency and refactors the SDSS PSF code a bit to be more object-oriented


jeff-regier (approved):
Outstanding!

PR #204, opened 2021-02-01
Improve encoder cnn

ZachariahPang:
### CNN models

|                          | epochs|val_detection_loss | val_acc_counts | val_locs_mae | val_star_fluxes_mae  |val_avg_ppv|val_avg_tpr| *training time|
|---                    |--- |---                        |---                       |---                   |---                                |---                 |---               |---          |
| **base line** 552k parameters |200|0.5995(0.492)       | 0                        | 0.5257(0.5079)|0.5712                      |0.8208       |0.5552(0.6269)|15m50s|
| **base line** 552k parameters |400|0.4456(0.285)      | ****                  | 0.4971|0.4753                     |0.8856     |0.6175(0.6857)|31m40s|
|**Residual downsample with pooling** 901k parameters|200|0.5244(0.496)|           ***           |0.4617             |0.4737                         |0.874          |0.5954(0.6572)|38m8s|
|**Residual downsample with pooling** 901k parameters|400|0.3409|***|0.4818(0.4043)|0.4408|0.8908(0.9001)|0.6342(0.681)|1h19m21s|

*`check_val_every_n_epoch=10, plotting=False `
**(): best iter
***For 2 out of 3 training runs, the model predicted the number of stars correctly on m2 for 1 batch. 
Screen Shot 2021-02-05 at 18 30 13
****For one training run, the model predicted the number of stars correctly on m2 for more batches.
Screen Shot 2021-02-05 at 19 20 14

--- 
update Mar 17
### ViT 
Our current implementation does not support model-parallel multi-gpu so I can only try with the parameters that would fit on a single gpu.

ViT with the original Transformer backend is too big to fit on a single GPU since our equivalent sequence length is 2500 and the complexity is O(n^2), so can't really test it.
I have tried to use Linformer as it has O(n) complexity. Using some much smaller models comparing to the ones proposed in the original ViT paper,  the best training loss is around 2, much worse than the CNN models above.

~One thing we can consider to improve and would allow us to try a slightly bigger model is to remove the padding as the original ViT uses positional encoding to give the model a sense of neighboring instead of using overlapping padding. I'm not sure if this would affect other parts of the model. Do you think this is a path we should try? @ismael-mendoza~

### Task
- [ ] Clean up the ResNet model and check improvement on the Hubble data with `pytest test/test_m2.py`
- [ ] Try Slot Attention Machine in another PR


jeff-regier:
Should we close this pr now that #245 is open?

PR #215, opened 2021-02-09
bliss predict mode

### LINKED ISSUE #172: bliss predict mode
Following #169, we'll have an executable script called `bliss` that will take a mode of operation as it's first argument. At first, these modes are just `train`, `tune`, and `generate`. I propose adding a forth mode: `predict`. From the command line, it might be called like this
```
$ poetry shell
$ bliss predict sdss 3900 6 299
```
Here the second argument, "sdss", specifies the data source is SDSS. Following #170, bliss will know where to look for SDSS data. The remaining three fields specify which run, camcol (camera column), and field from SDSS to process. Bliss would load this data using the `SloanDigitalSkySurvey` class and run our inference routines.

Eventually, it would be nice to specify regions of sky to process as a bounding box of WCS coordinates, but that's beyond the scope of this issue.

The output from `bliss predict` would be variational distribution for the image.

ismael-mendoza:
- [x] create `predict.py` save variational parameters to disk of each tile in an sdss image. 
- [x] `predict.py` loads encoder from a pytorch lightning checkpoint of a trained sleep_phase object
- [x] added example yaml file for predict `config/predict/sdss_basic.yaml`
- [x] tiling works on rectangular images. 
- [x] `map_estimate` works on rectangular images (probably more relevant after this PR when we evaluate the variational parameters we are saving?)

Closes #172


ismael-mendoza:
Hi @Runjing-Liu120 does my adjustment of `get_full_params` on `encoder.py` seem like the right idea for rectangular images? I'll test it on tomorrow on some rectangular images, it's just a little involved because `decoder.py` does not produce rectangular images at the moment.

bliss/models/encoder.py
```diff
@@ -126,18 +134,19 @@ def _sample_class_weights(class_weights, n_samples=1):
...
-    tile_coords = torch.LongTensor([return_coords(i) for i in range(n_ptiles)])
-
-    return tile_coords
+    tile_coords = []
+    for i in range(n_ptiles1):
+        for j in range(n_ptiles2):
+            tile_coords.append(
+                ((i % n_ptiles1) * tile_slen, (j % n_ptiles2) * tile_slen)
+            )
+    return torch.LongTensor(tile_coords)
```
jeff-regier, line 149: Is this tensor on the cpu?

bliss/main.py
```diff
@@ -9,6 +9,8 @@ def main(cfg):
         from bliss.tune import main as task
     elif cfg.mode == "generate":
         from bliss.generate import main as task
+    elif cfg.mode == "predict":
+        from bliss.predict import main as task
```
jeff-regier, line 13: Perfect

bliss/models/encoder.py
ismael-mendoza: Yes, but then it's moved to the GPU when returned using `type_as`

jeff-regier: Is there a way to create it on the GPU in the first place?

ismael-mendoza: yup I just pushed a change to do this

ismael-mendoza:
Hi @dereklhansen , I had to modify `test_fnp.py` both to speed up (in the CPU) and to make sure it passes the pylint precommit. Could you take a look to make sure I didn't break anything? Thanks!

tests/test_fnp.py
```diff
@@ -37,17 +37,18 @@ def test_fnp_onedim(self, devices):
                 gpus=devices.gpus,
                 max_epochs=1000 if devices.use_cuda else 2,
                 logger=None,
-                check_val_every_n_epoch=100,
+                check_val_every_n_epoch=100 if devices.use_cuda else 1,
```
dereklhansen, line 40: This change will actually slow things down; this arguments means run every 100 epochs

dereklhansen: You can probably revert this part; I actually didn't see your review request and made changes to speed up FNP on my own...

ismael-mendoza: yes! but I wanted to have coverage for the validation. Thanks for the changes!

ismael-mendoza:
Closes #172

PR #216, opened 2021-02-09
make website more beautiful

zzhaozhe-profolio:


jeff-regier:
Looks really nice!

Is there a way to add a color theme to the html?

zzhaozhe-profolio:
> Looks really nice!
> 
> Is there a way to add a color theme to the html?

@jeff-regier Yes, we can. In the conf.py, we can set:
```
html_theme_options = {
    'logo': 'logo.png'
}
``` 

For alabaster theme, there are many options for [colors](https://alabaster.readthedocs.io/en/latest/customization.html#Style%20colors ) . Of course, if we don't like this theme, we can change it anytime very easily.

PR #217, opened 2021-02-10
Cache Poetry Folder on Github Actions

dereklhansen:
This adds caching of the poetry cache folder, which should greatly speed up subsequent runs on Github Actions.

The cache will be regenerated every time the poetry.lock file is changed. We actually don't need to do this, but we don't want this cache folder getting too big since Github has a 5GB limit on cache sizes.

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -43,6 +43,13 @@ jobs:
       - name: Install fftw3
         run: |
           sudo apt-get install libfftw3-dev
+          
+      - name: Poetry cache
+        uses: actions/cache@v2
+        with:
+          path: |
+              ~/.julia
+          key: ${{ runner.os }}-${{ hashFiles('./poetry.lock') }}
 
       - name: Install poetry dependencies
         run: |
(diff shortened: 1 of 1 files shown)


dereklhansen:
I'll need to wait for the action to run once, then I'll restart it and report the difference in times

dereklhansen:
The first run of the action took 17m 31s, with 7m 29s in "Install poetry dependencies"

.github/workflows/config.yml
```diff
@@ -43,6 +43,13 @@ jobs:
       - name: Install fftw3
         run: |
           sudo apt-get install libfftw3-dev
+          
+      - name: Poetry cache
+        uses: actions/cache@v2
+        with:
+          path: |
+              ~/.cache/pypoetry
+          key: ${{ runner.os }}-${{ hashFiles('./poetry.lock') }}
```
jeff-regier, line 52: Nice.  Hashing the lock file is pretty clever way to decide whether to reinstall the dependencies.

jeff-regier (approved):

dereklhansen:
New time is 11m 24s. It takes about a minute to download and unzip the cache, but the "poetry install" command runs fast now.

PR #225, opened 2021-02-17
Evaluation of sleep_net on SDSS

ismael-mendoza:
# Progress

## Main steps
- [x] After rebasing and a few small updates the m2 test fails (the n_bands in each flux is incorrect!) So I'm looking into that
- [x] Create a case study and corresponding README and jupyter notebook
- [x] extract data from stripe 82
```
# run the make command inside the `data/sdss` directory after changing 
RUN:=94
CAMCOL:=1
FIELD:=12
```
- [x] extract catalog info provided by Bryan and understand columns
- [x] try simulating an image like the SDSS but with our decoder
- [x] try using Bryan's method to do one pass of the real data and solve the problem with `.train()` vs `.eval()`
- [ ] train VAE on larger variety of fluxes and more complex galaxy model 

## Other features

- [x] change black `--line-length=100`
- [ ] is there a way we can exclude more border and do more classification with the encoder?


ismael-mendoza:
Example

Screen Shot 2021-02-17 at 15 58 12

image

ismael-mendoza:
Here is a more recent example with `.eval()` turned on.

Screen Shot 2021-02-25 at 4 45 56 PM

ismael-mendoza:
> It looks like you found a good 200x200 region to focus on. We should try to get everything right here (at least that is above the detection threshold) before moving on. One direction forward may be to get more realistic galaxies from galsim. Another direction may be to use the other stripe 82 data to train the encoder and decoder.

Sounds good, I'm probably going to merge this branch and start looking into that.

ismael-mendoza:
all tests pass in cpu and gpu will continue this with the steps in issue #231

PR #229, opened 2021-02-24
main -> predict, train, tune, generate

### LINKED ISSUE #222: Change names of executable functions from main to respective names
For example in `train.py` we need to change `main` to `train` as @jeff-regier  suggested. Then propagate changes to `main.py` and avoid non top-level imports.

jeff-regier:
Give more descriptive names to four functions which were named `main`.

closes #222


ismael-mendoza (approved):
Looks good! merging

PR #233, opened 2021-03-04
Im/sdss like galaxies

### LINKED ISSUE #231: more realistic SDSS galaxies
In #225 , we saw that we need more realistic galaxy morphology and fluxes  in order to achieve better performance at the observed chunk. We should also use the correct SDSS PSF (I was using a toy one)

## main tasks

- [x] use cosmoDC2 catalog reference SDSS magnitudes, see column `mag_u/g/r/i/z/y_sdss`
- [x] see [here](https://github.com/LSSTDESC/WeakLensingDeblending/blob/cosmoDC2_ingestion/notebooks/wld_ingestion_cosmoDC2.ipynb) for how to convert cosmoDC2 catalogs to more standard bulge+disk format  and for how to extract the cosmoDC2 catalog from GCRCatalogs
- [x] fix psf and fluxes using the info [here](https://www.sdss.org/dr12/scope/). We can calculate fluxes with zero points [here](https://github.com/LSSTDESC/WeakLensingDeblending/issues/19) after changing argument to `'sdss2010-r' and flux equation [here](https://github.com/LSSTDESC/WeakLensingDeblending/blob/ddfdcb68d4a153181152784e4777618f3633c615/descwl/survey.py#L125).  
- [x] What is SDSS Sky brightness and is it consistent with background? (check if flux calculation is correct)
- [x] do we still have the issue where galaxies are too small looking? 
- [x]  get correct PSF of stripe 82 using `psf_at_points` function from `sdss.py`(use psf at middle of chunk for training since it won't change much)
- [x] double check galaxies and their sizes. 
- [x] train VAE with these galaxies and check for convergence (cache galaxies if necessary)
   - There is a problem where the loss decreases and converges, the residuals are not too big, but the model keeps converging to circular galaxies. The fluxes estimated also look ok after ~121 epochs. Optimization doesn't seem to be the problem but rather getting stuck in an undesirable mode. 
   - **Updates:** So far I have tried changing `slen` from 51-> 41, changing amount of noise via `noise_factor` between 0.1 and 0.01, and `batch_size`. Every time loss converges to something significantly different than starting point, converges to circular galaxies. 
   - [x] investigate if this could be due to posterior collapse and follow-up on this [message](https://probml.slack.com/archives/CQA9CND1V/p1611867228007400).
   - [x] more tests include: making sure randomization is OK, saving all galaxies and loading them much faster, checking if too many of galaxies are too circular (but if this is correct distribution, then...) 
   - [x] annealing does not help
   - [ ] save all galaxies into disk, so that we can train on all of them faster (and inspect some). Use `generate.py`
- [ ] retrain full BLISS sleep model
- [ ] check which objects in stripe 82 chunk are below detection limit.
- [ ]  calculate fluxes from stripe 82 using the same equation as before? (are they ab_mags?) See resources below.
- [ ] run bliss model in eval mode again on this chunk

## other tasks

- [ ]  avoid plotting region outside border that is undetected
- [ ]  is there a way we can exclude more border and do more classification with the encoder?
- [ ]  check magnitude and mark galaxies that are below detection threshold (maybe make crosses proportional to magnitude)

## Resources
* https://www.sdss.org/dr12/algorithms/magnitudes/
* https://speclite.readthedocs.io/en/latest/filters.html

### Graveyard
- ~~Try using Bryans' PSF stored in `bliss/data`, if not we can use a Gaussian approx (with the info above) for now if for some reason that doesn't work or we get weird results.~~

ismael-mendoza:
Closes #231


ismael-mendoza:
Here is an example of the more realistic galaxies I'm producing on SDSS, notice the wider variety of fluxes and morphologies: 

image

ismael-mendoza:
This example uses the `psf_at_points` function to get the psf 

image

ismael-mendoza:
I'm trying to start with an easy example and and then work from there. I subselected the brightest and most sizable galaxies from the dataset, here are some examples: 
[sdss_galaxies_example_images.pdf](https://github.com/prob-ml/bliss/files/6116660/sdss_galaxies_example_images.pdf)


Does anything look inherently "off"?

ismael-mendoza:
I'm closing this PR soon to merge the SDSS galaxy generation at least (I think these galaxies look good). In the next PR try out BIVA

PR #236, opened 2021-03-10
Fix ThreadWarning

### LINKED ISSUE #209: PytestUnhandledThreadExceptionWarning
https://github.com/prob-ml/bliss/runs/1823736764

ZachariahPang:
close #209 

@jeff-regier @ismael-mendoza 
This problem might be tricky. 
Originally, `main()` in `tune.py` sometimes would show as not covered. This was probably a false alert related to how Raytune handles multiprocess. I tried to insert some false assertion on purpose and confirmed that those lines were covered.  
To fix this, I turned out `local mode` in Raytune during testing to force it to spawn only one process. This fixed the coverage issue but would throw a thread warning in Pytest as this PR is trying to address.
I tried to manually ignore this warning either globally or only for that particular function. But both would show as not covered again. I searched a bit but couldn't find similar issues elsewhere and I'm not sure how to resolve this.


jeff-regier:
If we have to pick, I'd say it's more important to avoid cryptic warnings (`PytestUnhandledThreadExceptionWarning`) than to have 100% pytest coverage. There may just not be a good way to test multiprocessing code.

ismael-mendoza:
I agree with @jeff-regier, whatever is easiest for you to do that avoids the cryptic warnings (even at the expense of coverage) would be a good solution for me.

PR #244, opened 2021-03-30
Im/try new vae for galaxy models

ismael-mendoza:
Third time is the charm

- [x] enable using multiple optimizers with their parameters via a `get_optimizer` function. 
- [x] Enable using different VAE architecture (swap out Encoder/Decoder) for the `OneCenteredGalaxy` class. That way we can preserve progress on different architectures and what is known to work. (also we could store Bastien's VAE vs Francois' VAE for example, and that why compare against their galaxy models in the future; for these cases though a whole new VAE might be better)
- [ ] Maybe go back to CatSim Catalog (that used to work and they were bulge+disk galaxies) -- could assign magnitudes at random from the cosmo catalog. (are magnitudes galaxy-size independent???)
- [ ] Make training stable for Gaussian galaxies first. As a milestone; play with AdamW? 
- [ ] Try 32 as `slen`? 
- [ ] Implement Francois and Bastien VAE **exactly** (including loss, normalization)

- [ ] Try new latent_dim = 32; and no pr_penalty. 

## Progress 

- Implemented Bastien-style VAE architecture in `galaxy_net.py` (number of parameters from CNN to linear layers is ~1M)
- Optimization is difficult for Gaussian galaxies, still sometimes see circular galaxies. I noticed once, when switching to AdamW that there was a "hop" where it was able to create elliptical galaxies when it seemed stuck in circular galaxies. 
- Still circular galaxies for SDSS galaxies.  

### attempts: 
- switch Adam to AdamW (using default parameters)
- batchsize 64vs128 
- warm-up, beta-VAE do not help
- latent dim to 16 does not help.
- also tried 16 channel's for everything (Jeff's architecture)
- try L1-loss

## Questions 
- Initially I was thinking there was something with the data; but since there is a problem with the Gaussian galaxies now... 
- Bryan mentioned that he was able to get it to work with only 10 galaxies, and when he added more it broke down. What does this mean? 
- Maybe should get the Gaussian galaxies stably optimizing first?


ismael-mendoza:
Hi @jeff-regier -

I incorporated some of your and Francois' advice on the galaxy VAE in `bliss/models/galaxy_net.py`. I'm currently shooting for getting a VAE that more easily learns the toy gaussian dataset (training is more stable). Is this a reasonable goal? It's been hard for me to figure out concrete steps that make progress towards our goal so any advice would be much appreciated. 

Currently the optimization of this new architecture is not doing much better than our old one, this architecture gets stuck in bad local optima (e.g. predicts only background or circular galaxies) **8/20 times** in the toy Gaussian dataset. If you or @Runjing-Liu120  have any additional advice on what to try or change in this architecture in `galaxy_net.py` that would also be really helpful. Thanks!

I also tried this architecture on the more diverse SDSS galaxy dataset (following the study by @Runjing-Liu120) and it still gets stuck in the same circular local optima.

jeff-regier:
Are you trying to get (non-probabilistic) autoencoder working first, before working with a VAE? I recommend that. It makes sense to continue focusing on the toy dataset.

jeff-regier:
Have you been using residual/skip connections? Spatial dropout (like @ZachariahPang has been using for starnet) also may helpful.

But that may be more advanced than we need. If you make the bottleneck layer wide enough, an autoencoder shouldn't have trouble reconstructing those galaxy images.

PR #247, opened 2021-04-14
Refactor tiling

ZachariahPang:
This PR refactors our tiling mechanism in `encoder.py`

The previous implementation uses an "identity" conv2d to make tiles, which requires ~50 lines of code and need to store a sparse weight matrix.
This new implementation uses 5 lines of code and might be more readable.

PS:
- We don't have to use `einops.rearrange` but I think it's more readable. I can change it to use the more standard chains of reshape and permute if you think `einops` is unnecessary.
- I haven't removed the old implementation so that I can assert that I don't accidentally change anything. If this looks good, I can remove the old ones and merge.

### DIFF
--- a/bliss/models/encoder.py
+++ b/bliss/models/encoder.py
@@ -299,11 +299,12 @@ class ImageEncoder(nn.Module):
         # misc
         self.register_buffer("swap", torch.tensor([1, 0]), persistent=False)
 
-    def make_patch(self, image):
+    def get_images_in_tiles_new(self, image):
         channels = image.shape[1]
         window = self.ptile_slen
         stride = self.tile_slen
         tiles = F.unfold(image, kernel_size=window, stride=stride)
+        # b=batch, c=channel, h=tile_height, w=tiles_width, n=num_of_tiles_for_each_batch
         tiles = rearrange(tiles, "b (c h w) n -> (b n) c h w", c=channels, h=window, w=window)
         return tiles
 
@@ -359,7 +360,7 @@ class ImageEncoder(nn.Module):
         old_implementation = output.reshape(
             n_ptiles, self.n_bands, self.ptile_slen, self.ptile_slen
         )
-        new_implementation = self.make_patch(images)
+        new_implementation = self.get_images_in_tiles_new(images)
         assert torch.allclose(old_implementation, new_implementation)
         return new_implementation
 
--- a/bliss/models/encoder.py
+++ b/bliss/models/encoder.py
@@ -1,4 +1,5 @@
 import numpy as np
+from einops import rearrange
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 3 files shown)


bliss/models/encoder.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+from einops import rearrange
```
jeff-regier, line 2: einops looks useful, but there's also some cost to introducing new dependencies -- especially if only some of our code is using it. @ismael-mendoza Would you take a look at [einops](http://einops.rocks/pytorch-examples.html) too? Do we want to use it throughout bliss?

bliss/models/encoder.py
```diff
@@ -347,7 +357,12 @@ def get_images_in_tiles(self, images):
         n_ptiles_per_image = slen * wlen / self.tile_slen ** 2
         assert n_ptiles_per_image % 1 == 0, "n_ptiles_per_image must be an int"
         n_ptiles = int(n_ptiles_per_image * batch_size)
-        return output.reshape(n_ptiles, self.n_bands, self.ptile_slen, self.ptile_slen)
+        old_implementation = output.reshape(
+            n_ptiles, self.n_bands, self.ptile_slen, self.ptile_slen
+        )
+        new_implementation = self.get_images_in_tiles_new(images)
+        assert torch.allclose(old_implementation, new_implementation)
```
jeff-regier, line 364: Well that's convincing :-)

bliss/models/encoder.py
ZachariahPang: Other than the one in this PR and the Multihead attention as shown [here](http://einops.rocks/pytorch-examples.html),
I just checked we have around 60+ of `torch.reshape`/`torch.view` in our code base. 
For example https://github.com/prob-ml/bliss/blob/d6cb336691de0d2a4d337dbe2b8c0a9cee8f27f2/bliss/metrics.py#L110
can be changed to something like 

```python
est_n_gal = einops.reduce(true_params["galaxy_bool"], "b n t -> b", "sum")
```
so that it's very clear which axes are being summed over

If `einops`'s syntax looks appealing , I can make another PR and change them altogether. But I agree, this is another layer of dependency that we may not want to add.

jeff-regier: That's very nice syntax. The tiling code we have is really complex now, and it seems like einop could make it much more readable. If you're up for submitting a second pr that update the rest of our codebase to eigop I'm very supportive of this change.

ismael-mendoza: Thanks @ZachariahPang for looking into this, it seems that `einops` is _really_ light weight (does it even have any dependencies other than python itself?). I'm also supportive since it's a lot more readable from what I can see.

bliss/models/encoder.py
```diff
@@ -298,6 +299,15 @@ def __init__(
         # misc
         self.register_buffer("swap", torch.tensor([1, 0]), persistent=False)
 
+    def get_images_in_tiles_new(self, images):
+        channel = images.shape[1]
+        window = self.ptile_slen
+        stride = self.tile_slen
+        tiles = F.unfold(images, kernel_size=window, stride=stride)
+        # b=batch, c=channel, h=tile_height, w=tile_width, n=num_of_tiles_for_each_batch
+        tiles = rearrange(tiles, "b (c h w) n -> (b n) c h w", c=channel, h=window, w=window)
```
ismael-mendoza, line 308: this looks a lot more readable!

jeff-regier:
Nice

PR #251, opened 2021-04-19
remove all np.random

ismael-mendoza:
Following @zzhaozhe-profolio comment, this[ blog post](https://tanelp.github.io/posts/a-bug-that-plagues-thousands-of-open-source-ml-projects/), and pytorch issue **5059**. I removed all `np.random` calls in the galaxy datasets I've been using. This is a pretty subtle bug, so I think it's probably best to resolve it now rather than later. 

I don't think `np.random` is used elsewhere in the codebase (only in some unittests to generate synthetic test data), but please let me know @jeff-regier or @zzhaozhe-profolio if I missed any other places.


jeff-regier (approved):
Makes sense

ismael-mendoza:
Thanks, merging now

PR #252, opened 2021-04-29
Consolidate autoencoder progress

### LINKED ISSUE #223: stability of test_galaxy_vae
Sometimes it doesn't converge, need to investigate why. Maybe we could increase SNR of galaxies simulated to stabilize it.

- [x] does the batch_size, SNR affect it? 
- [x] try using noise_factor instead, how does that change things? 
- [x] warm_up scheme helps? 
- [ ] using dropout vs not (ziteng's paper)
- [ ] experiment [Bastien's VAE](https://arxiv.org/pdf/2005.12039.pdf)

### LINKED ISSUE #240: Galaxy model for SDSS galaxies
- [x] try [BIVA](https://github.com/vlievin/biva-pytorch)
    - [x] can we treat everything as a black box? how to get ELBO for optimization? 
    - [x] not clear how many layers, etc. maybe copy CIFAR example? 
    - [x] take a look at the evaluator (VariationalInference, for how to get elbo)
- [x] try normalizing images using sinh/tanh normalization (scarlet, Bastien, etc.)
- [x] try adding more layers to our VAE, no dropout? 
- [ ] Maybe try a more diverse dataset of Gaussians first.
    - It seems like the diversity of fluxes/sizes does not represent a problem for the VAE. There must be something about the bulge+disk shape. It does have a bit more trouble with optimization though (stability)
    - [x] maybe another optimizer? 
    - [ ] To do it in a systematic/principled way, could create a new catalog of fluxes/hlr from the SDSS catalog/images and use that.
    - [x] check psf_fwhm from SDSS dataset
    - [x] try a bigger stamp_size just to discard that as an issue for the optimization.
    - [ ] what background in the SDSS Chunk we are optimizing for? 
- [ ] experiment [Bastien's VAE](https://arxiv.org/pdf/2005.12039.pdf) and [code](https://github.com/LSSTDESC/DeblenderVAE/blob/master/scripts/tools_for_VAE/tools_for_VAE/model.py). (normalizing with asinh mapping + Bernoulli distribution instead of Gaussian)
- [ ] Francois [VAE](https://github.com/McWilliamsCenter/deep_galaxy_models/blob/master/deepgal/VAEEstimator.py) and [paper](https://arxiv.org/pdf/2008.03833.pdf)?

### LINKED ISSUE #253: all classes should have explicit keyword arguments not cfg
i.e. no classes should have this structure: 

```
MyClass(cfg: DictConfig)
```

since this hides the parameters that are used for that class and makes it hard to initialize without the config object which can be annoying at times.

ismael-mendoza:
Consolidate autoencoder model progress. 

- Closes #240 
- Closes #223
- Closes #253 

# Summary 

This PR replaces the galaxy VAE with with a galaxy autoencoder. The galaxy autoencoder should work 100% of the time with the ToyGaussian dataset and the SDSSGalaxies dataset in `bliss.dataset.galsim_galaxies` as can be seen by running `test_galaxy_autoencoder.py`. 

I also incorporated the autoencoder into the `ImageDecoder` and the galaxies generated from this decoder are now SDSS-like bulge+disk galaxies. I skip the normalizing flow for now, but we can come back to it later. What is required is a `latents_file` (example in `data/latents_simulated_sdss_galaxies.pt`) which contain pre-processed encodings of a random sample of galaxies from `SDSSGalaxies`. The encodings can then be decoded by the trained autoencoder in `models/sdss_autoencoder.ckpt`; the `ImageDecoder` randomly samples from dataset of latent encodings and that is how galaxies can now be produced for sleep phase training. 

All our previous tests pass with these more complicated galaxies (instead of Gaussian galaxies) which is encouraging. The measurement results of these galaxies in the sleep-phase also seem sensible. Although we still have issues with detecting some of them and the measurement is not very good for some of the very large galaxies (`ptile_slen` might be too small?), see example  sleeph-phase reconstructions below. (note this image contains both stars and galaxies)

Probably more work should go into improving detections and reconstructions depending on SDSS results. 

example_reconstructions


ismael-mendoza:
All tests are passing in GPU and CPU so merging. See **Summary** above for what this PR actually did (diff is too big to review)

PR #259, opened 2021-05-05
Refactoring with Einops

### LINKED ISSUE #248: Use of self.forward()
I just notice this and I'm not sure if it's intentional, but should we use  `self()` instead of `self.forward()`? I think `self()` is the recommended way to do a forward pass since `__call__` applies hooks and other stuff besides just calling `self.forward()`. This might cause some subtle bugs later on.
https://github.com/prob-ml/bliss/blob/f18dd87cfab26bf0ef1b980db340179020977cad/bliss/models/encoder.py#L583
https://github.com/prob-ml/bliss/blob/f18dd87cfab26bf0ef1b980db340179020977cad/bliss/wake.py#L88

### LINKED ISSUE #249: einops
Replace calls to `view` and `squeeze`/`unsqueeze` with calls to `rearrange` (from the einops package) throughout the codebase.

Consider using `reduce`  and `repeat` from einops too.

ZachariahPang:
close #249
close #248 

This PR refactors `.view`, `.reshape`, `.squeeze`, `.sum`,  `.repeat` to `einops` counterparts except when:
- shapes were inferred somewhere else and it's already clear what each dimension represents. [example](https://github.com/prob-ml/bliss/blob/64d9f1138949ec8061ac7daf85d2f16c0a0d3faa/bliss/models/decoder.py#L392-L408)
- squeeze or unsqueeze for broadcasting. [example](https://github.com/prob-ml/bliss/blob/64d9f1138949ec8061ac7daf85d2f16c0a0d3faa/bliss/models/decoder.py#L184-L185) [example](https://github.com/prob-ml/bliss/blob/64d9f1138949ec8061ac7daf85d2f16c0a0d3faa/bliss/models/decoder.py#L234)

I disabled the "too many statements" pylint check so that I can make line-by-line assertions making sure I don't change anything. I'll remove the old implementation and enable the "too many statements" back altogether at the end.

UPDATE:
@jeff-regier @ismael-mendoza There are a few questions I left in in-line comments that I need some help with. 
Also, I didn't touch `/case_studies/starnet/plotting.py` because I'm not sure if it's still maintained. Please let me know what you think. Thank you very much.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -40,6 +40,7 @@ disable=too-many-ancestors,
 	too-many-arguments,
 	too-many-locals,
 	too-many-instance-attributes,
+	too-many-statements,
 
 	# we should remove fixme comments
 	fixme,
@@ -57,4 +58,3 @@ disable=too-many-ancestors,
 
 # Minimum lines number of a similarity to report duplicate-code
 min-similarity-lines=11
-
--- a/.pylintrc
+++ b/.pylintrc
@@ -40,7 +40,6 @@ disable=too-many-ancestors,
 	too-many-arguments,
 	too-many-locals,
 	too-many-instance-attributes,
-	too-many-statements,
 
 	# we should remove fixme comments
 	fixme,
--- a/bliss/sleep.py
+++ b/bliss/sleep.py
@@ -214,18 +214,26 @@ class SleepPhase(pl.LightningModule):
         if self.use_galaxy_encoder:
             batch_size = images.shape[0]
             max_detections = 1
-            # NOTE: tile_est["locs"].shape = (2, 100, 1, 2)
-            tile_locs = tile_est["locs"].reshape(-1, max_detections, 2)
+            tile_locs_old = tile_est["locs"].reshape(-1, max_detections, 2)
+            tile_locs = rearrange(
+                tile_est["locs"], "b n d xy -> (b n) d xy", b=batch_size, d=max_detections
+            )
+            assert torch.allclose(tile_locs_old, tile_locs)
             image_ptiles = self.image_encoder.get_images_in_tiles(images)
             tile_galaxy_params = self.forward_galaxy(image_ptiles, tile_locs)
+
             n_galaxy_params = tile_galaxy_params.shape[-1]
-            tile_galaxy_params = tile_galaxy_params.reshape(
+            tile_galaxy_params_old = tile_galaxy_params.reshape(
                 batch_size,
                 -1,
                 max_detections,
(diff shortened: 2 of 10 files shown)


.pylintrc
```diff
@@ -40,6 +40,7 @@ disable=too-many-ancestors,
 	too-many-arguments,
 	too-many-locals,
 	too-many-instance-attributes,
+	too-many-statements,
```
jeff-regier, line 43: Can we remove `too-many-statements` from here before merging?

ZachariahPang: Yes  I'll remove this later

bliss/sleep.py
```diff
@@ -213,6 +214,7 @@ def tile_images_map_estimate(self, images):
         if self.use_galaxy_encoder:
             batch_size = images.shape[0]
             max_detections = 1
+            # NOTE: tile_est["locs"].shape = (2, 100, 1, 2)
```
ZachariahPang, line 217: @ismael-mendoza can you recall what its second dimension `100` represents here?

bliss/sleep.py
```diff
@@ -223,6 +225,7 @@ def tile_images_map_estimate(self, images):
                 max_detections,
                 n_galaxy_params,
             )
+            # NOTE: tile_est["locs"].shape = (2, 100, 1, 8)
```
ZachariahPang, line 228: @ismael-mendoza  same as above. Can you recall what its second dimension 100 represents here?

bliss/sleep.py
```diff
@@ -264,7 +267,10 @@ def get_galaxy_loss(self, batch):
         )
 
         recon_losses = -Normal(recon_mean, recon_var.sqrt()).log_prob(images)
-        recon_losses = recon_losses.view(batch_size, -1).sum()
+        recon_losses_old = recon_losses.view(batch_size, -1).sum()
+        # NOTE: why do we need .view here
```
ZachariahPang, line 271: Not sure if there's something else subtle going on here but why do we need `.view().sum()` instead of directly `.sum()`?

jeff-regier: Just using `.sum()` is fine, better even. It's probably the way it is because someone was debugging the code and was looking at just the recon_loss for a particular sample

bliss/sleep.py
jeff-regier: You may not hear from Ismael while he's traveling, but don't let that stop you. It's likely that 100 is the number of tiles in a 10x10 grid of tiles.

PR #263, opened 2021-06-11
Cuda visible devices

ismael-mendoza:
This small PR introduces the changes that we discussed in our last meeting related to `CUDA_VISIBLE_DEVICES` flag. Assuming that everyone now sets this flag in their `.bashrc`, we now expect that the `gpus` flag of pytorch_lightning  be `1, 2, ...` depending on how many gpus to use (not their indices). 

This change has been propagated to `config/config.yaml` and  unit tests (where always 1 gpu is used). Any additional suggestions @ZachariahPang ?

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -13,8 +13,7 @@ mode: train
 general:
   overwrite: False
 
-gpus:
-  - 0
+gpus: 1 # use a single gpu by default.
 
 paths:
   root: ${oc.env:BLISS_HOME}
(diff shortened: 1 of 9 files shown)


ismael-mendoza:
Hi @ZachariahPang - I decided to revert the changes on updating the poetry packages and will do that in a separate PR

ZachariahPang:
I add a multi_gpu smoke test so that we don't need to manually test those to maintain what we did in #180. 
By default, the newly added `test_multi_gpu.py` would only run with `--gpu` and when two or more gpus are available, otherwise it would be skipped instead of `pass` which might be confusing.

ZachariahPang (approved):

PR #265, opened 2021-06-14
Misclassification accuracy

ismael-mendoza:
Implements various metrics/plots which will be useful for evaluating galaxy/star misclassification issues. 

- [x] Marker size as a function of confidence. 
- [ ] Misclassification accuracy (on true detected objects) implemented in `metrics.py`
- [ ] New results on simulated data and SDSS chunk added.


ismael-mendoza:
The marker sizes as a function of prediction confidence has been implemented. Here are some examples with simulated data: 

image

ismael-mendoza:
with annotated probabilities 

image

PR #271, opened 2021-07-02
only need to crop = tile_slen not 2*tile_slen

ismael-mendoza:
I realized that we only need to crop `tile_slen` instead of `2*tile_slen` in the `encoder.center_ptiles` function. 

Recall that in each padded tile which contains exactly one galaxy, we center the padded tile on that galaxy using the `grid_sample` function to train the galaxy encoder in the sleep phase. We decided to crop this "centered padded tile" in order to avoid sharp boundaries with 0s after the `grid_sample`. I thought that I needed to crop `2*tile_slen` on each side of this padded tile to avoid the sharp boundaries, but I realized I only need to crop half! The reason is that if a galaxy is at the border of a tile, the maximum distance it can be from the center of tile is `sqrt(2)/ 2 * tile_slen` which means that `grid_sample` only shifts the image by at most this quantity and `sqrt(2)/ 2 < 1`  :)

This change will throw away less information from going to the padded tiles -> centered padded tiles. This could potentially improve improve the galaxy encoder performance since more information will be available to it. (I still need to test)

Does this make sense @jeff-regier and @Runjing-Liu120 ? If not, we can discuss in the next meeting too. Also, please let me know if I'm misunderstanding grid_sample?

### DIFF
--- a/bliss/models/encoder.py
+++ b/bliss/models/encoder.py
@@ -320,7 +320,7 @@ class ImageEncoder(nn.Module):
         assert tile_locs.shape[1] == 1
         assert image_ptiles.shape[-1] == self.ptile_slen
         n_ptiles = image_ptiles.shape[0]
-        crop_slen = 2 * self.tile_slen
+        crop_slen = self.tile_slen
         ptile_slen = self.ptile_slen
         assert tile_locs.shape[0] == n_ptiles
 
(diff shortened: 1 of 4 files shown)


jeff-regier (approved):
Sounds good

ismael-mendoza:
Thanks!

PR #273, opened 2021-07-14
Improve galaxy_net `plot_reconstruction`

ZachariahPang:
Small improvements, instead of the independent color range in the resulting reconstruction, now it has:
- constant color range across all *residual* images so that they are easier to be compared visually
- constant color range for each pair of *image* and *recon_mean*
- plotted samples are the ones with the largest average residual instead of being randomly selected
- group metrics by `val` and `train`
- group images by `epoch`
- add input recon grid
- remove `self.optimizer_params = optimizer_params` since it's automatically assigned to `self.hparams["optimizer_params"]` by `self.save_hyperparameters()`

## plotted at 499 epoches
individualImage

---
- [x] fix `FIXME`'s
- [x] plot best examples
- [x] update `plot_reconstruction` comment
- [x] see if we really need to train 500 epochs
- [ ] ~~research on activation functions~~


bliss/models/galaxy_net.py
```diff
@@ -97,6 +97,7 @@ def __init__(
         self.register_buffer("zero", torch.zeros(1))
         self.register_buffer("one", torch.ones(1))
 
+        # FIXME it's available through self.hparams["optimizer_params"]
```
ismael-mendoza, line 100: it would be great if you could resolve this FIXME's in this PR :)  (I tend to forget about inline TODO comments)

bliss/models/galaxy_net.py
```diff
@@ -139,43 +140,68 @@ def validation_step(self, batch, batch_idx):  # pylint: disable=unused-argument
...
         if self.logger:
             self.logger.experiment.add_figure(f"Images {self.current_epoch}", fig)
 
     def plot_reconstruction(self, images, recon_mean):
-        # only plot i band if available, otherwise the highest band given.
+
+        # 1. only plot i band if available, otherwise the highest band given.
+        # 2. plot the `num_examples` of images with the largest average residual
+        #    across all batches in the last epoch
+        # 3. residual color range (`vmin`, `vmax`) are fixed as across all samples
```
ismael-mendoza, line 161: It might be good to mention here that you fix them across all images to be the maximum/minimum value across all images. (right?)

bliss/models/galaxy_net.py
```diff
@@ -139,43 +140,68 @@ def validation_step(self, batch, batch_idx):  # pylint: disable=unused-argument
...
             plt.subplot(num_examples, num_cols, num_cols * i + 1)
             plt.title("images")
-            plt.imshow(images[i, 0].data.cpu().numpy())
+            plt.imshow(image.numpy(), interpolation=None, vmin=vmin, vmax=vmax)
             plt.colorbar()
 
             plt.subplot(num_examples, num_cols, num_cols * i + 2)
             plt.title("recon_mean")
-            plt.imshow(recon_mean[i, 0].data.cpu().numpy())
+            plt.imshow(recon.numpy(), interpolation=None, vmin=vmin, vmax=vmax)
```
ismael-mendoza, line 194: Thanks the `interpolation=None` is actually important. I think this is done by default but it's good to be explicit here.

bliss/models/galaxy_net.py
ismael-mendoza: since they are pretty small I wouldn't bother with a separate PR

ismael-mendoza (approved):
Thanks @ZachariahPang! Please merge when my minor comments are resolved

ismael-mendoza:
@ZachariahPang thanks for posting some of the autoencoder reconstructions - it's definitely useful to see in which examples we are doing the 'worst' (at least using the average residual definition). From these examples, it's clear that we have trouble capturing "long-sharp" wings - in this case it's really clear to see: 

Screen Shot 2021-07-14 at 8 02 57 AM

It's good to be sure now that we fail at doing this at the autoencoder level (and not only in the sleep phase). 

I'm surprised to see these two  among the 'worst' examples: 

Screen Shot 2021-07-14 at 8 05 41 AM

since the galaxies are small and not very elliptical. By eye it seems the second one is a bit sharper than the reconstruction and maybe the first one is just too bright? Our network is not very deep so we might have trouble capturing these features perfectly.

ismael-mendoza:
Overall, if these are the worst examples I think the reconstructions are pretty good ( ideally the residuals should look like gaussian noise and if these are the worst outliers [-5, 5] makes sense). I'm leaning towards moving on to sleep phase, but it might still be good to check some of the more interpretable metrics we talked about first (we can talk more in the meeting today)

@ZachariahPang I had a couple of follow-up questions on this useful exploration:

(1) Do you think the short-comings in the residual are just because our network is not very deep?
(2) Are there any simple changes (e.g. hyper-parameter changes) that you can think of to make the network training easier/faster?
(3) Could you confirm whether residuals kept getting better up until 500 epochs or are these plots essentially the same to the 300 epoch ones? It would be good to have a sense for how many epochs/galaxies are necessary to have accurate reconstructions before moving on to the sleep phase.

bliss/models/galaxy_net.py
ZachariahPang: Not sure this was set in a global config somewhere else. I thought `interpolation='antialiased'` was the default setting. If this is the case, is there other places I should look into and maybe change those as well?

ZachariahPang:
> @ZachariahPang thanks for posting some of the autoencoder reconstructions - it's definitely useful to see in which examples we are doing the 'worst' (at least using the average residual definition). From these examples, it's clear that we have trouble capturing "long-sharp" wings - in this case it's really clear to see that:
> 
> Screen Shot 2021-07-14 at 8 02 57 AM
> 
> It's good to be sure now that we fail at doing this at the autoencoder level (and not only in the sleep phase).
> 
> I'm surprised to see these two among the 'worst' examples:
> 
> Screen Shot 2021-07-14 at 8 05 41 AM
> 
> since the galaxies are small and not very elliptical. By eye it seems the second one is a bit sharper than the reconstruction and maybe the first one is just too bright? Our network is not very deep so we might have trouble capturing these features perfectly.

The last two cases look really weird. I didn't show it here. But generally, the AE did a pretty good job (less than 0.2 residuals) on those small rounded-shaped galaxies. I should probably also plot the best ones just to see if the other side of the spectrum.

On that matter, do you think it would a good idea to log all those examples to a `.pt` file and use another script such as `plotly` to explore them fully, or maybe plotting the worst and the best ones directly in Tensorboard is enough?

ZachariahPang:
> (1) Do you think the short-comings in the residual are just because our network is not very deep?

I feel like I don't have enough intuition here but given that our galaxynet only has two conv layers, I think we can benefit from a deeper model.

> (2) Are there any simple changes (e.g. hyper-parameter changes) that you can think of to make the network training easier/faster?

I'll try to research a bit on the activation function since I remember I read somewhere that those play a more important role in AE and sometimes Relu might not even work very well.

> (3) Could you confirm whether residuals kept getting better up until 500 epochs or are these plots essentially the same to the 300 epoch ones? It would be good to have a sense for how many epochs/galaxies are necessary to have accurate reconstructions before moving on to the sleep phase.

The loss is plateaued pretty quickly but I'll check some early epoch examples as well to see if those make any difference visually. I guess it's probably not the case? Since our galaxy model for now only sees the loss so it should not have an idea of what is more visually pleasing?

bliss/models/galaxy_net.py
```diff
@@ -139,43 +140,68 @@ def validation_step(self, batch, batch_idx):  # pylint: disable=unused-argument
...
+        residual_vmax = torch.ceil(residuals.max().cpu()).numpy()
+        residual_vmin = torch.floor(residuals.min().cpu()).numpy()
         plt.ioff()
 
         fig = plt.figure(figsize=(10, 25))
 
         for i in range(num_examples):
+            image = images[i, 0].data.cpu()
+            recon = recon_mean[i, 0].data.cpu()
+            vmax = torch.ceil(torch.max(image.max(), recon.min()))
```
ismael-mendoza, line 184: here it should be `recon.max()` right?

jeff-regier:
@ZachariahPang Is this ready to merge?

PR #285, opened 2021-07-28
Decouple galaxy encoder and sleep phase

ismael-mendoza:
This will make it easy for @Runjing-Liu120 to incorporate the FluxEncoder into bliss more directly

## Summary of changes

* Sleep Phase and GalaxyEncoder can be trained independently.
* Corresponding Configuration files created
* Added test for galaxy_encoder
* `predict.py` expects both models (sleep_net and galaxy_encoder) to be loaded 
* `predict.prediction` automatically takes care of map estimates and variational parameters based on both models. Can be used broadly for any image where we want BLISS predictions after both models have been loaded. 
* Updated notebooks with new models and way of doing predictions and new results

## Misc changes 
* `train_loss` -> `train/loss` (in sleep.py and galaxy_encoder.py)
* remove `hp_metric` from tensorboard 
* Overall refactoring of sleep/decoder/encoder.py


ismael-mendoza:
This PR seems ready, I'm just running in the GPU and then will merge.

ismael-mendoza:
Latest galaxy reconstruction results look pretty good on simulated data
individualImage

ismael-mendoza:
Note, the precision/recall decreased a bit but the confusion matrix improved in the `aggregate-results.ipynb` notebook. I think since I trained the image_encoder for more epochs it probably overfitted to the training i.e. we need to adjust the prior or train less?

PR #287, opened 2021-08-01
refactoring / flake8 compliance

jeff-regier:
This PR improves the compliance of our codebase with the [wemake-python-styleguide](https://github.com/wemake-services/wemake-python-styleguide), which checks syntax using flake8. (It also disables many wemake-python-styleguide checks that aren't useful for our situation.) These changes should make our code more readable and maintainable.


jeff-regier:
All tests are passing on the GPU too.

.pre-commit-config.yaml
```diff
@@ -1,23 +1,47 @@
 repos:
-  - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.4.0
+  - repo: local
```
jeff-regier, line 2: This way pre-commit runs the same versions that we (developers) run locally. All versioning is done by poetry.

bliss/datasets/galsim_galaxies.py
```diff
@@ -229,9 +229,9 @@ def __init__(
         assert sdss_kwargs["bands"][0] == 2
         assert len(list(psf_points)) == 2
         sdss_data = SloanDigitalSkySurvey(**sdss_kwargs)
-        _psf = sdss_data.rcfgcs[0][-1]
```
jeff-regier, line 232: The convention is that a variable beginning with an underscore will not be used. But for some reason lots of variables in our codebase start with underscores even though they are used. Really confusing.

bliss/datasets/sdss.py
```diff
@@ -1,16 +1,16 @@
 import pathlib
 import pickle
 import warnings
-from math import floor, ceil
+from math import ceil, floor
```
jeff-regier, line 4: The `isort` program sorts all the imports. That way there's only one way to do it, so the diffs for subsequent PRs will be smaller.

bliss/models/decoder.py
```diff
@@ -484,12 +480,9 @@ def _construct_full_image_from_ptiles(image_ptiles, tile_slen, border_padding):
                 )
 
                 # get canvas
-                canvas[
-                    :,
-                    :,
-                    (i * tile_slen) : (i * tile_slen + canvas_len),
-                    (j * tile_slen) : (j * tile_slen + canvas_len),
-                ] += image_tile_cols
```
jeff-regier, line 492: Black auto-formatting has led to some really unreadable code. I refactored some of it.

jeff-regier:
I'm going to go ahead and merge because refactoring PRs like this one can cause a lot of merge conflicts if they linger.

ismael-mendoza:
Thanks @jeff-regier looks like a useful PR !

PR #288, opened 2021-08-03
Improvements to Sphinx

dereklhansen:


dereklhansen:
Read the docs theme example:
image

dereklhansen:
My rough plan for what I intend to do:
- [  ] Change theme to the better-looking (IMO) read the docs, which has wider margins and better color-coding.
- [  ] Overwrite all the method docstrings so that we don't see the Pytorch lightning defaults - these take up a lot of space and aren't that informative.
- [  ] Set up some .rst files showing the syntax on how to run different commands of the package.
- [  ] Use (enforce?) Google-style docstrings throughout the package
- [  ] Figure out how to deploy to the github.io website

pyproject.toml
```diff
@@ -17,6 +17,7 @@ python = "^3.8"
 pytorch-lightning = "^1.0.8"
 scipy = "^1.4.1"
 torch = "1.7.1"
+sphinx-rtd-theme = "^0.5.2"
```
jeff-regier, line 20: I'd think of this as a development dependency

jeff-regier:
Sounds like a great plan. What do think of adding static type decorations too? Those sort of function as documentation too. (They may also let us perform static analysis with mypy.)

pyproject.toml
dereklhansen: That's right - fixed that

dereklhansen:
> Sounds like a great plan. What do think of adding static type decorations too? Those sort of function as documentation too. (They may also let us perform static analysis with mypy.)

I can do that as well - it makes more sense than specifying types in the docstring

dereklhansen:
Since it will be a big endeavor to document everything, I'll start by documenting the "datasets" module since I'm most familiar with it, and then get feedback from the rest of the group.

dereklhansen:
image

dereklhansen:
I'm actually a fan of just hiding the Pytorch Lightning methods altogether from Sphinx by giving them empty docstrings. The point of Sphinx is to document the **interface**, not the implementation, for classes and functions. Someone should be able to read the Sphinx page and learn how to initialize and run the object (which is what we have below, although the exact calling structure is out of date c.f. #291)

image

dereklhansen:
Regarding #291 and #293, we may want to use pytest's functionality for testing docstrings as well. This can be a separate PR

dereklhansen:
Going to merge this now since it touches a lot of files. I'll add some other things discussed in separate PRs

PR #297, opened 2021-08-31
Enforce Google docstring style and consistency using flake8-docstrings and darglint

dereklhansen:


dereklhansen:
This PR enforces docstrings in two different ways:
- pydocstyle (via flake8-docstrings): This checks the style of each docstring to make sure it conforms to the Google spec.
- darglint: This checks that, if there are arguments/parameters documented, that they match the function signature.

I set darglint to only check docstrings for which an "Arguments" section exists ("strictness=long"). For consistency, we may want to set "strictness=short" to check all docstrings except those which are a single line.

PR #300, opened 2021-09-01
Binary classifier

ismael-mendoza:


jeff-regier:
Looks like there's duplicated code in binary.py and sleep.py.

jeff-regier (approved):
Amazing

PR #310, opened 2021-09-14
Consolidate config & Remove FluxNet

### LINKED ISSUE #254: consolidate config files
The number of config files is really large which makes it hard to make changes to the configuration. We should review them and only keep the ones that are strictly necessary, using overrides when wanting to customize them.

ismael-mendoza:
Closes #254 

I realized having most of them is useful, but the organization was a bit lacking and made them hard to modify. This PR fixes this. I think they can potentially be refactored more but this is OK for now. 

I also remove `flux_net` since we decided it was better to just train the orignial model for more epochs


ismael-mendoza:
Just to be certain I'm running in the GPU then I will merge

PR #311, opened 2021-09-14
Dual autoencoder

ZachariahPang:
~The current version is ugly. I'm going to clean it up~
To reproduce the result I used in the last meeting, run`bliss mode=train model=galaxy_net dataset=sdss_galaxies optimizer=adam training.n_epochs=1000`


ZachariahPang:
@ismael-mendoza Would you give me some pointers on how should the new dual autoencoders architecture be incorporated in `decoder.py`? 
https://github.com/prob-ml/bliss/blob/0c4cf98b798b2bb30a810301f8e148dd3759798c/bliss/models/decoder.py#L762-L767

ZachariahPang:
Should I close this one since #323 is up?

PR #314, opened 2021-09-15
Use PSF at the center of the frame

ismael-mendoza:
I realized that I was running prediction on the Stripe-82 frame using models trained on a PSF close to the edges rather than at the center of the frame.  I don't think this will change results drastically but it's for the sake of consistency in the paper.

Clean some unused code in `galsim_galaxies` since we are not using the SDSS PSF as input anymore. 

I also fixed a small pytorch_lightning warning related to the number of batches. 

# Takeaways

- Based on #317 , I'm using the fitted PSFile parametric PSF for convolution with galaxies.
- Some of the results in the SDSS chunk (aggregate metrics notebook, precision specifically) seems +- 10% sensitive to the sleep phase training model. There is a correlation with validation_loss but it's not perfect, this is expected as our model is not 100% like the real data (perhaps this is an avenue for future work). 
- For now I included the sleep model that had the lowest validation loss in all the runs but the level of precision is not the same we had before (the classification accuracy improved). I'm in the process of creating more fine-grained metrics as a function of magnitude etc. and I think we should look at those before casting a judgement on the performance of the overall algorithm.


ismael-mendoza:
Just need to check uploaded models give comparable results in SDSS frame then will merge

PR #315, opened 2021-09-15
Refactor hydra configs

ZachariahPang:
### sep 22
- use [`hydra.utils.instantiate`](https://hydra.cc/docs/advanced/instantiate_objects/overview/) to instantiate objects directly instead of relying on the extra layer of `name` and `kwargs` in the configuration files for
  - dataset
  - models
  - trainer: `callbacks`, `logger`, and `profiler` are overwritten directly when instantiating `Trainer` instead of creating a new config dict
- Create two new nested config groups `decoder` and `encoder` within `config/model` to include the relevant configurations.  Those can be merged with `model` config, similar to how different config groups are merged in `config/default.yaml`, using 
```{yaml}
defaults:
  - decoder: sdss_standard
  
  ... normal configs below
 ```
- `sleep.SleepPhase` now takes in instantiated `decoder` and `encoder` objects instead of instantiating them with their kwargs. Both are instantiated and passed to `sleep.SleepPhase` by Hydra
----
### sep 29
- since `sleep.SleepPhase` takes in the `encoder` and `decoder` object during initialization and we should probably not save the entire objects in hyperparameter, we need to supply the `encoder` and `decoder` object when loading from checkpoint since those two objects are not part of the checkpoint (their `state_dict` are in checkpoint). The immediate impact is that during predict, we also need to add `model=sleep_sdss_detection` to load the correct configs for the `encoder` and `decoder` in hydra
 - add an `experiment` config group to run common training task so that instead of 
```{bash}
# sdss_autoencoder.ckpt
poetry run bliss mode="train" model="galaxy_net" dataset="sdss_galaxies" optimizer="adam" \
training.n_epochs=1001 training.trainer.check_val_every_n_epoch=10 dataset.kwargs.noise_factor=0.01 \
dataset.kwargs.num_workers=5 training.trainer.checkpoint_callback=True
```
we can do
```{bash}
poetry run bliss +experiment=sdss_autoencoder
```
with all the above config saved and tracked in a `config/experiment/sdss_autoencoder.yaml` file which is only loaded with `+experiment=sdss_autoencoder` and not used by Hydra otherwise **(note the additional `+` sign since `experiment` is not a default config group)**

---
### oct 4
- Completely disable `Hydra` from managing working dirs and logging`stout`. `Pytorch-lightning` logger is not affected
- Change all absolute paths to relative paths


ismael-mendoza (commented):
This looks great so far @ZachariahPang , I noticed you only changed this for datasets (which is already very helpful). 

Do you think there is a way we can structure the model configs so that we can instantiate the models in the same way? 🤔 

I'm thinking specifically of `binary_sdss.yaml` which has the `decoder` in the config just so that it can get passed in to the dataset config but it's not actually part of the `BinaryEncoder` model. For the other model configs, we could just get rid of the separate `decoder` dict entirely. Let me know what you think and thanks for working on this!

ZachariahPang:
@ismael-mendoza, thank you!. `binary_sdss.yaml` is exactly the problem I'm having. Other ones are very straightforward since their `kwargs.decoder_wargs: {$model.decoder.kwargs}` just pointed to their own `decoder.kwargs`.

For `binary_sdss.yaml`, the possible ways are:
1. add a `**kwargs` in the corresponding `__init__` to take the extra `decoder` configs but this would then pollute the local namespace.
2. package everything related to the actual model into another key in the config so that we can instantiate it with `instantiate(cfg.models.some_new_key)`. This would then create an extra layer in our config not needed for most of the other model config which goes against our intention for this PR
3. use a new nested config in the existing `config/model` and then use the corrected one if needed and `none.yaml` when the model doesn't need a decodoer
```
|--- config/model/
|            |--- binary_sdss.yaml
|            |--- galaxy_net.yaml
|            |--- etc
|            |--- decoder/
|            |              |--- decoder_1.yaml
|            |              |--- decoder_1.yaml
|            |              |--- none.yaml
```
4. Move `decoder`'s in the model configs to the dataset configs

I think 4 and 5 are the preferred way to go

---
Further, for the trainer, callbacks, model checkpoints, we can move them into their new config group similar to `config/optimizer` instead of setting them up in `train.py`

ismael-mendoza:
> I think 4 and 5 are the preferred way to go

What is 5 ?

ZachariahPang:
> > I think 4 and 5 are the preferred way to go
> 
> What is 5 ? Maybe you meant 3. and 4. ?

Sorry, I meant 3. and 4.

ismael-mendoza:
I was thinking about 4. briefly but I think that because in the future we might want to train the decoder (this is what @dereklhansen was working on with the fnp) in the wake phase I think we want to keep it as a model. 

Perhaps the ideal solution would be to do 3. , I think hydra has a way to specify multiple config groups : https://hydra.cc/docs/patterns/select_multiple_configs_from_config_group. So we could do for example: 

```
poetry run bliss mode='train' model=['sleep_sdss_detection', 'decoder/decoder_1'] ... 
```

and then we just enforce somehow that the first model is the one trained and the second one is the decoder if needed.

ismael-mendoza:
> Further, for the trainer, callbacks, model checkpoints, we can move them into their new config group similar to config/optimizer instead of setting them up in train.py

This sounds great!

dereklhansen:
I like this approach a lot! I also agree that option 3 looks the best, especially because recursive initialization is possible: https://hydra.cc/docs/advanced/instantiate_objects/overview/#recursive-instantiation

ZachariahPang:
@ismael-mendoza, I think now the unittest should be passing. I'm running another round for `sdss_sleep` since the loss for the current ckpt is not below -0.09, but the refactoring code should be ready. Two API changes are noted below [**sep 29**](https://github.com/prob-ml/bliss/pull/315#issue-997545598). Would you mind take a look and see if those changes make sense? Thanks

bliss/models/galaxy_encoder.py
```diff
@@ -50,17 +50,18 @@ def center_ptiles(
...
         self,
+        decoder,
         hidden: int = 256,
-        decoder_kwargs: dict = None,
         optimizer_params: dict = None,
     ):
         super().__init__()
-        self.save_hyperparameters()
+        # prevent saving `decoder`
+        self.save_hyperparameters("hidden", "optimizer_params")
```
ismael-mendoza, line 59: is there a way we can still save decoder input arguments? are those still being saved elsewhere?

bliss/predict.py
```diff
@@ -75,8 +76,14 @@ def predict(cfg: DictConfig):
...
-    sleep_net = SleepPhase.load_from_checkpoint(cfg.predict.sleep_checkpoint)
-    galaxy_encoder = GalaxyEncoder.load_from_checkpoint(cfg.predict.galaxy_checkpoint)
+    encoder = instantiate(cfg.model.encoder)
+    decoder = instantiate(cfg.model.decoder)
+    sleep_net = SleepPhase.load_from_checkpoint(
+        cfg.predict.sleep_checkpoint, encoder=encoder, decoder=decoder
+    )
+    galaxy_encoder = GalaxyEncoder.load_from_checkpoint(
+        cfg.predict.galaxy_checkpoint, decoder=decoder
+    )
```
ismael-mendoza, line 86: I see that now you need to instantiate the encoder/decoder separately when loading the checkpoint. The encoder that you pass in to `SleepPhase` is not trained, will the `sleep_net.image_encoder` be the trained version from the checkpoint?

ismael-mendoza: Could you also explain why you need to pass these two arguments at all? I would imagine that the checkpoint saved these models too, but maybe I'm missing something. Thanks!

bliss/tune.py
```diff
@@ -5,31 +5,29 @@
...
 def sleep_trainable(search_space, cfg: DictConfig):
     # set up the config for SleepPhase
-    cfg.model.kwargs.encoder_kwargs.channel = search_space["channel"]
-    cfg.model.kwargs.encoder_kwargs.hidden = search_space["hidden"]
-    cfg.model.kwargs.encoder_kwargs.spatial_dropout = search_space["spatial_dropout"]
-    cfg.model.kwargs.encoder_kwargs.dropout = search_space["dropout"]
+    cfg.model.encoder.channel = search_space["channel"]
+    cfg.model.encoder.hidden = search_space["hidden"]
+    cfg.model.encoder.spatial_dropout = search_space["spatial_dropout"]
+    cfg.model.encoder.dropout = search_space["dropout"]
```
ismael-mendoza, line 22: why are these lines in `tune.py` not covered by tests? Shouldn't `test_tune.py` cover them?

models/README.md
```diff
@@ -7,48 +7,41 @@ in this directory. Please also specify most recent **validation loss** and numbe
...
+poetry run bliss +experiment=sdss_galaxy_encoder
 ```
 
 * ``sdss_sleep.ckpt``
 
 ```bash
-# Validation Loss = -0.102
-# Epochs = 1349
+# Validation Loss = -0.042
+# Epochs = 1499
```
ismael-mendoza, line 40: This is significantly higher than -0.102, are you rerunning it now?

models/README.md
```diff
@@ -7,48 +7,41 @@ in this directory. Please also specify most recent **validation loss** and numbe
...
 # Notes:
 # - Optimization can be a bit unstable and not always reach < -0.09 level (which seems to be significant cutoff)
 # - Results on SDSS Stripe 82 frame vary +- 10% depending on optimization (at least current metrics), might become
 # clear if differences are significant once we make metrics as a function of magnitude. Another possibility
 # is that the sleep phase data model misspecification?
-poetry run bliss mode="train" model="sleep_sdss_detection" dataset="default" \
-optimizer.kwargs.lr=1e-4 training.n_epochs=1501 training.trainer.checkpoint_callback=True \
-model.decoder.kwargs.mean_sources=0.03 training.save_top_k=10
+poetry run bliss +experiment=sdss_sleep
```
ismael-mendoza, line 46: oh this is great 💯

models/README.md
```diff
@@ -7,48 +7,41 @@ in this directory. Please also specify most recent **validation loss** and numbe
...
-# Validation Loss = 6.571
-# Epochs = 359
+# Validation Loss = 7.472
+# Epochs = 439
 # Notes:
 # - 20/09/21: Slight increase since galaxy PSF exactly equals decoder PSF
-poetry run bliss mode="train" model="binary_sdss" dataset="default" \
-optimizer.kwargs.lr=1e-4 training.n_epochs=501 training.trainer.checkpoint_callback=True \
-model.decoder.kwargs.mean_sources=0.03 training.save_top_k=5
+# - 20/09/29: Slightly worse than the previous one (0.571)
```
ismael-mendoza, line 22: what is `0.571` ? what are the reported accuracies in tensorboard? Can you maybe paste the plot produced in tensorboard here?

config/model/sleep_star_basic.yaml
```diff
@@ -1,32 +1,8 @@
...
-    hidden: 185
-  annotate_probs: False
+defaults:
+  - decoder: star_basic
+  - encoder: star_basic
+
+_target_: bliss.sleep.SleepPhase
+
+optimizer_params: ${optimizer}
+annotate_probs: True # FIXME: True or False?
```
ismael-mendoza, line 8: You can set it to `False`

ismael-mendoza (changes requested):
Hi @ZachariahPang - thanks so much for this refactoring PR. I can see all your changes being very useful down the line. I think we can merge soon - I just have a couple of minor comments mostly having to do with the pre-trained models. Please merge after resolving them.

bliss/tune.py
ismael-mendoza: Don't worry too much if it doesn't it just seemed strange to me

ZachariahPang: I think this is related to #236. It would occasionally show as not covered due to RayTune's multiprocessing.

models/README.md
ZachariahPang: I tried one more time and it was -0.044. Does this look normal and we should just retrain it with #323, or does it suggesting something bigger?

models/README.md
ZachariahPang: Sorry, this was probably me misreading. I'll delete this line.

models/README.md
ismael-mendoza: That is suspicious, let me reran the model on master and see what happens

ismael-mendoza: Hi Ziteng, thanks for checking this. I find this strange so I will run this in the master branch and investigate this further

ZachariahPang:
@dereklhansen Sorry for taking so long. I was trying to merge it before Friday but found several issues harder than I imagined.  @ismael-mendoza helped me fixed most of them but there is still one issue that might be more related to the overall design with no immediate clear solution.

The core issue is that our `decoder` needs the following fields to be able to correctly construct. Those are automatically saved by `self.save_hyperparameter()` like what we have been doing on the master branch. https://github.com/prob-ml/bliss/blob/b3722ff39546ac7394fd5a20b782491c1877eb6f/config/model/decoder/sdss_standard.yaml#L12
However, on the master branch, it is saved as the unresolved version `${paths.data}/sdss/94/1/12/psField-000094-1-0012.fits` so it's not specific to any machine that it was originally trained on. Now, because the whole model config is handled by `hydra.utils.instantiate()` which resolves interpolations, it becomes something like `zitengp/bliss//sdss/94/1/12/psField-000094-1-0012.fits` that is specific to my machine. This is the reason why this PR is not passing `test_predict.py` because it was trying to load the file from my local path. 

I looked through Hydra's doc and it seems like there isn't a option to partially resolve the config. I tried using `_recursive_=False` and it behave the same. I think there are maybe 3 choices, each has pros and cons

1. Use relative path: One of the reasons we used absolute path in the past is that Hydra was configured to create new layers of working dir upon each new run and we didn't have a fixed entry point but I think both are not an issue now. The only thing is that we have to find a way to resolve the path in `RayTUne` as it creates new subdirs for each run as well. I not sure if there is a built-in solution for `RayTune` to do that. If not, then the solution might be ugly since we might need to manually parse path from strings in our config passed in.
2. Revert to the approach in `e550fc4`: We pass the initialized `encoder` and `decoder` object to `SleepPhase` without saving the objects and their constructing arguments with `self.save_hyperparameters()`. We make sure to select the correct model config along with the ckpt and use the model configs to constructive `encoder` and `decoder`. This might be as close as we can get to what we initially planned for this PR. However, it then would have the potential risk of using the wrong pair of model config and ckpt that wasn't there before as mentioned in https://github.com/prob-ml/bliss/issues/328#issuecomment-932623117.    
3. Close this PR without merging: This was initially trying to clean things up but it may end up creating more hassle because of the issue above.

jeff-regier:
Solutions 1 & 2 both sound good to me. @ismael-mendoza @dereklhansen which do you prefer?

Either way, let's get this PR merged quickly so there aren't a lot of merge conflicts. The PR doesn't need to be perfect, it just needs to make the codebase better than it was. Things can always be further improved in a subsequent PR.

ZachariahPang:
@ismael-mendoza I just added the changes we discussed earlier and I think this is ready to merge.

ismael-mendoza (approved):
Amazing work @ZachariahPang! I run some tests and this all works great now. 🎉 

A couple of small notes for future changes: (I'll open an issue)
- Now that everything is relative paths, the code only runs when the cwd is the local bliss repo (even when inside the poetry shell) .
- Sometimes it would be convenient to log the dataset use into hyper-parameters / tensorboard. Is there a way to do that easily ? (save the whole `cfg` rather than only the `cfg.model`)
- What does the line in `config.yaml` : `- override hydra/job_logging: stdout` do ? Is is part of disabling the hydra logging?
- Maybe get rid of unusued paths in the config, if there are any?

PR #316, opened 2021-09-16
add code to load images in decoder

ismael-mendoza:
This will enable us to use the PSF at the center of the stripe 82 frame rather than the one obtained from the PSF fit parameters at `psField-000094-1-0012.fits` in the image decoder. See image below for differences 

(first PSF image is from decoder)
Screen Shot 2021-09-16 at 3 07 38 PM


ismael-mendoza:
This is the difference after the changes
Screen Shot 2021-09-16 at 12 23 04 PM

PR #322, opened 2021-09-22
Identify blends in real SDSS data

ismael-mendoza:
Extract reconstruction to create figures in paper that illustrate BLISS performance on real blends. 

It appears there are not many blends in the stripe-82 frame I'm using. Here is the most dramatic blend I could find: 

image

Here are other candidates: 

image

image


jeff-regier:
Nice!

PR #323, opened 2021-09-24
Fully Convolutional Galaxy Autoencoder

dereklhansen:
This builds on #311 by replacing the encoder/decoder networks with a fully-convolutional approach. So far, residuals seem to be improved.


dereklhansen:
Fast facts:
- The bottleneck for each network (main and residual) is 64 dimension (2x2x16); this implies a total bottleneck of 128
- There are 4 downsampling blocks, followed by 4 upsampling blocks; these are called ResidualConvDownsampleBlock and ResidualConvUpsampleBlock
- There are residual skip connections around each upsample/downsample block. These are accomplished by bilinear interpolation of the input to either downscale it or upscale it. To match the number of channels, on downsample blocks, channels are copied. On upsample blocks, we keep only the first half of the channels of the input before adding.
- Within each ResidualConvDownsampleBlock, there are residual skip connections after each convolution+ReLU.
- Within each ResidualConvDownsampleBlock, the number of channels are expanded at the beginning, then reduced back down (similar to what was described in NVAE).

Here's what the residuals look like on the latest version (as of adc58ae):

image

dereklhansen:
Comparison of max residual for adc58ae:

image

dereklhansen:
Grid (as of adc58ae):

image

dereklhansen:
Lowest achieved max residual was 1.601:

image

dereklhansen:
I'll start some runs of the original models so we can have a full comparison

dereklhansen:
Original galaxy autoencoder:

image

dereklhansen:
Original dual autoencoder: best max residual was 2.0
image

dereklhansen:
Image for original dual autoencoder:

image

image

dereklhansen:
There's some qualitative differences between the original and new dual autoencoder, but not a lot of daylight between them. Moreover, the latent dimension of the original is half that of the new, so that could explain any marginal improvement in the new one.

dereklhansen:
Updated results after running original dual-autoencoder vs new fully-convolutional approach:

## Comparison of max_residual
(version_39 is new, version_40 is original)
image

## Comparison of predicted images

Both models exhibit patterns on the worst residuals. I would say the pattern for the new residuals look less distinct than those of the original.

### New (fully-convolutional)
image


### Original
image

dereklhansen:
Results for fully-convolutional model, now with bottleneck size 32 (b7e819a). There was also a slight bug where I forgot to add a ReLU at one point, so things might work better now.

The max residual performance is roughly the same as the previous. Here we compare it (version_43) to the original dual autoencoder (version_40).
image

dereklhansen:
Pictures after training sdss_galaxy_encoder look a bit better than before (8366bda):
Comparison of validation loss (95 is this PR, 96 is before):
image


This PR:
image

Before:
image

ismael-mendoza:
This look great, thanks @dereklhansen ! Looks like a substantive improvement. Out of curiosity, how long did they take to train?

dereklhansen:
Image after training sdss_sleep (ef80ff9):

image

dereklhansen:
Now starts the long road to see if we can get all tests passing...

dereklhansen:
All tests pass on 4ce03a3; now going to merge (which will require re-training all model checkpoints)

dereklhansen:
I have the merge figured out; there weren't any real difficulties. To redo the checkpoints, however, I'll have to start retraining after the server restart today.

bliss/models/galaxy_net.py
```diff
@@ -1,174 +1,262 @@
...
-        fig_random = self.plot_reconstruction(images, recon_mean, residuals, mode="random")
-        fig_worst = self.plot_reconstruction(images, recon_mean, residuals, mode="worst")
         if self.logger:
-            self.logger.experiment.add_figure(f"Random Images {self.current_epoch}", fig_random)
-            self.logger.experiment.add_figure(f"Worst Images {self.current_epoch}", fig_worst)
+            self.logger.experiment.add_figure(
+                f"Epoch:{self.current_epoch}/Random Images", fig_random
+            )
+            self.logger.experiment.add_figure(f"Epoch:{self.current_epoch}/Worst Images", fig_worst)
+            self.logger.experiment.add_figure(
```
ismael-mendoza, line 188: Maybe split declaring the title and the `add_figure` command into two lines to avoid multi-line commands?

bliss/models/galaxy_net.py
```diff
@@ -197,26 +288,210 @@ def plot_reconstruction(
...
+
+
+class OneCenteredGalaxyEncoder(nn.Module):
+    def __init__(self, main_encoder, main_decoder, residual_encoder, slen=None, allow_pad=False):
+        super().__init__()
+        self.main_encoder = main_encoder
+        self.main_decoder = main_decoder
+        self.residual_encoder = residual_encoder
+        self.slen = slen
+        self.allow_pad = allow_pad
```
ismael-mendoza, line 351: Could you explain why you have this padding? or what exactly does `allow_pad = True` does?

bliss/models/galaxy_net.py
```diff
@@ -197,26 +288,210 @@ def plot_reconstruction(
...
+        self.residual_decoder = residual_decoder
+
+    def forward(self, latent):
+        latent_main, latent_residual = torch.split(latent, latent.shape[-1] // 2, -1)
+        recon_mean_main = F.relu(self.main_decoder(latent_main))
+        recon_mean_residual = self.residual_decoder(latent_residual)
+        return F.relu(recon_mean_main + recon_mean_residual)
+
+
+class CenteredGalaxyEncoder(nn.Module):
```
ismael-mendoza, line 383: It might be cool to have some documentation for `CenteredGalaxyEncoder`, `CenteredGalaxyDecoder` and `ResidualConvBlock` just explaining what they do at a high level and descriptions of some of the more unintuitive arguments. A couple of comments here and there for some of the more complicated constructions might be helpful. 

I don't think this should stop you from merging, it would just help me understand better how you construct this neural network and some of the modelling choices you made, which would also be helpful for writing the paper :)

bliss/utils.py
```diff
@@ -139,3 +140,42 @@ def forward(self, mean_z, logscale_z):
         if self.minscale is not None:
             logscale_z = torch.log(self.minscale + (1 - self.minscale) * F.softplus(logscale_z))
         return Normal(mean_z, logscale_z.exp())
+
+
+# adapt from torchvision.utils.make_grid
+@torch.no_grad()
+def make_grid(
```
ismael-mendoza, line 147: Can you add documentation just explaining the output of this function (or how it is different from `torchvision.utils.make_grid`, anything else that is important?)

bliss/utils.py
```diff
@@ -139,3 +140,42 @@ def forward(self, mean_z, logscale_z):
         if self.minscale is not None:
             logscale_z = torch.log(self.minscale + (1 - self.minscale) * F.softplus(logscale_z))
         return Normal(mean_z, logscale_z.exp())
+
+
+# adapt from torchvision.utils.make_grid
+@torch.no_grad()
```
ismael-mendoza, line 146: Does this basically wrap the function call in a `with torch.no_grad` block?

config/experiment/sdss_galaxy_encoder.yaml
```diff
@@ -18,3 +18,6 @@ model:
 optimizer:
   kwargs:
     lr: 1e-4
+
+dataset:
+  batch_size: 8
```
ismael-mendoza, line 23: newline missing here 
```suggestion
  batch_size: 8
  
```

models/README.md
```diff
@@ -1,47 +1,33 @@
 # Models
+-  All checkpoints were generated with `./scripts/train_all.sh`.
+-  Results from the latest run are in `./scripts/results`.
+-  This README should be regularly updated with any notes on training
```
ismael-mendoza, line 4:
```suggestion
-  This README should be regularly updated with any notes on significant changes to model training or performance.
```

tests/test_galaxy_autoencoder.py
```diff
@@ -2,6 +2,7 @@ def test_galaxy_autoencoder_toy_gaussian(model_setup, devices):
     use_cuda = devices.use_cuda
     overrides = {
         "model": "galaxy_net",
+        "model.residual_delay_n_steps": 50 if use_cuda else 0,
```
ismael-mendoza, line 5: Could you modify this tests so that in the GPU it ensures that we achieve the higher performance you observed after training? You could either check the `max_residual` or the loss itself (whatever seems more stable).

ismael-mendoza (changes requested):
Hi @dereklhansen, thanks a lot for this excellent PR and it's very nice to see the performance increase. I think it's really close to merging, I just had a couple of questions and comments below. Feel free to let me know if anything is unclear, etc. Thanks!

bliss/models/decoder.py
```diff
@@ -133,10 +134,23 @@ def __init__(
                 self.autoencoder_ckpt,
             )
             # load dataset of encoded simulated galaxies.
-            self.register_buffer("latents", torch.load(latents_file))
+            if self.latents_file.exists():
+                latents = torch.load(self.latents_file, "cpu")
+            else:
```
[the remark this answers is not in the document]
dereklhansen: I think this will (hopefully) be replaced by #339.

bliss/models/galaxy_net.py
```diff
@@ -1,174 +1,262 @@
...
-        loss = self.get_loss(images, recon_mean)
-        self.log("train/loss", loss)
+        if optimizer_idx == 0:
+            recon_mean_main = self._main_forward(images, background)
+            loss = self._get_likelihood_loss(images, recon_mean_main)
+            self.log("train/loss_main", loss, prog_bar=True)
+        if optimizer_idx == 1:
+            with torch.no_grad():
+                recon_mean_main = self._main_forward(images, background)
+            recon_mean_residual = self._residual_forward(images - recon_mean_main)
```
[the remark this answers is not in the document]
dereklhansen: Yes, that's right - it cancels out in the subtraction

bliss/models/galaxy_net.py
```diff
@@ -1,174 +1,262 @@
...
+        self.residual_autoencoder = nn.Sequential(self.residual_encoder, self.residual_decoder)
 
-        self.register_buffer("zero", torch.zeros(1))
-        self.register_buffer("one", torch.ones(1))
+        self.mse_residual_model_loss = mse_residual_model_loss
+
+        self.residual_delay_n_steps = residual_delay_n_steps
+        assert slen == 53, "Currently slen is fixed at 53"
+        self.slen = slen
+        assert latent_dim == 64, "Currently latent_dim is fixed at 64"
```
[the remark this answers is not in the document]
dereklhansen: I agree this is ugly. Right now the code is not adaptable to arbitrary galaxy "slen" or latent dimension sizes. The encoder consists of five downsampling layers, which take the image down to a `(32,1,1)` size representation. Thus, this will only work for images within a certain slen. Moreover, since I saw reduced performance adding linear layers after, there's no way to tweak the latent dimension size.

I'll try to think of a less ugly way to do this. Likely it will be:
-  Make the number of downsample layers depend on "slen". Specifically, add downsample layers until the image is reduced to `(2**L,1,1)`, where `L` is the number of layers.
-  Set `latent_dim` equal to `2*L`.

bliss/utils.py
dereklhansen: I think @ZachariahPang can better answer this question

bliss/utils.py
dereklhansen: I believe so; this is only ever used for plotting and might be memory-intensive with gradient tracking

bliss/models/galaxy_net.py
dereklhansen: The reason I add padding is because the galaxy encoder passes images of size 44x44, while the encoder needs an image of size 53x53. 

I'll write a docstring for this class so this reason is documented in the code

bliss/models/galaxy_net.py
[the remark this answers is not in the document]
dereklhansen: Made this a new issue in #343, because it's not causing problems in our pipeline right now, but it should be addressed. I'll likely have to refactor the code some.

dereklhansen:
Thanks for the review @ismael-mendoza; I implemented most of the changes you suggested. The rest I've opened as new issues since they aren't urgent and can be solved with smaller PRs.

PR #339, opened 2021-10-08
Turn galaxy autoencoder into a VAE

dereklhansen:
This adds functionality to use a normalizing flow to learn the distribution of latent variables induced by the trained galaxy autoencoder.


dereklhansen:
Here's an example of sampled galaxies from a very basic flow that I literally copied from the front page of the "nflows" repo:

image

dereklhansen:
After adding standardization layer and making flows deeper:
image

ismael-mendoza:
Interesting! Most of these images look pretty good except of a couple of malformed ones. Did you already try turning the AE into a VAE (by optimizing kl-qp, etc.) ?

dereklhansen:
Updates: I've revisited the training procedure for the VAE to improve things. 

- First, I raised the noise factor in the generated galaxies to 1.0. Before, it was at 0.01. I wonder if this was the source of some issues. 
- I lowered the latent dimension from 64 (32+32) to 16 (8+8).
- I added Weight Normalization to all of the conv nets in the model. This really improved training for the 16-dimensional case

In total, the 16-dimension VAE paired with weight norm trains as well as the 64-dimension VAE:
image

dereklhansen:
The residuals for Random and Worst images are the best I've seen yet:

image

image

dereklhansen:
Things look good for the galaxy encoder on most images (as a VAE), but occasionally it will miss really bad and make a galaxy overly bright. I can't say why this is happening.

image

dereklhansen:
The problem has been fixed by training the galaxy encoder weights from scratch. Before, we initialized at the weights for the normal galaxy vae

image

ismael-mendoza:
> The problem has been fixed by training the galaxy encoder weights from scratch. Before, we initialized at the weights for the normal galaxy vae

Is the galaxy encoder probabilistic also or deterministic?

dereklhansen:
> Is the galaxy encoder probabilistic also or deterministic?

It's probabilistic now

dereklhansen:
I believe this is ready to go, but I'm not sure how to properly check the performance on real data. @ismael-mendoza can you point me to the scripts where SDSS predictions are run?

jeff-regier:
^ @ismael-mendoza

ismael-mendoza:
> I believe this is ready to go, but I'm not sure how to properly check the performance on real data. @ismael-mendoza can you point me to the scripts where SDSS predictions are run?

Yeah the easiest would just be to run the script I was talking about last meeting that reproduces all the plots in one go. I'll try to get this done soon. In the meantime I'll go ahead and review the code

bliss/models/galaxy_net.py
```diff
@@ -65,30 +67,42 @@ def __init__(
...
         self.residual_decoder = CenteredGalaxyDecoder(
-            slen=slen, latent_dim=latent_dim, n_bands=n_bands
+            slen=slen, latent_dim=latent_dim // 2, n_bands=n_bands
         )
         self.residual_autoencoder = nn.Sequential(self.residual_encoder, self.residual_decoder)
 
+        self.register_buffer("prior_mean", torch.tensor(0.0))
+        self.register_buffer("prior_var", torch.tensor(1.0))
+        self.dist_main = Normal(self.prior_mean, self.prior_var)
+        self.dist_residual = Normal(self.prior_mean, self.prior_var)
```
ismael-mendoza, line 93: does this defeat the purpose of `ensure_dist_on_device` or viceversa?

bliss/models/prior.py
```diff
@@ -87,10 +84,9 @@ def __init__(
 
         self.prob_galaxy = float(prob_galaxy)
         if prob_galaxy > 0.0:
-            latents = get_galaxy_latents(latents_file, n_latent_batches, autoencoder_ckpt)
+            self.vae = galaxy_net.OneCenteredGalaxyAE.load_from_checkpoint(autoencoder_ckpt)
```
ismael-mendoza, line 87: at some point we should change the class name and checkpoint names from AE to VAE :)

tests/test_galaxy_autoencoder.py
```diff
@@ -2,7 +2,7 @@ def test_galaxy_autoencoder_toy_gaussian(model_setup, devices):
     use_cuda = devices.use_cuda
     overrides = {
         "model": "galaxy_net",
-        "model.residual_delay_n_steps": 50 if use_cuda else 0,
+        "+model.residual_delay_n_steps": 50 if use_cuda else 0,
```
ismael-mendoza, line 5: what is the purpose of this flag? does it delay using the residual part of the VAE when training until after 50 steps?

models/sdss_binary_results
```diff
@@ -0,0 +1,3 @@
+sdss_binary
+epoch=489-val_loss=1.565.ckpt
```
ismael-mendoza, line 2: It's interesting the loss was 3 times smaller than before. It'd be interesting to see some of the images or what the precision/recall graphs from tensorboard look like. Don't worry though if you didn't save these  files

tests/test_galaxy_autoencoder.py
ismael-mendoza: also, why did you remove it from the config? (so now you have use the `+`)

case_studies/sdss_galaxies/Galaxy Flow.ipynb
```diff
@@ -0,0 +1,248 @@
+{
```
ismael-mendoza, line 1: Do you think these two notebooks are useful to keep in the repo? Just from skimming them it seems that these notebooks demonstrate that the VAE and the flow galaxy samples look reasonable right?

bliss/models/galaxy_net.py
```diff
@@ -65,30 +67,42 @@ def __init__(
...
         self.min_sd = min_sd
 
     def forward(self, image, background):
         """Gets reconstructed image from running through encoder and decoder."""
-        recon_mean_main = self._main_forward(image, background)
-        recon_mean_residual = self._residual_forward(image - recon_mean_main)
-        return recon_mean_main + recon_mean_residual
+        recon_mean_main, pq_latent_main = self._main_forward(image, background)
+        recon_mean_residual, pq_latent_residual = self._residual_forward(image - recon_mean_main)
+        return recon_mean_main + recon_mean_residual, pq_latent_main + pq_latent_residual
```
ismael-mendoza, line 105: If this is now a VAE, should it also return the variance?

bliss/models/galaxy_net.py
```diff
@@ -110,50 +136,68 @@ def generate_latents(self, dataloader, n_batches):
             for _ in tqdm(range(n_batches)):
                 galaxy = next(iter(dataloader))
                 noiseless = galaxy["noiseless"].to(self.device)
-                latent_batch = enc(noiseless, 0.0)
+                latent_batch, _ = enc(noiseless, 0.0)
```
ismael-mendoza, line 139: why use the noiseless version of the images here? Doesn't the encoder assume imatges have noise ?

bliss/models/galaxy_net.py
```diff
@@ -110,50 +136,68 @@ def generate_latents(self, dataloader, n_batches):
...
 
         # metrics
         residuals = (images - recon_mean_final) / torch.sqrt(recon_mean_final)
         residuals_main = (images - recon_mean_main) / torch.sqrt(recon_mean_main)
         recon_mean_residual = recon_mean_residual / torch.sqrt(recon_mean_main)
         self.log("val/max_residual", residuals.abs().max())
+
+        # Aggregate posterior
+        u_main = self.main_encoder(images - background)
+        u_residual = self.residual_encoder(images - recon_mean_main)
```
ismael-mendoza, line 191: what are the `u_main` and `u_residual` ?

bliss/predict.py
```diff
@@ -77,7 +77,7 @@ def predict_on_image(
     assert image_encoder.border_padding == galaxy_encoder.border_padding
     assert image_encoder.tile_slen == galaxy_encoder.tile_slen
 
-    galaxy_param_mean = galaxy_encoder(ptiles, tile_map["locs"])
+    galaxy_param_mean, _ = galaxy_encoder(ptiles, tile_map["locs"])
```
ismael-mendoza, line 80: Related question: does the galaxy_encoder now return a `galaxy_param_mean` and `galaxy_param_var` or am I missing something?

ismael-mendoza (changes requested):
Hi @dereklhansen thanks for this PR! I think it looks pretty good, see below for a couple of (mostly) questions and some suggestions. 

I think it would be nice to discuss some of these on Thursday before merging, in the meantime I will try to get my script to produce the plots with BLISS on real data merged so that we can test the VAE and see if we get comparable results.

models/sdss_binary_results
dereklhansen: image

dereklhansen: image

case_studies/sdss_galaxies/Galaxy Flow.ipynb
dereklhansen: No, I don't want to clutter the repo with these - I must have committed them by accident. I went ahead and removed them

bliss/models/galaxy_net.py
dereklhansen: It returns a single sample and the difference between the prior and variational density for that sample (pq_latent_*). I kept it this way to require less changes from what we had before. I will likely refactor this when we want to draw multiple samples at once (#363).

bliss/predict.py
dereklhansen: Right now galaxy_encoder returns a single sample from the variational distribution (and the associated density ratio of that sample) to minimize changes to the rest of the code. I'll likely refactor this in #363 by having the encoder forward method return a "Normal" object (which can be sampled, have the mean and var extracted, etc.)

dereklhansen: Since this isn't the mean, I'll rename it

bliss/models/galaxy_net.py
dereklhansen: These are samples from the aggregate posterior (i.e. latent space samples from the real data). I renamed them to be more clear

bliss/models/galaxy_net.py
dereklhansen: Good catch - this was left over from before. I fixed it to use the images and background, but I may just delete this method since we don't use it anymore.

dereklhansen: I actually went further and just deleted the method since we don't use it anymore.

bliss/models/galaxy_encoder.py
```diff
@@ -78,6 +80,10 @@ def __init__(
         # will be trained.
         autoencoder_ckpt = decoder["autoencoder_ckpt"]
         autoencoder = OneCenteredGalaxyAE.load_from_checkpoint(autoencoder_ckpt)
+        if from_scratch:
+            autoencoder = OneCenteredGalaxyAE(
+                slen=autoencoder.slen, latent_dim=autoencoder.latent_dim
+            )
```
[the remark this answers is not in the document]
dereklhansen: Right now, probably not, but we could use it if we are training a wake-phase update down the line.

bliss/models/galaxy_net.py
```diff
@@ -110,50 +136,68 @@ def generate_latents(self, dataloader, n_batches):
...
             self.log("train/loss", loss, prog_bar=True)
+            self.log("train/loss_recon", loss_recon, prog_bar=True)
         return loss
 
     def validation_step(self, batch, batch_idx):
         """Validation step (pytorch lightning)."""
         images, background = batch["images"], batch["background"]
-        recon_mean_main = self._main_forward(images, background)
-        loss_main = self._get_likelihood_loss(images, recon_mean_main)
+        recon_mean_main, p_latent_main = self._main_forward(images, background)
```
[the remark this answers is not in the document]
dereklhansen: No, it was a typo - I corrected it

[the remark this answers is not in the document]
dereklhansen: I think this points to a underlying issue of code duplication in the VAE. We can probably refactor so we don't calculate the training/validation loss on different code paths

bliss/models/galaxy_net.py
dereklhansen: For some reason these registered buffers weren't getting properly moved to the device when the prior is inside a data loader...

dereklhansen: It's ugly but I couldn't find a better way to get it to work

config/experiment/sdss_autoencoder.yaml
```diff
@@ -7,11 +7,11 @@ defaults:
...
 training:
-  n_epochs: 8001
+  n_epochs: 10001
   trainer:
     check_val_every_n_epoch: 10
     checkpoint_callback: true
 
 dataset:
-  noise_factor: 0.01
+  noise_factor: 1.0
```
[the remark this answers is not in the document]
dereklhansen: I can remove it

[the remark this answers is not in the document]
dereklhansen: I'll make this a separate issue since it's orthogonal to the VAE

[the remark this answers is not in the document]
ismael-mendoza: sounds good, thank you

bliss/models/galaxy_net.py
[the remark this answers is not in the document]
ismael-mendoza: yes that makes sense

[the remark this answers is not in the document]
ismael-mendoza: I don't think it should stop you from merging now though

bliss/models/galaxy_net.py
ismael-mendoza: I see, yeah I suspected something like that... It's a weird issue

tests/test_galaxy_autoencoder.py
ismael-mendoza: > I removed the flag because I think we should only specify arguments in the Hydra configs when we explicitly want to override the defaults from the "target" class.

Good point, otherwise the default in the config and the default of the class might get out of sync

models/sdss_binary_results
ismael-mendoza: That's interesting the accuracy definitely increased (it was hovering around 0.95 before). Not sure what caused it though

ismael-mendoza: I think we should look at the accuracy in real data, I'll get that working before this afternoon

ismael-mendoza: Thanks for posting the images!

bliss/predict.py
ismael-mendoza: Sounds good, thanks for clarifying. I think some of the questions I had about these changes now make a lot more sense and we can merge once we test on real data.

bliss/models/prior.py
dereklhansen: I agree - I opened up a separate issue since I want to update the names of all the modules in the "galaxy_net.py" file

dereklhansen:
I updated to include changes in #374; fingers crossed the real images look good!

ismael-mendoza:
I'll reproduce the images sometime today and post them here

dereklhansen:
Wow, these results are not good at all... I'll have to investigate

[sdss_reconstruction1.pdf](https://github.com/prob-ml/bliss/files/7528443/sdss_reconstruction1.pdf)

dereklhansen:
The problem seems to be going from the autoencoder to the VAE. I re-ran the same image on the current master branch and it looks a lot better
[sdss_reconstruction1_ae.pdf](https://github.com/prob-ml/bliss/files/7528465/sdss_reconstruction1_ae.pdf)
:

dereklhansen:
My working theory is that the VAE isn't learning a good generative model. I removed the normalizing flow to simplify things for this PR, but it might be necessary to re-incorporate it.

The aggregate posterior for the VAE at the end of training doesn't look normal, so that means there's a mismatch between samples from the VAE and Galsim:

image

ismael-mendoza:
> My working theory is that the VAE isn't learning a good generative model. I removed the normalizing flow to simplify things for this PR, but it might be necessary to re-incorporate it.

Seems like this could be affecting things for sure. Another thing I wanted to note about the reconstructions you sent above is that most of the dim/small galaxies get predicted as circular (apart from the most visible issue that bright ones have their flux/shape misspecified )

dereklhansen:
Aggregate posterior before flow transform:
image

Aggregate posterior after:
image

dereklhansen:
The aggregate posterior pre-normalizing flows does not resemble a standard normal. In addition to the shape being wrong, the scale is also bigger than a standard normal. This means that the samples we were generating for sleep phase training were under-dispersed. Thus, it makes sense why we did so poorly on real data; our models weren't getting the full range of galaxies.

On the other hand, fitting the generative latent distribution to match the aggregate posterior leads to noise which much more resembles standard normal, meaning the samples from this generative model will display a larger range of galaxies. 

Because we are fitting KL(q || p), samples from the flow-augmented model will be overdispersed relative to the aggregate posterior. This is probably a good thing for training the model. However, this inevitably means we'll see some galaxies in training that were not in Galsim. This is why an eyeball test of samples can be misleading. The previous, underdispersed galaxies all look like realistic galaxies, while the flow-augmented, overdispersed galaxies contain some weird looking ones.

dereklhansen:
Previous samples (underdispersed):
image

New samples (overdispersed):
image

dereklhansen:
It's not really clear from just looking at images what the big picture of the galaxy model is. It might be a good idea to check the distribution of functionals of galaxies and compare the VAE+flow to Galsim.

models/sdss_binary_results
```diff
@@ -0,0 +1,3 @@
+sdss_binary
+epoch=229-val_loss=5.414.ckpt
```
dereklhansen, line 2: @ismael-mendoza After retraining binary with a VAE+flow, the loss is around where it was before. So I think it was a red flag that the loss went down a lot.

dereklhansen:
Interesting; when we re-train the prior as a flow distribution, the results aren't better, but they are different. Left is original VAE, right is new VAE:
image

dereklhansen:
Now, the recall is a lot better but with much worse precision, especially for brighter galaxies. This makes sense if the generative model for galaxies is too dispersed; it could span objects which look like stars.

dereklhansen:
We may need to look beyond the vanilla VAE training objective for ideas.
- [ ] WAE-MMD
- [ ] WAE-GAN
- [ ] AE-MMD (WAE where we use autoencoder without noise)
- [ ] AE-GAN

jeff-regier:
> if the generative model for galaxies is too dispersed; it could span objects which look like stars.

Even if the galaxy model can generate star-like objects, it should still be less likely to do so than a star, which exclusively generates star-like objects.

jeff-regier:
Are these precision/recall numbers for star/galaxy classification, or for detections?

ismael-mendoza:
> Are these precision/recall numbers for star/galaxy classification, or for detections?
@jeff-regier : Detections

PR #348, opened 2021-10-12
Add functions to evaluate models on SDSS

### LINKED ISSUE #283: move code from `case_studies/sdss_galaxies` to bliss codebase
- [ ] looping over chunks and getting/saving variational parameters should be done in `predict.py`
- [ ]  confusion matrix and precision/recall could be used by other parts of the code, so maybe useful to put in a file called `reporting.py`

ismael-mendoza:
Closes #283 

Next PR will simplify code for reconstructions, search for new blends and make it easy to plot them.

- [x] Functions in `predict.py` to get variational parameters on all tiles on all scene and map estimates on full scene
- [x] functions on `report.py` to load results from predict and report metrics `precision, recall, f1, classification accuracy, confusion matrix`
- [x] (maybe) functions on `report.py` to create plots of metrics as a function of magnitude and example reconstructions

### DIFF
--- a/case_studies/sdss_galaxies/aggregate-results.ipynb
+++ b/case_studies/sdss_galaxies/aggregate-results.ipynb
@@ -14,8 +14,8 @@
    "id": "c9334866",
    "metadata": {
     "ExecuteTime": {
-     "end_time": "2021-10-19T20:14:10.264778Z",
-     "start_time": "2021-10-19T20:14:10.253505Z"
+     "end_time": "2021-10-20T02:35:23.919882Z",
+     "start_time": "2021-10-20T02:35:23.909330Z"
     }
    },
    "outputs": [],
@@ -30,8 +30,8 @@
    "id": "b0413b12",
    "metadata": {
     "ExecuteTime": {
-     "end_time": "2021-10-19T20:14:10.406782Z",
-     "start_time": "2021-10-19T20:14:10.384867Z"
+     "end_time": "2021-10-20T02:35:24.063186Z",
+     "start_time": "2021-10-20T02:35:24.043635Z"
     }
    },
    "outputs": [
@@ -59,8 +59,8 @@
    "id": "e7123298",
    "metadata": {
     "ExecuteTime": {
-     "end_time": "2021-10-19T20:14:10.819381Z",
-     "start_time": "2021-10-19T20:14:10.542710Z"
+     "end_time": "2021-10-20T02:35:24.465247Z",
+     "start_time": "2021-10-20T02:35:24.207799Z"
     }
    },
    "outputs": [],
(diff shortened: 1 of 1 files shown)


bliss/predict.py
```diff
@@ -27,107 +52,190 @@ def prediction(image, image_encoder, galaxy_encoder=None, binary_encoder=None):
...
+                    # locations in pixels consistent with full scene.
+                    # 0.5 comes from pt/pr definition (plotting both -> off by half a pixel).
+                    x = est_locs[:, 1].reshape(-1, 1) * clen + x1 - 0.5
+                    y = est_locs[:, 0].reshape(-1, 1) * clen + y1 - 0.5
+                    est_locs = torch.hstack((x, y)).reshape(-1, 2)
+
+                    # get fluxes from galaxy parameters by passing them though galaxy_decoder.
+                    latents = est_params["galaxy_params"].to(device).reshape(-1, latent_dim)
+                    single_galaxies = galaxy_decoder(latents)
+                    est_galaxy_flux = single_galaxies.sum((-1, -2, -3)).cpu().reshape(-1)
```
ismael-mendoza, line 173: slack on bins of magnitude

bliss/predict.py
```diff
@@ -5,19 +5,44 @@
...
 from bliss.sleep import SleepPhase
 
 
-# TODO: Go back here: https://gist.github.com/ismael-mendoza/85e6c245a177c82b071557d9d097ebc0
-def prediction(image, image_encoder, galaxy_encoder=None, binary_encoder=None):
+def predict_on_image(
+    image: torch.Tensor,
+    image_encoder: encoder.ImageEncoder,
+    binary_encoder: BinaryEncoder,
+    galaxy_encoder: GalaxyEncoder,
```
ismael-mendoza, line 20: consider merging the 3 models into a model.

bliss/report.py
```diff
@@ -0,0 +1,102 @@
+"""Functions to generate metrics on real data."""
+import torch
```
jeff-regier, line 2: Would it be OK to include the code from metrics.py and plot.py into report.py (perhaps better called reporting.py), so we can delete those two files, and keep all the reporting code together?

ismael-mendoza: Yes, I like this idea

ismael-mendoza:
Here is a plot of the metrics as a function of magnitude. They get a bit noisy towards the left because the number of objects becomes small, but the overall trend makes sense.

Screen Shot 2021-10-19 at 6 16 00 PM

Just for our reference here is a plot of the true and estimated magnitudes: 

Screen Shot 2021-10-19 at 6 19 54 PM

Screen Shot 2021-10-19 at 6 19 56 PM

PR #358, opened 2021-11-01
Separate "ImagePrior" from "ImageDecoder"

### LINKED ISSUE #356: Refactor prior sampling decoder.py
Right now, I'd say `decoder.py` is the most grueling part of the code base. A big part of this is that the "ImageDecoder" class has a lot of attributes:
https://github.com/prob-ml/bliss/blob/99c54fd5c4be20828ae0a72d53026bb9c04d24e4/bliss/models/decoder.py#L28-L55

Many of these attributes pertain to the prior distribution:
https://github.com/prob-ml/bliss/blob/99c54fd5c4be20828ae0a72d53026bb9c04d24e4/bliss/models/decoder.py#L164-L183

I think would help to encapsulate the prior into separate objects. In particular, star-specific and galaxy-specific prior methods could be placed into their respective classes "StarTileDecoder" ("_get_fluxes", "_pareto_cdf", and "_draw_pareto_mixed") and "GalaxyTileDecoder" ("_sample_galaxy_params"). Just this change alone would cut down many of the attributes

dereklhansen:
This closes #356


dereklhansen:
- [ ] Retrain sleep phase to fix error on prediction
- [ ] Any flake8, pylint errors

PR #364, opened 2021-11-05
Use F.fold within decoder for image reconstruction

dereklhansen:
nn.Fold (https://pytorch.org/docs/1.7.1/generated/torch.nn.Fold.html?highlight=fold) is designed for stitching together a large canvas from subimages that overlap. 

This PR replaces our previous reconstruction method, which used nested for-loops within Python. The biggest advantage is performance and scalability, since I believe nn.Fold takes advantage of the GPU. Moreover, this will simplify the most complex part of decoder.py.


bliss/models/decoder.py
```diff
@@ -349,6 +246,62 @@ def _validate_border_padding(self, border_padding):
...
+    # In default settings of ImageDecoder, no borders are cropped from
+    # output image. However, we may want to crop
+    max_padding = (ptile_slen - tile_slen) / 2
+    assert max_padding % 1 == 0
+    max_padding = int(max_padding)
+    crop_idx = max_padding - border_padding
+    return folded_image[:, :, crop_idx : (-crop_idx or None), crop_idx : (-crop_idx or None)]
+
+
+def calc_output_size(
```
ismael-mendoza, line 296: This function should not be public right? Since it really will only be used inside `reconstruct_image_from_ptiles` ? Maybe add an underscore or put it inside `reconstruct_image_from_ptiles`?

ismael-mendoza (approved):
This is so much cleaner :) Thanks!

bliss/models/decoder.py
```diff
@@ -349,6 +246,62 @@ def _validate_border_padding(self, border_padding):
...
+
+    folded_image = F.fold(image_ptiles_prefold, output_size, kernel_size, stride=stride)
+
+    # In default settings of ImageDecoder, no borders are cropped from
+    # output image. However, we may want to crop
+    max_padding = (ptile_slen - tile_slen) / 2
+    assert max_padding % 1 == 0
+    max_padding = int(max_padding)
+    crop_idx = max_padding - border_padding
+    return folded_image[:, :, crop_idx : (-crop_idx or None), crop_idx : (-crop_idx or None)]
```
ismael-mendoza, line 293: Why do you need the `or None` ? Is that in case crop_idx = 0 ?

dereklhansen: Yes, the "or None" will kick in when crop_idx=0, which means we don't want to crop anything.

bliss/models/decoder.py
dereklhansen: > Just as sanity checks, did you check if all tests are passing in the GPU ? Have you looked at some of the images by eye just to ensure they look ok / similar to before ?

I ran all tests ensuring that "torch.allclose()" never failed for the original and new method, but I didn't do it on the GPU. I'll make sure tests pass on GPU before merging.

dereklhansen:
> Just as sanity checks, did you check if all tests are passing in the GPU ? Have you looked at some of the images by eye just to ensure they look ok / similar to before ?

I ran all tests ensuring that "torch.allclose()" never failed for the original and new method, but I didn't do it on the GPU. I'll make sure tests pass on GPU before merging.

ismael-mendoza:
> I ran all tests ensuring that "torch.allclose()" never failed for the original and new method, but I didn't do it on the GPU. I'll make sure tests pass on GPU before merging.

Sounds good if you checked it with `torch.allclose()` you can probably merge. It wouldn't hurt to run in the GPU just to make sure. Thanks!

dereklhansen:
This passed all GPU tests (other than multi GPU, which was taking hours so I cancelled it)

PR #374, opened 2021-11-11
Reproduce all figures with a single script

ismael-mendoza:
- [x] remove notebooks from case_studies
- [x] script `main.py` in `case_studies/sdss_galaxies` to reproduce all figures easily via `./main.py --fig all` 
- [x] To recreate cache data use the `--overwrite` flag.


ismael-mendoza:
Here are the latest figures compiled in a single PDF: 

https://www.dropbox.com/s/e9dqxzj7njgdshs/latest-bliss-figures-11-11-21.pdf?dl=0

PR #377, opened 2021-11-11
Remove test_multi_gpu

### LINKED ISSUE #365: test_multi_gpu.py takes a really long time
I recently ran the test suite on the GPU to test #364, and test_multi_gpu seems to take a really long time. I ended up killing the process since it was taking hours. 

I know that this test will be skipped when there's one GPU (either on the device or set via CUDA_VISIBLE_DEVICES). Do we use the multi-GPU functionality right now?

ZachariahPang:
close #365


ismael-mendoza:
Thanks!

PR #380, opened 2021-11-19
Create "Predict" torch module that encapsulates encoding real image

### LINKED ISSUE #352: combine encoder models into one for prediction
In our last meeting, @dereklhansen mentioned that combining the 3 models into one could be beneficial. Currently in `predict.py` we need to manage combining the output of each of the 3 models consistently into a single output, while it feels like it would be better to manage this logic elsewhere i.e. predict.py should just have to call `model.predict()` given a `model`. 

- [ ] Ensure that galaxy parameters returned do not correspond to stars like it's done in `predict.py`

dereklhansen:
This creates a new class called "Predict" that encapsulates encoding an astronomical image. Closes #352. 

The input to "Predict" is a whole image or smaller chunk, and the output is the estimated locations, fluxes, classifications, and variational parameters. This can be further broken down into three distinct actions: 

- Predict.locate_objects() (ImageEncoder)
- Predict.classify_objects() (BinaryEncoder)
- Predict.encode_galaxies() (GalaxyEncoder)

It's important to break up these methods, since we only want to do the first two actions (locate and classify objects) when retraining the GalaxyEncoder on real data.

For reconstruction, this has synergy with what we call right now "ImageDecoder". We pass the estimated/sampled latent values from the "Predict" class into "ImageDecoder" to get back the whole image.


dereklhansen:
We may not want to call this module "Predict"; "ImageEncoder" sounds more apt, but that name is already taken. I suggest renaming what is now called "ImageEncoder" to something like "LocationEncoder", but that might be better in another PR.

dereklhansen:
The functionality from this will be in #391; for now we leave the predict module separate from the new "encoder" module.

PR #386, opened 2021-12-13
Refactor ImageEncoder

dereklhansen:


dereklhansen:
I've noticed that the term `var_params` is overloaded in usage throughout the code. It's used 1) for the output of the encoder network and 2) for samples from the encoded distribution. I think (2) should be called something else (like "latents", "samples", etc)

dereklhansen:
Things that remain:
- [ ] Investigate get_full_params()
- [ ] Add docstrings to everything public in encoder.py (do this last since things may change 

Regarding `get_full_params`, this method needs shapes in a particular way, which adds complexity to the other code surrounding it. I want to look into how this could be made easier.

PR #389, opened 2022-01-04
Refactor get_full_params() 

dereklhansen:
This refactors `get_full_params()` by splitting the functionality into smaller methods to disentangle what it does. Hopefully this will make it easier to maintain.

Another key change is that `get_full_params()` takes in only the `tile_slen` and infers the other variables (`slen`, `wlen`, etc). Before we were using `slen`, `wlen`, and the tile parameters to infer `tile_slen`. The latter was more complicated because we needed to check a lot of assertions (i.e. that `n_tiles_per_image` divided evenly into `slen**2`).

### DIFF
--- a/bliss/models/binary.py
+++ b/bliss/models/binary.py
@@ -5,7 +5,12 @@ from torch import nn
 from torch.nn import BCELoss
 
 from bliss.models.decoder import get_mgrid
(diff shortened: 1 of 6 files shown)


dereklhansen:
I'll leave this up for a little bit, but I need it for the real-data training so I'll merge it likely Thursday (assuming tests are passing)

PR #391, opened 2022-01-05
Train galaxy encoder on real SDSS images

dereklhansen:


dereklhansen:
Illustration of how the training currently works:

Each chunk effectively has "double borders". WIthin the outer border (indicated with red), the encoder will not make any detections, and these pixels are masked in the loss function. Within the inner border (indicated with green), the encoder will make detections, but these pixels are also masked in the loss function.

We also allow for adjusting the amount of overlap between chunks. This ensures that every galaxy in a scene will end up within the two borders. I've circled in yellow to illustrate how the same objects show up multiple times in training.

bliss_real_data_encoder

PR #396, opened 2022-01-21
Refactor Hydra configuration to be more robust and modular

### LINKED ISSUE #372: Reorganize hydra configs and reduce duplication
Example: 
```
config/dataset/toy_gaussian.yaml
8:noise_factor: 0.1

config/dataset/sdss_galaxies.yaml
6:noise_factor: 0.01

config/experiment/sdss_autoencoder.yaml
16:  noise_factor: 1.0

bliss/datasets/galsim_galaxies.py
28:        noise_factor=0.05,
46:        self.noise_factor = noise_factor
88:        noise = torch.sqrt(image) * torch.randn(*image.shape) * self.noise_factor
114:        noise_factor=0.05,
137:        self.noise_factor = noise_factor
217:        noise = image.sqrt() * torch.randn(*image.shape) * self.noise_factor
```

dereklhansen:
Resolves #372.


dereklhansen:
There are a couple of reasons for this PR:
- We currently have it so different "models" are mutually exclusive; i.e. we either load settings for "SleepPhase" or settings for "GalaxyEncoder", etc. This makes it impossible to access the configuration that was used to construct the other models. We've been avoiding this issue by using the model checkpoints, but this means that the configuration is now hidden in binary checkpoints. Without changing this config, we are effectively "locked in" to using the model checkpoints. 
- We have a lot of config files which are the same name ("m2.yaml", "sdss.yaml") in the folder for each model ("encoder", "prior", "decoder"). In reality, we should have one yaml file per experimental setup that configures all of the models for it. These yaml files can then live in "case_studies".

dereklhansen:
Checklists

Get the following "training" experiments running:
- [ ] sdss_autoencoder
- [ ] sdss_binary
- [ ] sdss_galaxy_encoder
- [ ] sdss_sleep
- [ ] sdss_galaxy_encoder_real
Get the following "generate" experiments running
- [ ] sdss_individual_galaxies

Other things
- [ ] Get the other non-SDSS galaxy experiments working (e.g. m2)
- [ ] Get tests working
- [ ] In config.yaml, make most settings default arguments for the respective objects
- [ ] Use the __target__ syntax for prior, encoder, and decoder
- [ ] Try to rename encoder -> location_encoder

dereklhansen:
Idea: Instead of having overriding training modules for each dataset, we have training.common and training.models.modelName,
then something like
```
cfg_training = OmegaConf.merge(training.common, training.models.[training.model])
instantiate(cfg_training.trainer)
```
Will have to change some of the functions in train.py, but shouldn't be too bad.

PR #402, opened 2022-01-26
Move SDSS Galaxies configuration, saved models, and tests to case_studies

dereklhansen:
This PR will move all configuration, saved state and tests which are directly related to the SDSS galaxies experiment into the `case_studies/sdss_galaxies` folder. 

This will help further modularize the BLISS codebase. Conceptually, we should treat the code in the `bliss/` folder as a library, which experiment-specific applications in `case_studies` all use (e.g. `m2`, `sdss_galaxies`). This applies to future apps (`sdss_galaxies_vae`)

PR #410, opened 2022-02-03
Hydra refactor fixes

dereklhansen:
This PR cleans up some issues pointed out in the meeting today from the recent Hydra refactor #401 and #402


dereklhansen:
@ismael-mendoza Now, upon training, the train function will automatically dump a JSON file alongside the state-dict with extension ".pt.log.json". This contains the validation loss like before, along with the folder where the full training logs are stored (useful for looking at training results with Tensorboard, and seeing who last trained the model, etc.)

image

dereklhansen:
Added a timestamp as well... we can always add more relevant data to this file as we need it
image

ismael-mendoza:
Thanks @dereklhansen !

PR #413, opened 2022-02-03
Remove BLISS_HOME dependency and improve path configuration

### LINKED ISSUE #361: Cannot specify a relative path for output (unless in bliss root directory)
When I run something like this:
```
bliss +experiment=$EXPERIMENT training.save_top_k=1 paths.output=$OUTPUT_DIR training.version=$EXPERIMENT
```
where `$OUTPUT_DIR= ../output/train_all`, I get an error because BLISS changes the working directory to `$BLISS_HOME`. I know this change was made for a reason (#330, #334), but it is confusing from an end-users' perspective.

dereklhansen:
Closes #361. This will greatly simplify the pathing issue, since the need to set BLISS_HOME is delegated out to the specific folder directory. Along with the recent improvements to the Hydra configs, this will make setting different data directories outside the BLISS codebase trivial. (e.g. set "paths.sdss" to "/data/sdss").

To figure out where testing data files are, tests will use their own `__file__` absolute path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,6 @@ poetry install
 poetry shell
 ```
 
-4. Finally, set an environment variable called `BLISS_HOME` to the absolute path of the `bliss` directory.
-
 # Latest updates
 ### Galaxies
    - BLISS now includes a galaxy model based on a VAE that was trained on Galsim galaxies.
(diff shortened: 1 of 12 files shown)


dereklhansen:
I should note that #361 is likely already not an issue since we removed the forced changedir to the bliss root directory. However, this will make sure the bliss *library* (located in `bliss/`) does not need to be any particular working directory. It is the responsibility of the *application* (located in `case_studies/sdss_galaxies`) to provide valid paths for data files, checkpoints, etc.

PR #419, opened 2022-02-09
fixes isort/black clashing

ismael-mendoza:
Fixes very annoying error when doing multi-line imports that basically makes committing impossible. 

Not sure if better to get rid of multi-line imports, but since the functions names are very explicit and long I think this is better.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,8 @@
+[settings]
+profile = black
+multi_line_output = 3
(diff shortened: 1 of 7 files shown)


ismael-mendoza:
@dereklhansen I suspect you turned it off because some of your files did not have the imports sorted (maybe due to do this?) Could you turn it on if possible and let me know if there are other issues? Thanks!

Issue #390, opened 2022-01-04
Replace n_tiles_per_image with n_tiles_h and n_tiles_w


dereklhansen:
Right now, we use a 1-D indexing into the tiles that make up an image. We should split this dimension into a height and width dimension. This will simplify downstream code without adding any performance penalty, because it's effectively just changing the view of the same data. This will also make it simpler to allow for non-square images as input to the encoder.

dereklhansen:
I thought about this while doing #389, but this should be a separate PR.

dereklhansen:
Added a function that will help with this in 1be9aa9dea90409d48dc1b4fe293628a6db5a288.

Issue #375, opened 2021-11-11
Negative flux predictions


ismael-mendoza:
A noticed that a small number (~0.1%) of galaxies can get predicted with a negative flux. The galaxies for which these happens have _very_ low flux to the point that we will never detect them. I think we should remove such dim galaxies from the galaxy sample and this would get rid of this issue 

- [ ] increase the `min_flux` in the `sdss_galaxies.yaml` config and rerun

jeff-regier:
Is this still happening? Shouldn't the galaxy autoencoder end with a relu so that its output is always nonnegative (regardless of how it's trained)?

ismael-mendoza:
I'll pin this issue so I don't forget about it because it is definitely important. @dereklhansen can comment more on this since he was the last one to tweak the autoencoder. I suggest we wait until we decide whether we want to keep the autoencoder at all (following the metrics in #425 ) and then we can just check this on the VAE.

dereklhansen:
The decoded image should always be non-negative in both the autoencoder and the VAE:

https://github.com/prob-ml/bliss/blob/02bf153dd87aa6eedf18eacbc0fcce274c2f9b16/bliss/models/galaxy_net.py#L394-L398

dereklhansen:
But we should double-check that the computed "galaxy_fluxes" from #425 are always non-negative (maybe make this a test?)

ismael-mendoza:
Yes definitely a small test would be a good idea

PR #425, opened 2022-02-10
Use inference not predict

ismael-mendoza:
This will allow Derek to use my detection/classification metrics and produce the same plots but with his VAE.

- [x] add parameters missing into `inference.py` from `predict.py` (galaxy_fluxes)
- [x] remove `predict.py` in `case_studies/sdss_galaxies/plots.py`, use `reconstruct_scene_at_coords` instead
- [x] remove predict.py and test_predict.py
- [x] correct usage (include bp vs not in `h`, `scene_length`, etc.) 
- [x] need to add `mags` to `recon_map` dictionary.


dereklhansen:
This closes #375 by adding an assertion when galaxy fluxes are computed.

case_studies/sdss_galaxies/plots.py
```diff
@@ -142,38 +143,34 @@ def fignames(self):
...
         assert isinstance(scene, (torch.Tensor, np.ndarray))
-        assert sleep_net.device == binary_encoder.device == galaxy_encoder.device
-        device = sleep_net.device
+        assert encoder.device == decoder.device
+        device = encoder.device
 
         bp = 24
-        clen = 300
+        slen = 300  # chunk side-length
         h, w = scene.shape[-2], scene.shape[-1]
```
dereklhansen, line 154: I don't think it will be a hard change to handle rectangular images.

dereklhansen: I can push some changes to this branch to allow it so we don't have to have this "FIXME" here.

case_studies/sdss_galaxies/plots.py
```diff
@@ -142,38 +143,34 @@ def fignames(self):
...
-        _, est_params = predict_on_scene(
-            clen,
-            scene_torch,
-            image_encoder,
-            binary_encoder,
-            galaxy_encoder,
-            galaxy_decoder,
-            device,
+        _, est_params = reconstruct_scene_at_coordinates(
+            encoder, decoder, scene_torch, bp, bp, scene_length, slen, bp, device=device
```
dereklhansen, line 173: I think this will work, but the intention of `reconstruct_scene_at_coordinates` is so you don't have to worry about border padding - you give it the image coordinates (h, w, scene_length) and it will give you back the reconstructed image and parameters.

dereklhansen:
@ismael-mendoza I think the part of `compute_data` after the scene is reconstructed should be refactored into a method that just takes the `est_params` and `coadd_params`.

https://github.com/prob-ml/bliss/blob/4d54eb55ca857e66b9cc31abf50d0da2ad4e67f6/case_studies/sdss_galaxies/plots.py#L135-L149

https://github.com/prob-ml/bliss/blob/4d54eb55ca857e66b9cc31abf50d0da2ad4e67f6/case_studies/sdss_galaxies/plots.py#L176-L205

```python
def catalog_metrics(est_params, coadd_params):
        mag_bins = np.arange(18, 23, 0.25)  # skip 23
        precisions = []
        recalls = []
        class_accs = []
        galaxy_accs = []
        star_accs = []
        for mag in mag_bins:
            res = reporting.scene_metrics(
                coadd_params, est_params, mag_cut=mag, slack=1.0, mag_slack=0.5
            )
            precisions.append(res["precision"].item())
            recalls.append(res["recall"].item())
            class_accs.append(res["class_acc"].item())


            # how many out of the matched galaxies are accurately classified?
            galaxy_acc = res["conf_matrix"][0, 0] / res["conf_matrix"][0, :].sum()
            galaxy_accs.append(galaxy_acc)


            # how many out of the matched stars are correctly classified?
            star_acc = res["conf_matrix"][1, 1] / res["conf_matrix"][1, :].sum()
            star_accs.append(star_acc)


        return {
            "mag_bins": mag_bins,
            "precisions": precisions,
            "recalls": recalls,
            "class_accs": class_accs,
            "star_accs": star_accs,
            "galaxy_accs": galaxy_accs,
        }
```

dereklhansen:
Hi @ismael-mendoza 

I think I got metrics working on the VAE, but I had to make some hacky edits to the coadd catalog so I put it in a different branch on top of this one (#430). The metrics looked decent at a quick glance (71% precision, 58% recall for faintest magnitude)

dereklhansen:
I think we need to fix #390 before getting this to work on non-square images. I'll work on that in a separate PR since it is a relatively big change

case_studies/sdss_galaxies/plots.py
dereklhansen: I merged in #433 so this can be fixed

ismael-mendoza:
Thanks @dereklhansen for your help with this PR! Here are some plots that I produce with the recently reverted AE code and `plots.py`.

Some notes: 

- The 'worst' reconstructions are gain between +5 and -5 with the familiar 8-pattern. This makes sense because the old architecture had this issue. I don't think this is a problem for the paper and BLISS's "future work" fixes this :) 
- The random reconstructions still look pretty good.
- Scatter true vs pred w/ contour plot: Mostly unchanged, small # of outliers close to edge of ellipticities and HLR is expected
- Scatter residual binned plot: g1 and g2 now look exactly the same yay! this was actually a bug I had to look into but never resolved. The magnitude looks to be systematically biased but the error is of order 0.1% which probably doesn't matter? 
- Size estimation looks funky, but it's just because of sparsity in bins (and outliers are less < 1%)

Some actions items from this 

- [ ] investigate dip in precision with an interactive plotting to ensure that indeed we are detecting things PHOTO is not
- [ ] think about magnitude systematic bias (probably just a footnote in the paper) 

# AE Single Galaxy Figures

random_reconstructions
worst_reconstructions
single_galaxy_measurements_contours
single_galaxy_scatter_bins



# SDSS Reconstructions
sdss_reconstruction0
sdss_reconstruction1
sdss_reconstruction2
sdss_reconstruction3


# SDSS Metrics
sdss-classification-acc
sdss-precision-recall

ismael-mendoza:
I'll merge now since plots are reproduce well and without predict.py. Follow up investigate the dip and make plots better pending recommendations in the meeting.

PR #426, opened 2022-02-11
Standardize param names

ismael-mendoza:
I think we should standardize param names throughout the code base to prevent confusion. I already spent time debugging two related bugs while working in #425 so I think this we should do this sooner rather than later.

I propose to rename all parameters as plurals and using `galaxy_` or `star_` prefix. See examples in checklist below. Moving forward we should try to keep this convention when possible. I'm happy to consider other conventions that you think might be better.

- [x] rename `galaxy_param` -> `galaxy_params`
- [x] rename `galaxy_bool` -> `galaxy_bools` 
- [x] rename `prob_galaxy` -> `galaxy_probs` 
- [x] rename `star_bool` -> `star_bools`
- [x] correct error in `test_reconstruct.py`
- [x] ~~recreate `coadd_cat` using correct names~~ -> actually will keep names of coadd separate from parameters but might revisit later if it becomes confusing. 
- [x] ~~Merge after #425  and #424 to prevent merge conflicts.~~


ismael-mendoza:
@dereklhansen any ideas what could be causing this? Seems to be coming from  a `galaxy_bools` misshapen tensor but all other test pass, so I'm confused why this one would be different

dereklhansen:
The issue is with the "galaxy_params" key. I think the extra dimension is getting added somewhere around here:
https://github.com/prob-ml/bliss/blob/cdd6a4ffb8f3048e5073b3229f139fad9648f12a/bliss/encoder.py#L93-L113

dereklhansen:
I'm going to replace the ".reshape()" with einops rearrange to make the intention more clear

ismael-mendoza:
You are right, it was the `galaxy_param` sorry for the confusion! In that case I think I will merge this before my own #425

dereklhansen (approved):

PR #432, opened 2022-02-14
Create TileCatalog class in place of using dictionaries

dereklhansen:
This creates a TileCatalog class, which subclasses the dictionary. It adds additional validation and checking over a standard dictionary. Moreover, many functions that were solely concerned with converting tile catalogs into full catalogs can be made methods of this new class.


ismael-mendoza:
Thanks for working on this @dereklhansen ! This is a great idea, feel free to ping me for a review if you feel that might be useful.

dereklhansen:
Will likely wait to work on this until #458 is finished, since the data structure is still evolving

dereklhansen:
- [ ] Place data structures for `TileCatalog` and `FullCatalog` in `bliss/catalog.py` since they are more widely applicable than just the `LocationEncoder.py`

dereklhansen:
@ismael-mendoza I know this is a big PR, but I think it would be good if you could skim over the changes. In particular, the changes in `reporting.py` should give you a good feel for the new `FullCatalog` class. Most of the assertions there are unnecessary now since they are checked upon initializing the FullCatalog.

ismael-mendoza:
Thanks @dereklhansen - I'll take a look now :)

bliss/models/location_encoder.py
```diff
@@ -31,223 +33,6 @@ def output_channels(self, input_channels: int) -> int:
         return 2 * input_channels
 
 
-def get_images_in_tiles(images, tile_slen, ptile_slen):
```
ismael-mendoza, line 34: This is great! 🎉

case_studies/sdss_galaxies_vae/reconstruction.py
```diff
@@ -66,13 +66,13 @@ def reconstruct(cfg):
...
         )
-        map_recon = get_full_params_from_tiles(tile_map_recon, encoder.tile_slen)
+        map_recon = tile_map_recon.to_full_params()
         map_recon["fluxes"] = (
             map_recon["galaxy_bools"] * map_recon["galaxy_fluxes"]
             + map_recon["star_bools"] * map_recon["fluxes"]
         )
         map_recon["mags"] = convert_flux_to_mag(map_recon["fluxes"])
-        map_recon["plocs"] = map_recon["plocs"] - 0.5
+        # map_recon["plocs"] = map_recon["plocs"] - 0.5
```
ismael-mendoza, line 75: Why don't you need the `-0.5` anymore ?

ismael-mendoza (approved):
Looks really good, definitely simplified a lot of things we did over and over 🎉 

I skimmed most of the code and couldn't find anything that seemed incorrect

case_studies/sdss_galaxies_vae/reconstruction.py
dereklhansen: I think I accidently included this change from #484 - I was experimenting trying to align the semi-synthetic coadd catalog. I'll remove it.

PR #434, opened 2022-02-17
Replace BatchNorm with weight norm in location and binary encoders

dereklhansen:


dereklhansen:
- [ ] Pass tests with GPU on M2 
- [ ] Retrain binary and sleep on sdss_galaxies
- [ ] Retrain binary and sleep on sdss_galaxies_vae

dereklhansen:
So far, just replacing BatchNorm with weight norm everywhere leads to very poor performance on the M2 example

ismael-mendoza:
That's unfortunate, maybe we do need batchnorm after all :( 

Maybe you have other ideas to improve weight norm but if we do have to stick with batchnorm we could create a custom version that doesn't trip us up

dereklhansen:
> That's unfortunate, maybe we do need batchnorm after all :(
> 
> Maybe you have other ideas to improve weight norm but if we do have to stick with batchnorm we could create a custom version that doesn't trip us up

It turns out batch norm wasn't the problem with the different output depending on image size. So I think it's fine to leave it for now.

PR #436, opened 2022-02-17
Fix dependency on batch size by making image_min a parameter

dereklhansen:
Using background values in #435 leads to performance degradation on the m2 cluster. Effectively, using the minimum value of the image is like a batch normalization. This is an alternative change that forces the user to pass the image min to the "forward()" method when in eval mode.


dereklhansen:
Closing in favor of the other approach in #435

PR #442, opened 2022-02-19
Resurrect old galaxy autoencoder architecture

dereklhansen:
This PR resurrects the old galaxy autoencoder architecture before I added the changes in #323. Specifically, we use the architecture in the commit 382fccfd51f617150d9cfd595272d86b48c88d89.

There are a few reasons for exploring this change. It's clear that there is a performance regression for the regular autoencoder on the real SDSS data. When I implemented #323, I hadn't yet run the full pipeline, so these changes weren't optimized for real-data performance. For @ismael-mendoza's paper, I think it is prudent to restore the architecture to where it was before. It seems to be more robust to moving to real-data.

Second, I want to see how well the VAE does with minimal changes to the original architecture. I've got some promising initial results just by doubling the dimension of the encoder and applying `F.softmax` to the second-half. I'd like to see how well this does on real data.

This doesn't mean that the new architecture is going away. Rather, I'd like to add back the new architecture improvements one-by-one and only keep them if they improve performance on both real data and single galaxies.


dereklhansen:
Here's the results on the autoencoder (not the VAE):
sdss-precision-recall
sdss-classification-acc
sdss_reconstruction3
sdss_reconstruction2
sdss_reconstruction1
sdss_reconstruction0

dereklhansen:
On the AE, we see that there's poor performance on galaxies with magnitudes between 19.5 and 21.5. I think this is because many galaxies are being mis-classfied as stars. Likely caused by #435, because nothing else should have changed in the estimation procedure.

dereklhansen:
Actually, the AE results are better than I thought. On #435 I just used a constant background value (`865.0`) in `predict.py` since we were intending on removing that code. After passing the real SDSS background values, detections are pretty good now!
sdss-precision-recall
sdss-classification-acc
sdss_reconstruction3
sdss_reconstruction2
sdss_reconstruction1
sdss_reconstruction0

dereklhansen:
The VAE is still having the same issue as before:
image

dereklhansen:
These images are from using MAP estimates for galaxy parameters instead of sampling
image
image

ismael-mendoza:
Hi @dereklhansen - I really appreciate you making this PR! This will allow me to get back to producing results a lot faster :) 

I have attached the old metrics below. It seems like the recall improved somewhat (except for a dip around 21.5) and the precision is mostly the same except for that dip around the values you mentioned. Do you understand that dip? Is it because of the background change? Interestingly there was no dip in precision in your first set of plots

Screen Shot 2022-02-20 at 7 12 05 PM

ismael-mendoza:
@dereklhansen - feel free to merge this PR sooner rather than later if this is the only issue remaining and I can help investigate that dip as well.

dereklhansen:
> It seems like the recall improved somewhat (except for a dip around 21.5) and the precision is mostly the same except for that dip around the values you mentioned. Do you understand that dip? Is it because of the background change? Interestingly there was no dip in precision in your first set of plots for the AE

I think the dip in precision is because the LocationEncoder is finding more "objects" that are not detected by SDSS. If you look at the VAE plots, which have the locations plotted, there are several locations where there isn't an SDSS object but we label it as one. You can also see this happening if you look closely at the AE plots; there are a few reconstructions from very faint / invisible objects.

dereklhansen:
I think I finally solved the main problem with the VAE. While padded tiles without galaxies do not factor into reconstruction loss, I was evaluating the divergence loss (`q(z | x) - p(z)`) at *every* padded tile. This was encouraging `q(z|x)` to be close to the prior `p(z)` whenever x was an empty image. This is why the decoder got so unstable when the galaxies were so faint.

dereklhansen:
This fixes the VAE - the only trouble spots are where the LocationEncoder + BinaryEncoder identified blends where there aren't any.

image
image

dereklhansen:
> @dereklhansen - feel free to merge this PR sooner rather than later if this is the only issue remaining and I can help investigate that dip as well.
> 
> [EDIT] Although I see you are investigating things with the VAE so please take your time.

I think at this point the only issue is the dip, so I'll merge this.

PR #459, opened 2022-02-24
Use tile catalog throughout inference and find galaxy blends

### LINKED ISSUE #454: Find new blends automatically
It would be nice to have a way of finding blends in real data automatically. @dereklhansen had a nice idea by looking over neighboring or co-neighboring MAPs of tiles. So far I only have one actual "zoom-up" real data blends for the paper so this way I could find more. (at least 3 would be nice)

dereklhansen:
Closes #454.

### DIFF
--- a/bliss/encoder.py
+++ b/bliss/encoder.py
@@ -130,6 +130,10 @@ class Encoder(nn.Module):
         return self.location_encoder.border_padding
 
     @property
+    def tile_slen(self) -> int:
+        return self.location_encoder.tile_slen
+
+    @property
     def device(self):
         return self._dummy_param.device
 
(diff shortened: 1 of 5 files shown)


dereklhansen:
@ismael-mendoza I implemented some code that detects blends in the Tile catalog. Basically, on the 2-d grid of tiles, you can do a convolution of "galaxy_bool" with the kernel equal to all one. The kernel size determines how close galaxies should be to be considered blends

dereklhansen:
Here's an example from my "reconstruction.py" when the kernel size is 2 (i.e. two galaxies are considered blends when they are in adjoining tiles):

Screen Shot 2022-02-24 at 1 07 51 PM

PR #460, opened 2022-02-25
Utilize square-root of background when log-transforming images

### LINKED ISSUE #450: Replace 100 with sqrt(background)*4 in `subtract_bg_and_log_transform`

dereklhansen:
Closes #450.

### DIFF
--- a/bliss/models/location_encoder.py
+++ b/bliss/models/location_encoder.py
@@ -7,8 +7,22 @@ from torch.distributions import categorical
 from torch.nn import functional as F
 
 
-def subtract_bg_and_log_transform(image, background, slack=100.0):
-    return torch.log1p(F.relu(image - background + slack, inplace=True))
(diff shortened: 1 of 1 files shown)


dereklhansen:
image

dereklhansen:
Quite a boost in performance with this change on real data.

ismael-mendoza:
That's impressive and surprising, I would've guess that we would get more false positives

ismael-mendoza:
Quick question, did you retrain the binary / location encoder after changing this function? This should also affect their training right?

jeff-regier:
If the particular image normalization procedure we apply makes such a difference, perhaps we should apply multiple different normalization procedures (e.g., one that subtracts 4 sds, one that subtracts 3.4 sds), stack the normalized images (one channel each), and pass the whole stack to the location autoencoder. It may help to dedicate two input channels to the background and the unnormalized image too, in case the location autoencoder can figure out how to make use of them.

Issue #461, opened 2022-02-27
Investigate blends in SDSS that we cannot deblend


ismael-mendoza:
I've identified three blends in SDSS that we cannot deblend. We should understand why for the paper I think, at least in terms of limitations even if it turns out the issue is deeper than a simple bug so that we might have to delay solving it.

<img width="766" alt="Screen Shot 2022-02-27 at 4 47 51 PM" src="https://user-images.githubusercontent.com/11745764/155902671-b816da69-f4f4-4bf9-8d3d-be8d137d9c70.png">
<img width="766" alt="Screen Shot 2022-02-27 at 4 47 51 PM" src="https://user-images.githubusercontent.com/11745764/155902674-5ccc29f9-6dfd-4bcf-9b24-ca544d12e561.png">
<img width="822" alt="Screen Shot 2022-02-27 at 4 59 58 PM" src="https://user-images.githubusercontent.com/11745764/155902679-f3b36eb6-64c0-474d-8af9-c32018b75c9f.png">

ismael-mendoza:
Here is one that we can 

<img width="784" alt="image" src="https://user-images.githubusercontent.com/11745764/155902747-742bf90f-fa76-4d04-a91d-430571251ee1.png">

jeff-regier:
Does PR #561 solve some/most of these problems?

ismael-mendoza:
I will retrain and investigate this soon

dereklhansen:
On my end, it doesn't look like BLISS gets these blends after #561.

For the first, BLISS and PHOTO make the same prediction (one galaxy in the center).
For the second, BLISS assigns a 34% detection probability and PHOTO predicts it.
For the third, BLISS assigns a 0% probability and PHOTO does not predict it.

So really only the second blend is something we should maybe be getting, and the good news is that the probability is pretty high

ismael-mendoza:
Thanks @dereklhansen - on my end I find that we are still not getting some of the easy blends. I guess I can take a look at PHOTO and see what it says on the frame but these look like pretty easy to detect. 

![image](https://user-images.githubusercontent.com/11745764/165403646-1d0bdf6e-e7e9-410d-91b3-d2adafd8bb43.png)

![image](https://user-images.githubusercontent.com/11745764/165403803-e7a3da3f-5320-4ae1-a892-1cddd77c7074.png)

jeff-regier:
> For the second, BLISS assigns a 34% detection probability and PHOTO predicts it.

[...]

Issue #451, opened 2022-02-23
Investigate dips in precision and recall metric


ismael-mendoza:
After recent changes from @dereklhansen (mostly like #428) we see changes in the precision metric in SDSS stripe-82 data compared to before. This makes BLISS look bad as Derek mentioned, we should investigate and justify it in the paper.

- [x] pull out examples where BLISS finds a source but coadd does not between ~19.5 and ~21.2
- [x] same but < 19 
- [x] pull out examples of objects BLISS misses between 21 and ~21.5
- [x] pull out some (ra,dec) and send to slack channel
- [x] compare precision/recall with PHOTO on frame rather than coadd.
- [x] Investigate whether any of these issues are due to what @dereklhansen mentioned of two galaxies detected in adjacent tiles? importance sampling procedure can help
- [x] merge #333
- [x] did the 19-20 dip correspond to mean of uniform prior on galaxy fluxes? 
- [ ] check PHOTO on frame catalog is reasonable on image and metrics make sense (ensure metrics are doing OK) - need to actually convert Frame catalogs to correct format, etc. (use FullCatalog class for consistency)
- [ ] use DeCals as truth instead of SDSS coadd (see [here](https://www.legacysurvey.org/dr9/catalogs/#region-tractor-aaa-tractor-brick-fits)) 
- [x] if too complicated, list exclusions one-by-one (painful)
- [ ] change plot to be per (mag truth) bin rather than below cut ? 
- [ ] set mag_slack to 1.0 or higher (compare flux estimation with separate metrics)

<img src="https://user-images.githubusercontent.com/11745764/155419891-8d30cf76-9ed5-47ce-ab2d-5b2dd50c672f.png" width=50% height=50%>

ismael-mendoza:
Interesting variation in curve as a function of mag_slack  (precision then recall). This demonstrates that our flux measurements could be better. The "general" dip features are present throughout though

<img width="563" alt="Screen Shot 2022-02-27 at 5 04 01 PM" src="https://user-images.githubusercontent.com/11745764/155901826-bbb52b3e-bfd1-41f7-ad07-a2583cf52c2d.png">


<img width="557" alt="Screen Shot 2022-02-27 at 5 03 25 PM" src="https://user-images.githubusercontent.com/11745764/155901804-ed41c14c-b3e3-4ac1-b88f-fb2f97e617dd.png">

ismael-mendoza:
[...]

PR #463, opened 2022-03-02
try pareto as prior of galaxy fluxes

### LINKED ISSUE #333: Retrain on dimmer galaxies
I realized that there are basically no galaxies with mag > 19 on original simulated sample of galsim galaxies. We need to retrain including this dimmer galaxies so that we get better performance on the sdss chunk. Also this might solve the issue seen on the scatter bin plot (see below, top left panel) where our AE systematically overestimates fluxes for galaxies

- [x] Look at flux/mag distribution on SDSS catalog so we can try matching it.
- [x] Sample uniformly in magnitude between 16 - 21 ? Sample from magnitudes in the catalog? or some smooth interpolation? use a pareto distribution?
- [x] retrain now that #460 is merged and see if still see improvement
- [ ] retrain with pareto fluxes for galaxies
- [ ] make a scatter plot of matched galaxies/star in frame, BLISS magnitude vs PHOTO coadd magnitude. Should see improvement.
- [ ] check precision/recall. Also improvement? 
- [ ] do we still see systematic downward shift in magnitudes Individual galaxy residual? (remember to include outliers)
- [ ] also check ellipticity plot just in case

image

ismael-mendoza:
Closes #333 

- [x] #490 
- [ ] Sample HLR uniformly, then a/b
- [ ] Account for correlation between size and flux.
- [ ] in `plot_image_and_locs` should be explicit name variable `plocs` or just `locs`? (at least in arguments and docstring maybe?)


ismael-mendoza:
After playing the truncated pareto for a bit this is how it compares with the SDSS catalog fluxes?
Do you have additional suggestions for making them the same @jeff-regier ?  Here is the funtion I used 

```
def pareto_sample(n, f_min=1e3, f_max=1e6, alpha=0.5, factor=2):
    u_max = 1 - (f_min / f_max) ** alpha
    uniform_samples = np.random.rand(n) * u_max
    return (f_min*factor) / (1.0 - uniform_samples) ** (1 / alpha)
```

Screen Shot 2022-03-02 at 9 49 26 AM
s

jeff-regier:
This PR will help, you could try it. I suspect the lack of support below 3.25 flux (on your graph--not sure what the units are here) will be problematic. Not modeling the correlation between size and flux could also be problematic. Seems the simplest thing would be to use an empirical prior, perhaps with a bit of Gaussian noise added to it

ismael-mendoza:
Thanks Jeff, another option is to sample directly from the coadd catalog. It would probably not be desirable to use coadd for the same frame we are predicting but if we could get coadds for other frames that would work.

ismael-mendoza:
Here are some results on retraining the autoencoder on the pareto prior for galaxy fluxes 

image

image


Screen Shot 2022-03-07 at 10 44 20 PM

ismael-mendoza:
Here are the most recent plots after retraining all the models with the pareto prior on galaxy fluxes specified above. We should probably discuss more before merging. Some observations: 

- The binary classification plot took a hit in galaxy accuracy, like 10% of galaxy accuracy was decreased. We should take a look at the confidence though.
- The precision curve is now better and without a dip. It remains constant before 22 magnitude which is interesting. I suspect though that we will see it increasing at high magnitude after masking the saturation. The recall curve remained unchanged so overall it looks like a good change for these curves. I think it might be worth examining by eye some of the previous detections we identified in #451 as false positives and see how this changed those. But also some of the detections we are missing still. 
- The SDSS reconstruction figures still look pretty good overall in terms of residuals, some classifications are more off compared to #425 but we should look at the confidence too.
- In one of the SDSS blends we now classify the object correctly as a galaxy (previously a star). These SDSS blends look basically unchanged. 
- The reconstruction of worst individual galaxies (second figure) suffered, which makes sense because we are no longer "over-training" on only bright galaxies. 
- You can see in the random reconstructions (first figure), that some galaxies are pretty much buried in the noise. They are indeed present in the SDSS sample in terms of their flux, but perhaps we should model the size correlation because some of the most extended ones will be completely invisible. The easiest way I can think of donig this is by matching quantiles between magnitude/flux and HLR which would be good enough I think. 
- The individual galaxy metrics look quite strange. With very faint objects I can see why ellipticity and size would be harder to estimate. Before our sample consistently almost exclusively of bright objects. Especially if trying to reconstrcut some of the "invisible" objects. You can see this in the last two plots. Another idea is to bin residual ellipticity errors by magnitude/SNR. 


random_reconstructions
worst_reconstructions
sdss_reconstruction0
sdss_reconstruction1
sdss_reconstruction2
sdss_reconstruction3
sdss-classification-acc
sdss-precision-recall
single_galaxy_measurements_contours
single_galaxy_scatter_bins

ismael-mendoza:
One of the blends in #461 is now detected. The other ones are not. 

Screen Shot 2022-03-08 at 8 27 06 AM
Screen Shot 2022-03-08 at 8 30 06 AM
Screen Shot 2022-03-08 at 8 26 19 AM

ismael-mendoza:
This figure is not very encouraging :/ 

Screen Shot 2022-03-08 at 8 49 40 AM

jeff-regier:
Great work, Ismael.

It's fishy that three ~16 magnitude stars that are misclassified. Is that due to saturation? It'd be good to exclude any tiles that  are near saturated pixels. That may help with many of the misclassifications -- not just these three.

ismael-mendoza:
Just to be clear, the results before this post are NOT with pareto distribution since we had to fix #490 first.

dereklhansen:
@ismael-mendoza Can this be merged?

ismael-mendoza:
Hi @dereklhansen - I guess this plot is the only thing I'm worried about which you confirm you saw in the VAE side too so it's not just a plotting isssue. Although I guess results are better overall for everything else so maybe we address it in a future PR ? 

sdss-classification-acc

ismael-mendoza:
After discussing, we realized that results are overall improved for classification of all galaxies (jump from 0.4 to 0.75) so we will go ahead and merge. We will look at the classification issue for bright galaxies in a separate issue.

PR #468, opened 2022-03-04
Making Argument Specificiation Required via Hydra

declanmcnamara:
Been working on this issue today. Not too much code to work with - if you check the diffs you'll see that it's deleting the default arguments.


case_studies/sdss_galaxies/config/config.yaml
```diff
@@ -36,13 +36,15 @@ datasets:
         batch_size: 64
         n_batches: 10
         background: 865
+        n_bands: 1
```
declanmcnamara, line 39: Added a few defaults to configs (like this one) that weren't there before. If this isn't desirable, I can remove them and add those defaults back as they were. No change in behavior.

bliss/models/prior.py
```diff
@@ -74,18 +74,18 @@ class ImagePrior(pl.LightningModule):
...
+        slen: int,
+        tile_slen: int,
+        min_sources: int,
+        max_sources: int,
+        mean_sources: int,
+        f_min: float,
+        f_max: float,
+        alpha: float,
+        prob_galaxy: float,
         loc_min: float = 0.0,
```
declanmcnamara, line 87: Some of the tests (at least, the M2 test) currently depend on certain arguments having default values, such as `loc_min, loc_max`. I can't change these without altering the tests, which I didn't want to get into for this issue.

case_studies/sdss_galaxies/config/config.yaml
dereklhansen: I think this is a good change. We should err on the side of being explicit, especially since we have adding multiple bands on the road map.

bliss/models/prior.py
dereklhansen: I believe you just need to modify `tests/m2/m2.yaml`, which should be okay.

dereklhansen: I'll suggest the change on here

bliss/models/location_encoder.py
```diff
@@ -250,14 +250,14 @@ class LocationEncoder(nn.Module):
...
-        ptile_slen: int = 6,
-        channel: int = 8,
+        max_detections: int,
+        n_bands: int,
+        tile_slen: int,
+        ptile_slen: int,
+        channel: int,
+        dropout,
+        hidden: int,
         spatial_dropout=0,
```
dereklhansen, line 260: Can we make `spatial_dropout` a non-default argument too?

bliss/datasets/galsim_galaxies.py
```diff
@@ -104,21 +104,21 @@ def test_dataloader(self):
...
+        min_flux,
+        max_flux,
+        min_a_d,
+        max_a_d,
+        min_a_b,
+        max_a_b,
+        background,
+        pixel_scale,  # SDSS
+        flux_sample,
         psf_image_file: str = None,
```
dereklhansen, line 121: I think `psf_image_file` having a `None` default value is a bug; I believe this function will raise an error if it's not a valid file. We should change this to not have a default value. 

@ismael-mendoza

bliss/models/prior.py
dereklhansen: Actually, I don't think these options are even necessary; we always have loc_min=0 and loc_max=1

dereklhansen (changes requested):

dereklhansen:
Looks great @declanmcnamara! I pushed a change to remove `loc_min` and `loc_max` since it turns out those weren't ever any different than `0` and `1`.

bliss/datasets/galsim_galaxies.py
ismael-mendoza: Yes I agree with this. I believe the default should be `data/psField-000094-1-0012-PSF-image.npy`

ismael-mendoza:
> Looks great @declanmcnamara! I pushed a change to remove loc_min and loc_max since it turns out those weren't ever any different than 0 and 1.

Thanks @dereklhansen, these arguments made more sense before we produced things in tiles/padded tiles in the decoder.

bliss/models/location_encoder.py
ismael-mendoza: I also don't see why not

bliss/datasets/galsim_galaxies.py
```diff
@@ -104,21 +104,21 @@ def test_dataloader(self):
...
+        min_flux,
+        max_flux,
+        min_a_d,
+        max_a_d,
+        min_a_b,
+        max_a_b,
+        background,
+        pixel_scale,  # SDSS
+        flux_sample,
+        psf_image_file: str = "data/psField-000094-1-0012-PSF-image.npy",
```
jeff-regier, line 121: I think we'd want this default value to be set in a hydra config file rather than as a keyword argument

declanmcnamara: Got it, now fixed.

jeff-regier (approved):
Excellent!

dereklhansen:
Glad this was merged!

PR #476, opened 2022-03-15
Reporting changes from MSSISS presentation

dereklhansen:
BLISS changes:
- Allow for toggling colorbar and changing color pallete in `reporting.plot_image`.
- Propogate forward `n_sources_log_prob` so we can calculate probabilities

VAE case study changes:
- Added model-expected recall, precision, and accuracies
- Expands range of mag cuts from 18 to 24
- Bunch of plots showing tiling, COADD on and COADD off, etc
- Functions for generating LaTeX tables of metrics
- Added notebook illustrating deblending


dereklhansen:
This mostly changes my case studies folder so there's no big changes to the actual BLISS codebase. This is useful for me to have the expanded reporting metrics.

PR #478, opened 2022-03-17
Fix incorrect expected recall and precision on detections

dereklhansen:
- Don't calculate metrics on just galaxies, but overall
- We can't actually calculate recall with a magnitude cut, since
we don't estimate galaxy fluxes for objects where P < 0.5. This
could be changed.

### DIFF
--- a/case_studies/sdss_galaxies_vae/reconstruction.py
+++ b/case_studies/sdss_galaxies_vae/reconstruction.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from typing import Tuple
 
 import torch
+import numpy as np
 import pandas as pd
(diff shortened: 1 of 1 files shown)


dereklhansen:
```
      recall  expected_recall  precision  expected_precision
18  0.816667              NaN   0.886792                 NaN
19  0.775862              NaN   0.913043                 NaN
20  0.960199              NaN   0.846154                 NaN
21  0.567639              NaN   0.745665                 NaN
22  0.646104              NaN   0.775194                 NaN
23  0.587106              NaN   0.615702                 NaN
24  0.611260         0.747129   0.630854            0.875508
```

PR #483, opened 2022-03-18
Fix Broken Tests

### LINKED ISSUE #445: remove / resurrect tests in the tests_broken directory

ismael-mendoza:
Closes #445 

Mosts of the tests in `test_broken` don't need to be revived since the code was was already tested elsewhere. 
One thing that I revived was the `test_basic_star` which might be useful since this is the easiest possible test of detection/flux estimation i.e. if this one is broken we are in trouble. I also added some assertions on precision/recall for the a sdss location encoder test in my case_studies folder. Most of the work was updating the `star_basic.yaml`.


ismael-mendoza:
Tests are running in the GPU, I will merge once that is done.

ismael-mendoza:
After running in GPU there are some issues see #485 , will resolve a couple in a separate PR.

Issue #485, opened 2022-03-20
Tests failing on GPU


ismael-mendoza:
I noticed we currently have some tests failing in GPU. @dereklhansen could you fix or change the vae ones depending on what you think is worth keeping? Thank you! 

```
FAILED case_studies/sdss_galaxies/tests/test_location_encoder.py::test_sdss_location_encoder - RuntimeError: scatter_add_cuda_kernel does not have a deterministic implementati...
FAILED case_studies/sdss_galaxies/tests/test_location_encoder.py::test_location_encoder_plotting - RuntimeError: scatter_add_cuda_kernel does not have a deterministic implemen...
FAILED case_studies/sdss_galaxies_vae/tests/test_galaxy_flow.py::test_galaxy_flow - RuntimeError: index_add_cuda_ does not have a deterministic implementation, but you set 'to...
FAILED case_studies/sdss_galaxies_vae/tests/test_galaxy_vae.py::test_galaxy_autoencoder_toy_gaussian - assert 55.28435134887695 < 25
FAILED case_studies/sdss_galaxies_vae/tests/test_galaxy_vae.py::test_galaxy_autoencoder_bulge_disk - assert 34.46749496459961 < 30
FAILED tests/test_encoder_forward.py::TestSourceEncoder::test_forward - RuntimeError: Input type (torch.FloatTensor) and weight type (torch.cuda.FloatTensor) should be the same
FAILED tests/test_encoder_forward.py::TestSourceEncoder::test_sample - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and ...
```

dereklhansen:
I'll just raise the threshold - I didn't change the numbers from the auto encoder so they aren't applicable

ismael-mendoza:
> EDIT: Didn't realize these were max residuals, so they are comparable. Still, I think raising the threshold is fine since the real data is working.

Sounds good, #487 should also address the 3rd error I think (along with the first two)

PR #487, opened 2022-03-20
Fix location encoder test on gpu

ismael-mendoza:
Partially adddresses #485  (the first 3 errors listed)

Fixes the issue below by removing all `deterministic=True` usages in the codebase. I don't think we were relying on this for anything but please correct me if I'm wrong @dereklhansen , @jeff-regier. A seed can still be set via `training.seed`. 

```
E       RuntimeError: scatter_add_cuda_kernel does not have a deterministic implementation, but you set 'torch.use_deterministic_algorithms(True)'. You can turn off determinism just for this operation if that's acceptable for your application. You can also file an issue at https://github.com/pytorch/pytorch/issues to help us prioritize adding deterministic support for this operation.
```

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
-exclude = .git, conftest.py
+exclude = .git
 max-line-length = 100
 inline-quotes = double
 docstring-convention=google
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
-exclude = .git
+exclude = .git, conftest.py
 max-line-length = 100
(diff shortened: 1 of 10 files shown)


dereklhansen:
I don't think the `deterministic` flag is needed for our use-case. Because of floating point error, the result of a summation can be slightly different depending on the order it is executed in, so "deterministic=True" ensures that all such operations are done in the exact same order (at the cost of computation speed). I think the random seed is more important

ismael-mendoza:
I agree, thanks @dereklhansen

Issue #400, opened 2022-01-25
Get test_sleep_star.py working under new configuration scheme


dereklhansen:
After #396 is merged, we'll need to rewrite some of the tests we had to work under the new configuration system. 

One of these tests is called "tests_broken/test_sleep_star.py". It consists of training the LocationEncoder on simulated images of stars under two different settings. TestSleepStarTiles runs on a "vanilla" generated image, while TestSleepStarOneTile runs on an image that consists of just one tile (to check that we handle that edge case).

The steps to get this working on the new configuration system are:
1. Make a folder called `tests/star_basic`.
2. Look at each test and its overrides. These tell you what settings are needed to recreate the new test.
https://github.com/prob-ml/bliss/blob/4b3d08202ee393d2909e2cd445fafd5dfd8b88a2/tests_broken/test_sleep_star.py#L6-L12
https://github.com/prob-ml/bliss/blob/4b3d08202ee393d2909e2cd445fafd5dfd8b88a2/tests_broken/test_sleep_star.py#L35-L40
3. Reference the old configuration: https://github.com/prob-ml/bliss/tree/f6f92fb36bebb919525454f6da66bab3c9b761f3/config to create a new configuration file `tests/star_basic/star_basic.yaml`. See `tests/m2/m2.yaml` for a template. You mostly just need to update the settings for the dataset, encoder, prior, and decoder.
4. Write a `conftest.py` which creates a pytest fixture that loads this configuration (see `tests/m2/conftest.yaml`) for an example.
5. Copy `tests_broken/test_sleep_star.py` over to `tests/star_basic` and rewrite it to work with the new documentation.
6. Run `pytest -k sleep_star tests` to check that it works
7. Run `pytest -k sleep_star tests --gpu` to check that it works on the GPU (and the model trains to reasonable accuracy).

dereklhansen:
I thought this would be a good first issue at first, but it requires getting up to speed with a lot all at once... I can write a PR which gets part of this done.

ismael-mendoza:
Closed by #483

PR #489, opened 2022-03-21
unify tests in tests folder

### LINKED ISSUE #488: conftest duplication between case_studies
I noticed that the conftests in both of the `sdss_galaxies_*` case_studies are essentially the same to each other and also to the one in `tests`. Could we just use the one in `tests` instead of having repeated code? ( I had to disable linting in #487 due to the code duplication ) 

If the problem is that they are in different parentn directories one solution is to have a `tests/case_studies/sdss_galaxies_*` folders and that way all tests can use the same conftest in `tests`.

What do you think @dereklhansen ?

ismael-mendoza:
Closes #488


jeff-regier:
@ismael-mendoza  Does this PR close #400 too?

ismael-mendoza:
Yes actually #483 does but yes

PR #494, opened 2022-03-24
Introduce separate galaxy flux encoder

dereklhansen:


ismael-mendoza:
I believe the sum over pixels works fine now after we resolved the issue with coadd in #579 , so I'm closing this @dereklhansen

Issue #344, opened 2021-10-12
Add back metrics for fluxes


ismael-mendoza:
Following #327 we need to think of some metrics for fluxes/magnitudes and add them to `metrics.py`. Perhaps avg/median l1-error is sufficient? Something better would be a plot showing scatter. 

- [ ] Add to relevant tests

ismael-mendoza:
Closed by #499

PR #498, opened 2022-03-25
Enable binning by magnitude, keeping cuts as a special case

dereklhansen:

PR #499, opened 2022-03-25
add probability vs magnitude scatter and mag_true vs mag_est plot 

ismael-mendoza:


ismael-mendoza:
Hi @jeff-regier @dereklhansen , here are two recent plots further evaluating the retraining on the pareto distribution. The first one (left) confirms what we thought, that we consistently underestimate fluxes for galaxies. Derek has some preliminary results suggesting that encoding the flux separately might alleviate that so we can revisit this plot after that change is made. The star flux estimation is looking pretty good :slightly_smiling_face: (we only mis-estimate the dimmest stars as expected). 

The second one shows classification probability as a function of magnitude and this one is looking a lot better than before. It seems the uncertainty is on dimmer objects (>20 mag) like it should be. We should compare this one with binning (instead of cutting) my magnitude the classification accuracy (#498) to make sure that it matches. 

flux-comparison
sdss-mag-class-scatter

jeff-regier:
Does this PR close #344?

ismael-mendoza:
yup, thanks

PR #503, opened 2022-03-29
Modify Galsim half-light-radii to follow gamma distribution

dereklhansen:


dereklhansen:
These results look really good to me! Overall galaxy accuracy is now 84% (stars are 87%) and overall recall is 76%.

dereklhansen:
@ismael-mendoza I refactored the Galsim Galaxies a bit by splitting it into a prior and decoder, so you may want to take a look before I merge

bliss/datasets/galsim_galaxies.py
```diff
@@ -106,51 +106,101 @@ def test_dataloader(self):
...
+            total_flux = self._uniform(self.min_flux, self.max_flux, n_samples=total_latent)
+        elif self.flux_sample == "pareto":
+            total_flux = self._draw_pareto_flux(n_samples=total_latent)
+        else:
+            raise NotImplementedError()
+        disk_frac = self._uniform(0, 1, n_samples=total_latent)
+        beta_radians = self._uniform(0, 2 * np.pi, n_samples=total_latent)
+        disk_q = self._uniform(0, 1, n_samples=total_latent)
+        bulge_q = self._uniform(0, 1, n_samples=total_latent)
+        if self.a_sample == "uniform":
```
ismael-mendoza, line 186: same for the type of size distribution

bliss/datasets/galsim_galaxies.py
```diff
@@ -163,24 +213,36 @@ def _draw_pareto_flux(self, n_samples=1) -> Tensor:
...
-            [total_flux, disk_frac, beta_radians, disk_q, disk_a, bulge_q, bulge_a], dim=1
-        )
+
+class GalsimGalaxyDecoder:
+    def __init__(
+        self,
+        slen,
+        n_bands,
+        background,
+        pixel_scale,  # SDSS
```
ismael-mendoza, line 228: this comment can be removed (referred to default value) 

```suggestion
        pixel_scale
```

bliss/datasets/galsim_galaxies.py
```diff
@@ -163,24 +213,36 @@ def _draw_pareto_flux(self, n_samples=1) -> Tensor:
...
+        self.n_bands = 1
+        self.background = torch.zeros((self.n_bands, self.slen, self.slen), dtype=torch.float32)
+        self.background[...] = background
+        self.pixel_scale = pixel_scale
+        self.psf = load_psf_from_file(psf_image_file, self.pixel_scale)
+
+    def __call__(self, z: Tensor):
+        images = []
+        for latent in z:
+            image = self.render_galaxy(latent)["noiseless"]
```
ismael-mendoza, line 243: I see you are returnign only the noiseless images here, what do you use them for?

ismael-mendoza (approved):
LGTM just a couple of questions, but I think you can merge basically. I like the change of splitting the galsim dataset into a prior / decoder

bliss/datasets/galsim_galaxies.py
dereklhansen: This is used in the ImageDecoder when rendering an entire image. Since background and noise are handled for the entire image, we just need the noiseless galaxy here. It's a drop-in replacement for the single galaxy decoder in galaxy_net.py

bliss/datasets/galsim_galaxies.py
```diff
@@ -163,24 +213,36 @@ def _draw_pareto_flux(self, n_samples=1) -> Tensor:
...
-            [total_flux, disk_frac, beta_radians, disk_q, disk_a, bulge_q, bulge_a], dim=1
-        )
+
+class GalsimGalaxyDecoder:
+    def __init__(
+        self,
+        slen,
+        n_bands,
+        background,
+        pixel_scale
```
dereklhansen, line 228:
```suggestion
        pixel_scale,
```

bliss/datasets/galsim_galaxies.py
```diff
@@ -106,51 +106,101 @@ def test_dataloader(self):
...
+        a_loc: Optional[float] = None,
+        a_scale: Optional[float] = None,
+        a_bulge_disk_ratio: Optional[float] = None,
+    ) -> None:
+        self.flux_sample = flux_sample
         self.min_flux = min_flux
         self.max_flux = max_flux
         self.alpha = alpha
+        if self.flux_sample == "pareto":
+            assert self.alpha is not None
```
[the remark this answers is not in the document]
dereklhansen: `min_flux` and `max_flux` are required arguments used by both, so I don't think we need to check since you'd have to explicitly set them to None.

bliss/datasets/galsim_galaxies.py
dereklhansen: I can add some checks here; better to catch this at init time

PR #505, opened 2022-03-30
Investigate misclassifications at low magnitudes

ismael-mendoza:


ismael-mendoza:
Not sure what changed but the results are look pretty good and consistent now : 


flux-comparison
sdss-classification-acc
sdss-mag-class-scatter
sdss-precision-recall

dereklhansen:
The missclassifications might be fixed by #503.

ismael-mendoza:
Resolved after accounting for 3 objects that were actually stars but classified by coadd as galaxies.

PR #512, opened 2022-03-31
Adjust probability of n_sources at inference time

dereklhansen:
Closes #509.

PR #514, opened 2022-03-31
Get PHOTO catalog up and running for comparisons

dereklhansen:


dereklhansen:
> Nice results.
> 
> Any idea what is happening with the 22 mag galaxies? It may be worthwhile to spot check a few of the 22 magnitude galaxies that bliss misses but that Photo detects. It seems fishy thay bliss would do better at all the magnitudes but this one

Before adjusting for the training Poisson rate in #512, our recall was much higher across the board (at the cost of precision):

```
detections
           recall  expected_recall  precision  expected_precision    n  n_galaxies  n_galaxies_detected
18       0.968750              NaN   1.000000                 NaN   32          10                    9
19       0.875000              NaN   0.853659                 NaN   56          29                   23
20       0.835294              NaN   0.823529                 NaN   85          59                   46
21       0.886364              NaN   0.746269                 NaN  176         132                  112
22       0.682008              NaN   0.380198                 NaN  239         158                   84
23       0.486726              NaN   0.500000                 NaN  113          47                    6
overall  0.761317         0.537483   0.594800            0.847694  729         445                  300
```

So it could simply be a matter of PHOTO being more sensitive.

jeff-regier:
That could be it. There may also be issues with the "ground truth" (coadd). To get a feel for how reliable our ground truth is, you might sample 10 or 20 objects where Photo and BLISS differ (either at detection or classification) and, for these objects, assess whether the coadd "ground truth" results agree with the DECaLs catalog. DeCALs results can be looked up here: https://www.legacysurvey.org/viewer#IC%202038

Neither coadd nor DeCALS is totally authoritative. Both should be a lot more accurate than results computed with a single frame. If coadd and DeCALs both agree about a particular detection or classification, despite being quite different algorithmically, and using totally different image data, we can have high confidence in our "ground truth."

dereklhansen:
Here's an illustration of how we're misspecifying the distribution flux. I show the current pareto distribution used in BLISS alongside a pareto fit to all the galaxies in COADD ("Fitted Pareto"). I then show the COADD Pareto distribution truncated at the same lower bound we're currently using. This shows that BLISS is sampling roughly twice as many galaxies around 1000 flux (~22.3 magnitude) as show up in COADD. This could explain why there is such a precision drop around 21-22 magnitude. 

image

PR #516, opened 2022-04-04
Set minimum galaxy flux to 622 (23 magnitude)

dereklhansen:


dereklhansen:
The recall is down slightly after this change (57% -> 52%) but expected recall is now closer to the truth (68% -> 46%), which perhaps implies the generative model is better.

image

dereklhansen:
image

PR #519, opened 2022-04-04
Fix matching algorithm by adding sharp penalty for points far away

### LINKED ISSUE #518: `linear_sum_assignment` can lead to incorrect matches
I was digging into some recall/precision metrics again, and I realized that the `linear_sum_assignment` function might be leading to some unintuitive matching. This could be causing the bug that @ismael-mendoza showed last week with Plotly, and might be causing us to look worse on recall/precision metrics.

 `linear_sum_assignment` minimizes the total sum of distances between matched points, but this isn't really what we want. Here's an example:
image

Match 1 matches the objects which are within "matching distance" (1), but has a higher cost than the optimal solution Match 2, which yields no matches.

dereklhansen:
Fixes #518.

### DIFF
--- a/bliss/reporting.py
+++ b/bliss/reporting.py
@@ -184,6 +184,10 @@ def match_by_locs(true_locs, est_locs, slack=1.0):
     locs_err = (rearrange(locs1, "i j -> i 1 j") - rearrange(locs2, "i j -> 1 i j")).abs()
     locs_err = reduce(locs_err, "i j k -> i j", "sum")
 
+    # Penalize all pairs which are greater than slack apart, to ensure that they are not
+    # matched if possible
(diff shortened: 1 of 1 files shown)


dereklhansen:
This PR fixes the issue I was seeing in the recall/precision numbers. @ismael-mendoza Can you check your plotly script and see if this fixes the issue you were showing us last Wednesday?

dereklhansen:
On my end the recall and precision numbers are slightly improved:
image

ismael-mendoza (approved):
This makes sense @dereklhansen - thanks for looking into this more. I'll check my plotly script and post the results below.

ismael-mendoza:
Yeah it seems like all the mismatches/matches in plotly make sense now too! The examples i had looked at from the last meeting now make sense

PR #527, opened 2022-04-08
refactor location encoder

### LINKED ISSUE #388: Don't allow for different shapes of "tile_n_sources" in ImageEncoder.encode_for_n_sources
https://github.com/prob-ml/bliss/blob/50c775df202ba242faabf3e0936537617634a41b/bliss/models/encoder.py#L387-L393

We currently allow the `tile_n_sources` variable to be shape `(n_samples, n_tiles)` or `(n_tiles, )`. We should change it to only allow the former. The actual change will be a little tricky since this function is used in multiple places.

jeff-regier:
I have in mind overhauling the location encoder more thoroughly (begun on branch `jr/deeper_refactoring`), but I'm not sure when I'll get to it, so we may as merge this now. It's already a significant simplification. Closes #388 more or less. In a deeper refactoring, I'd aim to get rid of `encode_for_n_sources` (and `encode_for_n_sources_flat`, and `_get_hidden_indices`) entirely.

### DIFF
--- a/bliss/models/location_encoder.py
+++ b/bliss/models/location_encoder.py
@@ -75,24 +75,25 @@ class LocationEncoder(nn.Module):
         self.mean_detections = mean_detections
         self.max_detections = max_detections
         self.n_bands = n_bands
+
+        assert tile_slen <= ptile_slen
         self.tile_slen = tile_slen
         self.ptile_slen = ptile_slen
-        border_padding = (ptile_slen - tile_slen) / 2
-        assert tile_slen <= ptile_slen
-        assert border_padding % 1 == 0, "amount of border padding should be an integer"
-        self.border_padding = int(border_padding)
 
-        # Number of variational parameters used to characterize each source in an image.
+        assert (ptile_slen - tile_slen) % 2 == 0
+        self.border_padding = (ptile_slen - tile_slen) // 2
+
+        # Number of distributional parameters used to characterize each source in an image.
         self.n_params_per_source = sum(param["dim"] for param in self.variational_params.values())
 
-        # There are self.max_detections * (self.max_detections + 1) total possible detections.
+        # the number of total detections for all source counts: 1 + 2 + ... + self.max_detections
+        # NOTE: the numerator here is always even
+        n_total_detections = self.max_detections * (self.max_detections + 1) // 2
+
         # For each param, for each possible number of detection d, there are d ways of assignment.
         # NOTE: Dimensions correspond to the probabilities in ONE tile.
-        self.dim_out_all = int(
-            0.5 * self.max_detections * (self.max_detections + 1) * self.n_params_per_source
-            + 1
-            + self.max_detections,
-        )
+        self.dim_out_all = n_total_detections * self.n_params_per_source + 1 + self.max_detections
+
         dim_enc_conv_out = ((self.ptile_slen + 1) // 2 + 1) // 2
 
         # networks to be trained
@@ -116,11 +117,6 @@ class LocationEncoder(nn.Module):
     def forward(self, image_ptiles, tile_n_sources):
         raise NotImplementedError("The forward method has changed to encode_for_n_sources()")
 
-    def _get_images_in_ptiles(self, image: Tensor, background: Tensor) -> Tensor:
-        return get_images_in_tiles(
-            self.input_transform(image, background), self.tile_slen, self.ptile_slen
-        )
-
     def encode(self, image: Tensor, background: Tensor) -> Tensor:
         """Encodes variational parameters from image padded tiles.
 
@@ -136,7 +132,9 @@ class LocationEncoder(nn.Module):
         """
         # get h matrix.
         # Forward to the layer that is shared by all n_sources.
-        image_ptiles = self._get_images_in_ptiles(image, background)
+        image_ptiles = get_images_in_tiles(
+            self.input_transform(image, background), self.tile_slen, self.ptile_slen
+        )
         log_image_ptiles_flat = rearrange(image_ptiles, "b nth ntw c h w -> (b nth ntw) c h w")
         var_params_conv = self.enc_conv(log_image_ptiles_flat)
         var_params_flat = self.enc_final(var_params_conv)
@@ -236,7 +234,7 @@ class LocationEncoder(nn.Module):
         )
         map_n_sources = torch.argmax(n_source_log_probs, dim=1)
 
-        pred = self.encode_for_n_sources(var_params_flat, map_n_sources)
+        pred = self.encode_for_n_sources_flat(var_params_flat, map_n_sources)
 
         is_on_array = get_is_on_from_n_sources(map_n_sources, self.max_detections)
         is_on_array = is_on_array.unsqueeze(-1).float()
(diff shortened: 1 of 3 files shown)


bliss/models/location_encoder.py
```diff
@@ -305,87 +303,81 @@ def get_n_source_prior_log_prob(self, detection_rate):
...
-            "log_flux_mean": {"dim": self.n_bands, "transform": _identity_func},
-            "log_flux_logvar": {"dim": self.n_bands, "transform": _identity_func},
+            "loc_mean": {"dim": 2},
+            "loc_logvar": {"dim": 2},
+            "log_flux_mean": {"dim": self.n_bands},
+            "log_flux_logvar": {"dim": self.n_bands},
         }
 
-    # These methods are only used in testing or case studies, Do we need them or can
-    # they be moved to the code that they test? ------------------------
```
jeff-regier, line 361: Not sure what methods this comment referred to. The methods below were in use.

bliss/models/location_encoder.py
```diff
@@ -394,19 +386,15 @@ def _get_normal_samples(mean, sd, tile_is_on_array):
     def _get_hidden_indices(self):
         """Setup the indices corresponding to entries in h, cached since same for all h."""
 
-        # initialize matrices containing the indices for each variational param.
+        # initialize matrices containing the indices for each distributional param.
```
jeff-regier, line 389: I'm gradually changing "variational" to "distribution" in this file. "Distributional" isn't that descriptive, but "variational" is incorrect. The variational parameters are the weights of the encoders networks.

bliss/models/location_encoder.py
```diff
@@ -305,87 +303,81 @@ def get_n_source_prior_log_prob(self, detection_rate):
...
+                var_param,
+                "np (ns d pd) -> ns np d pd",
+                np=n_ptiles,
+                ns=tile_n_sources.size(0),
+                d=self.max_detections,
+                pd=param["dim"],
+            )
+
+        # what?!? why is sigmoid(0) = 0?
+        loc_mean_func = lambda x: torch.sigmoid(x) * (x != 0).float()
```
jeff-regier, line 363: It's a mystery to me why we have a special case for `x == 0` here. It breaks some tests if I remove it though, and I didn't want to get into it in this PR.

bliss/models/location_encoder.py
```diff
@@ -75,24 +75,30 @@ def __init__(
...
+        n_total_detections = self.max_detections * (self.max_detections + 1) // 2
+
+        # most of our parameters describe individual detections
+        n_source_params = n_total_detections * self.n_params_per_source
+
+        # we also have parameters indicating the distribution of the number of detections
+        count_simplex_dim = 1 + self.max_detections
+
+        # the total number of distributional parameters per tile
+        self.dim_out_all = n_source_params + count_simplex_dim
```
jeff-regier, line 100: The old code for computing `self.dim_out_all` was shorter but harder to decipher.

bliss/models/location_encoder.py
```diff
@@ -305,87 +303,81 @@ def get_n_source_prior_log_prob(self, detection_rate):
         log_probs_last = torch.log1p(-torch.logsumexp(log_probs, 0).exp())
         return torch.cat((log_probs, log_probs_last.reshape(1)))
 
-    def encode_for_n_sources(self, var_params_flat, tile_n_sources):
-        """Get variational parameters conditioned on number of sources in tile.
+    def encode_for_n_sources_flat(self, var_params_flat: Tensor, tile_n_sources: Tensor):
+        """A wrapper the calls encode_for_n_sources.
+        Deprecated: it's better to call encode_for_n_sources directly,
+        and to always store tiles as a 2D array (i.e., not flattened).
```
jeff-regier, line 309: Ultimately I'd like to get rid of `encode_for_n_sources_flat` and `encode_for_n_sources`. In the meantime, by splitting the old `encode_for_n_sources` into two functions, we avoid branching based on the dimension of an input.

bliss/models/location_encoder.py
```diff
@@ -305,87 +303,81 @@ def get_n_source_prior_log_prob(self, detection_rate):
...
+            var_params_for_n_sources[k] = rearrange(
+                var_param,
+                "np (ns d pd) -> ns np d pd",
+                np=n_ptiles,
+                ns=tile_n_sources.size(0),
+                d=self.max_detections,
+                pd=param["dim"],
+            )
+
+        # what?!? why is sigmoid(0) = 0?
```
ismael-mendoza, line 362: This might have to do with keeping sources that are "off", "off"

bliss/models/location_encoder.py
ismael-mendoza: This might have to do with keeping sources that are "off", "off"

bliss/models/location_encoder.py
```diff
@@ -305,87 +303,81 @@ def get_n_source_prior_log_prob(self, detection_rate):
...
-            raise ValueError("tile_n_sources must have shape size 1 or 2")
-
-        assert var_params_flat.shape[0] == tile_n_sources.shape[1]
-        # get probability of params except n_sources
-        # e.g. loc_mean: shape = (n_samples x n_ptiles x max_detections x len(x,y))
-        var_params_for_n_sources = self._get_var_params_for_n_sources(
-            var_params_flat, tile_n_sources
-        )
+        n_ptiles = var_params_flat.size(0)
+        vpf_padded = F.pad(var_params_flat, (0, len(self.variational_params), 0, 0))
```
jeff-regier, line 345: It's not clear to me why `var_parameters_flat` need to be padded. Some of the indices are out-of-bounds though if I don't pad it. The old code was padding `var_parameters_flat` too, just in a less explicit way (which copied the parameters more times)

bliss/models/location_encoder.py
dereklhansen: I made this comment when I refactored it last but I ended up deleting those methods - it should be removed.

bliss/models/location_encoder.py
dereklhansen: We shouldn't depend on the locations being zero to know whether a source is on or off.

dereklhansen: It's okay to leave this behavior since this is a refactoring PR and we don't want to break tests, but let's make it a new issue

bliss/models/location_encoder.py
dereklhansen: This is great!

bliss/models/location_encoder.py
```diff
@@ -305,87 +303,81 @@ def get_n_source_prior_log_prob(self, detection_rate):
...
-            transform = param["transform"]
-            param = self._indx_h_for_n_sources(h, n_sources, indx_mat, param_dim)
-            param = transform(param)
-            est_params[k] = param
-
-        return est_params
-
     @staticmethod
-    def _get_normal_samples(mean, sd, tile_is_on_array):
+    def _sample_gated_normal(mean, sd, tile_is_on_array):
```
dereklhansen, line 380: It looks like this is could be the source of the sigmoid problem. Why are we zeroing out the "off" sources before and after the sigmoid transformation?

bliss/models/location_encoder.py
dereklhansen: I never understood very well this part of the code, so I appreciate making this more explicit. This could probably be simplified further in a different PR

bliss/sleep.py
```diff
@@ -239,7 +239,7 @@ def _get_loss(self, batch):
         var_params = self.image_encoder.encode(images, background)
         var_params_flat = rearrange(var_params, "b nth ntw d -> (b nth ntw) d")
```
dereklhansen, line 240: This could be refactored to avoid using the now-deprecated `encode_for_n_sources_flat`

tests/test_encoder_forward.py
```diff
@@ -77,7 +77,7 @@ def test_forward(self, devices):
                 var_params[:, :-h_tile_cutoff, :-w_tile_cutoff], var_params2, atol=1e-5
             )
             var_params_flat = var_params.reshape(-1, var_params.shape[-1])
-            pred = star_encoder.encode_for_n_sources(var_params_flat, n_star_per_tile)
+            pred = star_encoder.encode_for_n_sources_flat(var_params_flat, n_star_per_tile)
```
dereklhansen, line 80: I think this could also be refactored so we can remove this method

dereklhansen (approved):

ismael-mendoza (approved):
Looks good @jeff-regier - thanks for improving the location_encoder

bliss/models/location_encoder.py
ismael-mendoza: Yes this is a good question and it's a bit subtle. It's been a while that I worked on this code but from what I remember the "off sources" had be indexed separately in the `h` matrix and this is why we had this line: 

```
h = torch.cat((h, torch.zeros(n_ptiles, 1, device=h.device)), dim=1)
```

it's a bit hard to explain but we could discuss in the next meeting. This part of the code is very convoluted so a separate PR will definitely be needed

bliss/models/location_encoder.py
ismael-mendoza: oh good point, we should use `tile_is_on_array` to track "off sources" not depend on `loc` being 0.

bliss/models/location_encoder.py
ismael-mendoza: Here is an issue https://github.com/prob-ml/bliss/issues/529

bliss/models/location_encoder.py
ismael-mendoza: Here is another: https://github.com/prob-ml/bliss/issues/530

bliss/models/location_encoder.py
dereklhansen: #529

bliss/models/location_encoder.py
dereklhansen: I pushed a commit to get rid of this method

bliss/sleep.py
dereklhansen: Fixed this

bliss/models/location_encoder.py
dereklhansen: To be clear, I fixed the underlying problem so we don't need two methods... I didn't revert the change or anything lol

Issue #529, opened 2022-04-08
do not depend on zeroed out `locs` for indicating sources are 'off'


ismael-mendoza:
use `tile_is_on_array` or `is_on_array` instead. This might require extensive refactoring I suspect there are several parts of the codebase that make this assumption. 

This might also be why we have `sigmoid(0)=0` in `location_encoder.py` 

```
loc_mean_func = lambda x: torch.sigmoid(x) * (x != 0).float()
```

Relevant comment: https://github.com/prob-ml/bliss/pull/527#discussion_r846285582

dereklhansen:
This isn't just a refactoring fix, since I realized that zero-ing out the locations causes matching issues when we want to use a detection threshold lower than 50%.

dereklhansen:
I solved my issue in #543, so we can re-close this.

PR #543, opened 2022-04-14
Add n_source_weights argument to LocationEncoder

dereklhansen:


dereklhansen:
Doesn't solve #529, but this gives a proper way to access counterfactuals for sources below <0.5 detection threshold.

Issue #540, opened 2022-04-13
location encoder refactoring


jeff-regier:
I've been thinking lately about `location_encoder.py`, and I've come away with a new appreciation of this code. It's an engineering triumph, solving a really difficult task: inferring a _variable_ number of light sources efficiently on a GPU. Because of this complexity, `location_encoder.py` is also challenging code to understand/maintain/extend. I have some ideas that may help to simplify it.

- [ ] `loc_mean_func` doesn't just apply a sigmoid transformation -- as a special case it returns zero if the input is zero, perhaps to preserve the zeros introduced by the dummy indices and `vpf_padded`? But this is a confusing way to do it, and somewhat brittle, as it relies on floating point equality for control flow. 
- [x] the `loc_mean` parameters are transformed by a sigmoid in `encode_for_n_sources`, whereas other parameters are transformed elsewhere (e.g., `loc_logvar` is transformed to `locs_sd` in `max_a_post`). It'd be better to do all the transformation in one place.
- [x] Perhaps we can delete `_sample_gated_normal` and just directly call `Normal(mean, sd).rsample()`? It seems like the gating isn't helpful frequently, e.g., in the sleep phase training loop. It may be clearer to only do the gating when necessary.
- [x] `var_params` aren't really variational parameters: the variational parameters are the parameters that are fitted by variational inference (i.e., the encoder weights). "Distributional parameters" may be better: it's kind of vague, but at least it isn't incorrect.
- [x] In `max_a_post` don't sample from a normal distribution with zero variance to get the mode, just return the mode/mean directly
- [x] `encode` currently returns all the variational parameter in flat-ish (tile x all_params) tensor that is typically referred to as `var_params_flat`. The thing is, this tensor contains two very different types of distributional parameters. One type governs the number of sources/detections per tile, whereas the other type governs the properties of each possible source/detection. By splitting `var_params_flat` up into two tensors, say `source_params` and `count_params`, `source_params` be reshaped to have dimension (tile x source x params_for_this_source). This way the same parameter for all the sources can be referenced as a slice, e.g., `source_params[:, :, 0:2]` would reference location mean -- no tedious indexing is necessary.
- [x] Don't instantiate an `NllLoss` object every forward pass, just use `F.nll_loss`

dereklhansen:
Just to chime in, after #536 most of the methods in location_encoder.py are private, so we can feel more confident eliminating/refactoring them.

The only public methods for LocationEncoder are:
- `encode`
[...]

PR #545, opened 2022-04-14
LocationEncoder: Split encoder output into 'per_source_params' and 'n_source_log_probs'

dereklhansen:


dereklhansen:
This is a first step towards #540. We immediately split the output from the neural network into per-source parameters and log probabilities. This already should simplify some of the logic and allow us to eliminate a few methods and attributes.

PR #551, opened 2022-04-19
remove mag_slack

### LINKED ISSUE #548: remove mag_slack in scene_metrics (or make it np.inf by default)
As suggested by @dereklhansen in the last meeting. Basically, we should not mix magnitude estimation with detection and classification metrics. I think this should also simplify some explanations and plots.

I will be doing this in an upcoming PR soon

ismael-mendoza:
Fixes #548 

In this PR I get rid of `mag_slack` in the `scene_metrics` functions, since as we discussed we should separate the metrics that evalute detections/classifications from the metrics that evaluate magnitude estimation.

The new plots agreed with Derek's observations that PHOTO is now doing better than before, but this is due to the fact that PHOTO has a higher flux misestimation bias which should be capture by a different metric/plot (e.g. the one below)

Screen Shot 2022-04-19 at 4 21 46 PM


ismael-mendoza:
@dereklhansen please let me know if keeping the keyword arguments could be helpful for you. But I assume that you also don't have a reason to keep the mag_slack when calculating your metrics

bliss/reporting.py
```diff
@@ -184,7 +184,7 @@ def match_by_locs(true_locs, est_locs, slack=1.0):
     locs1 = true_locs.view(-1, 2)
     locs2 = est_locs.view(-1, 2)
 
-    # entry (i,j) is l1 distance between of ith loc in locs1 and the jth loc in locs2
+    # entry (i,j) is l-infinity distance between of ith loc in locs1 and the jth loc in locs2
```
dereklhansen, line 187: This only applies to locs_err_l_infty. We should either expand this comment or delete it altogether to avoid confusion

dereklhansen: locs_abs_diff is still the L1 distance

dereklhansen (changes requested):

dereklhansen:
LGTM. Just had a concern about a comment in the matching function which is tangential to the actual code changes

bliss/reporting.py
ismael-mendoza: good point, I deleted the comment

ismael-mendoza:
Thanks for the review @dereklhansen - I'm merging now

PR #552, opened 2022-04-19
Hill Climbing Optimization algorithm for improving MAP estimates

dereklhansen:
This is a start to an optimization routine which uses the true posterior distribution to improve upon the variational MAP from BLISS


ismael-mendoza:
I don't think we have plans to implement this into the codebase in the near future. So I'm closing this issue - @dereklhansen feel free to reopen when you are back if you would like to continue it

Issue #363, opened 2021-11-05
Incorporate uncertainty in galaxy parameters into predictions


dereklhansen:
This can follow once #339 is merged.

Right now I plan on just using the mean from the galaxy VAE in the predict.py pipeline, since this most closely matches the previous behavior. However, it might be interesting to utilize this uncertainty. For example, when we estimate fluxes, integrating over a few samples would give a better estimate of the posterior mean (and also highlight uncertainty).

dereklhansen:
In #339, we use just one sample from the variational distribution (rather than the mean). With the residual VAE, the MAP estimate from the variational distribution is no longer tractable.

dereklhansen:
Copy-pasting my explanation earlier so I remember to do this
> Right now galaxy_encoder returns a single sample from the variational distribution (and the associated density ratio of that sample) to minimize changes to the rest of the code. I'll likely refactor this in #363 by having the encoder forward method return a "Normal" object (which can be sampled, have the mean and var extracted, etc.)

jeff-regier:
closed by #620

PR #556, opened 2022-04-20
Enable sampling from variational distribution

dereklhansen:


dereklhansen:
Closes #363.

dereklhansen:
- [ ] Need to make Binary and Galaxy Encoders able to accomodate multiple samples at once

dereklhansen:
This adds a way to draw samples from the variational distribution for the entire frame in the same manner in which we calculate the MAP. 

I decided to make a new class `TileCatalogSamples` which subclasses 'TileCatalog' to differentiate the samples from the original `TileCatalogs`. This is because `TileCatalogSamples` requires an extra dimension. To make the dimensions as clear as possible, the names of each dimension are stored in a class-level attribute called "shape_names": 

```
tile_cat = TileCatalog(...)
tile_cat.shape_names # ('batch_size', 'n_tiles_h', 'n_tiles_w', 'max_sources')

tile_cat_samples = TileCatalogSamples(...)
tile_cat_samples.shape_names # ('n_samples', 'batch_size', 'n_tiles_h', 'n_tiles_w', 'max_sources')

```
I hate having to think about differing dimensions, but I think this strikes a good balance. Using the `TileCatalog` class for both the MAP and samples would require rewriting a lot of our existing code to account for the extra `n_samples` dimension. Having a separate class lets us keep track where we are using samples vs using a single catalog. In particular, MyPy will fail if we try to use a method that expects a `TileCatalog` with a `TileCatalogSamples` object.

dereklhansen:
@ismael-mendoza I changed the TileCatalog class a little bit, so wanted to alert you in case it broke anything in `plots.py`.

ismael-mendoza:
Thanks @dereklhansen - I'm taking a look now

bliss/catalog.py
```diff
@@ -9,7 +9,7 @@
 from torch.nn import functional as F
 
 
-class TileCatalog(UserDict):
+class AbstractTileCatalog(UserDict):
```
ismael-mendoza, line 12: if you would like, you could make this an abstract class explicitly by inheriting from `abc.ABC` but I don't think it's necessary

bliss/encoder.py
```diff
@@ -74,47 +79,38 @@ def forward(self, x):
...
-        Returns:
-            A dictionary of the maximum a posteriori
-            of the catalog in tiles. Specifically, this dictionary comprises:
-                - The output of LocationEncoder.max_a_post()
-                - 'galaxy_bools', 'star_bools', and 'galaxy_probs' from BinaryEncoder.
-                - 'galaxy_params' from GalaxyEncoder.
-        """
-        assert isinstance(self.map_n_source_weights, Tensor)
+    def sample(
+        self, image: Tensor, background: Tensor, n_samples: Optional[int] = None
```
ismael-mendoza, line 83: it might be better to have two separate methods in the encoder just like for the location_encoder to avoid the branching, one for MAP and one for samples. But this doesn't have to be in this PR.

bliss/inference.py
```diff
@@ -18,29 +18,33 @@
 
 
 def reconstruct_scene_at_coordinates(
+    n_samples: Optional[int],
```
ismael-mendoza, line 21: isn't it a bit strange to have theh samples be optional as a positional argument rather than a keyword argument?

ismael-mendoza (approved):
LGTM - I just had one suggestion in the inference module

bliss/catalog.py
dereklhansen: Yeah I worried that would make things too complicated, especially since I want to implement some shared functionality

bliss/inference.py
dereklhansen: Yeah, I was thinking the user should have to explicitly pass "None" to know that they are getting the MAP, vs passing a 1 to get a single sample. It might just be better to split up the methods as you point out earlier

bliss/encoder.py
dereklhansen: My thinking was actually to change location encoder to just have one function for both the MAP and sampling to reduce duplication... But I agree let's discuss in a separate PR

PR #557, opened 2022-04-20
correct photo frame flux calculation

### LINKED ISSUE #553: PhotoFullCatalog computes fluxes of galaxies incorrectly
The PhotoFullCatalog is using `psfflux` for returning fluxes of galaxies too but it should use a different colum (is it `cmodel_mag` / `cmodel_flux` ?). 

This might also explain why PHOTO was doing so poorly when compared to coadd fluxes.

ismael-mendoza:
Fixes #553


ismael-mendoza:
Indeed, the PHOTO mags look better now in the scatter plot

PR #561, opened 2022-04-21
VAE Case Study: Lower LocationEncoder training rate to 0.004

dereklhansen:


dereklhansen:
Results with this change are fantastic:
image

PR #567, opened 2022-04-22
Fix device issue when running M2 on GPU

### LINKED ISSUE #564: test_m2 fails because tensors are on different devices
Runs of `pytest --gpu` raise the error below while executing `test_m2.py`: 
```
E           RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!

bliss/catalog.py:178: RuntimeError
```

dereklhansen:
Fixes #564.


jeff-regier:
Great, thanks!

Issue #563, opened 2022-04-22
three tests fail due to AssertionErrors on the GPU


jeff-regier:
When I run `pytest --gpu` (on the master branch, both before and after merging #560), three AssertionErrors are raised:
```
tests/case_studies/sdss_galaxies/test_galaxy_autoencoder.py:56: AssertionError
tests/case_studies/sdss_galaxies/test_location_encoder.py:39: AssertionError
tests/case_studies/sdss_galaxies_vae/test_galaxy_vae.py:71: AssertionError
```

jeff-regier:
If these AssertionError are due to random variation and/or incomplete optimization, we may want to start thinking about different testing strategies. Because the tests are already quite slow and (IIRC) the tolerances are already quite lax. For testing, perhaps we could initialize encoders with pre-trained weights? Or use more simplistic training and testing data (e.g., all stars have the same brightness)?

We may also want to consider having Github CI run pytest both with and without the `--gpu` flag.

dereklhansen:
> We may also want to consider having Github CI run pytest both with and without the `--gpu` flag.

I like this idea, because I never think to run the `--gpu` flag. Is there a way to use Github CI with our GPU server?

jeff-regier:
I have separate server with one GPU for running Github CI tests that is hosted in West Hall. We're already using it for Prayag and Roman's projects. Those are private repos though whereas this one is public, so we may want to be a bit more secure for this one (i.e., launch a new docker instance for each test).

PR #570, opened 2022-04-25
more location encoder refactoring

### LINKED ISSUE #540: location encoder refactoring

jeff-regier:
This PR pretty much closes #540.

I didn't remove some of the masking that should be unnecessary because I was concerned about introducing bugs, and I wasn't sure if I could rely on our tests to catch any new bugs while #563 remains an issue.

There's still quite a bit more we could do to improve the readability of the location encoder. For instance, many of the calls to `rearrange` and `squeeze` could likely be avoided by not flattening structures in the first place. But this PR may be enough for now.

### DIFF
--- a/bliss/models/location_encoder.py
+++ b/bliss/models/location_encoder.py
@@ -98,7 +98,7 @@ class LocationEncoder(pl.LightningModule):
         self.border_padding = (ptile_slen - tile_slen) // 2
 
         # Number of distributional parameters used to characterize each source in an image.
-        self.n_params_per_source = sum(param["dim"] for param in self.variational_params.values())
+        self.n_params_per_source = sum(param["dim"] for param in self.dist_params.values())
 
         # the number of total detections for all source counts: 1 + 2 + ... + self.max_detections
         # NOTE: the numerator here is always even
@@ -146,7 +146,7 @@ class LocationEncoder(pl.LightningModule):
     def encode(
         self, image: Tensor, background: Tensor, eval_mean_detections: Optional[float] = None
     ) -> Dict[str, Tensor]:
-        """Encodes variational parameters from image padded tiles.
+        """Encodes distributional parameters from image padded tiles.
 
         Args:
             image: An astronomical image with shape `b * n_bands * h * w`.
@@ -156,7 +156,7 @@ class LocationEncoder(pl.LightningModule):
         Returns:
             A dictionary of two components:
             -  per_source_params:
-                Tensor of shape b x n_tiles_h x n_tiles_w x D of variational parameters
+                Tensor of shape b x n_tiles_h x n_tiles_w x D of distributional parameters
                 per tile.
             -  n_source_log_probs:
                 Tensor of shape b x n_tiles_h x n_tiles_w x (max_sources + 1) indicating
@@ -203,11 +203,11 @@ class LocationEncoder(pl.LightningModule):
             "n_source_log_probs": n_source_log_probs,
         }
 
-    def sample(self, var_params: Dict[str, Tensor], n_samples: int) -> Dict[str, Tensor]:
-        """Sample from encoded variational distribution.
+    def sample(self, dist_params: Dict[str, Tensor], n_samples: int) -> Dict[str, Tensor]:
+        """Sample from the encoded variational distribution.
 
         Args:
-            var_params: The output of `self.encode(ptiles)` which is the variational parameters
+            dist_params: The output of `self.encode(ptiles)` which is the distributional parameters
                 in matrix form. Has size `n_ptiles * n_bands`.
             n_samples:
                 The number of samples to draw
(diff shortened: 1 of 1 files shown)


bliss/models/location_encoder.py
```diff
@@ -216,55 +216,40 @@ def sample(self, var_params: Dict[str, Tensor], n_samples: int) -> Dict[str, Ten
...
+            "fluxes": rearrange(tile_fluxes, unflatten, b=b, nth=nth, ntw=ntw),
+            "n_sources": tile_n_sources,
+        }
 
     def max_a_post(
-        self, var_params: Dict[str, Tensor], n_source_weights: Optional[Tensor] = None
+        self, dist_params: Dict[str, Tensor], n_source_weights: Optional[Tensor] = None
     ) -> TileCatalog:
-        """Derive maximum a posteriori from variational parameters.
+        """Compute the mode of the variational distribution.
```
jeff-regier, line 249: It's not exactly a map estimate.

bliss/models/location_encoder.py
```diff
@@ -399,25 +365,20 @@ def training_step(self, batch, batch_idx, optimizer_idx=0):
...
+        }
+        true_catalog = TileCatalog(self.tile_slen, catalog_dict)
 
-        true_catalog = TileCatalog(
-            self.tile_slen,
-            {
-                "locs": batch["locs"][:, :, :, 0 : self.max_detections],
-                "log_fluxes": batch["log_fluxes"][:, :, :, 0 : self.max_detections],
-                "galaxy_bools": batch["galaxy_bools"][:, :, :, 0 : self.max_detections],
-                "n_sources": batch["n_sources"].clamp(max=self.max_detections),
```
jeff-regier, line 409: Is all the slicing/clamping necessary here? How would we have more than `self.max_detections` sources in the first place?

bliss/models/location_encoder.py
```diff
@@ -374,16 +335,21 @@ def _encode_for_n_sources(
...
-        var_params_split = torch.split(pps5, split_sizes, 3)
-        names = self.variational_params.keys()
-        var_params_for_n_sources = dict(zip(names, var_params_split))
+        split_sizes = [v["dim"] for v in self.dist_params.values()]
+        dist_params_split = torch.split(pps5, split_sizes, 3)
+        names = self.dist_params.keys()
+        pred = dict(zip(names, dist_params_split))
 
-        # what?!? why is sigmoid(0) = 0?
+        # I don't think the special case for `x == 0` should be necessary
```
jeff-regier, line 343: Bryan agrees that this probably isn't necessary. I'm wary of making the change though without more reliable tests

bliss/models/location_encoder.py
dereklhansen: For the m2 example, up to five sources can be generated per tile during sleep phase training. We may need to ask Bryan about this

bliss/models/location_encoder.py
dereklhansen: I agree - we maybe should rename the ".max_a_post()" methods to be ".mode()" instead, but that can be a separate PR

dereklhansen:
> There's still quite a bit more we could do to improve the readability of the location encoder. For instance, many of the calls to rearrange and squeeze could likely be avoided by not flattening structures in the first place. But this PR may be enough for now.

I agree. I think the only place we absolutely need to flatten is for the conv layer in `.encode()`

dereklhansen (approved):
Looks great!

bliss/models/location_encoder.py
dereklhansen: Opened a new issue

PR #579, opened 2022-04-27
Code to download coadd

### LINKED ISSUE #470: Add code for download coadd_catalog
We have the coadd catalog for a particular sdss frame in [coadd_catalog_94_1_12.fits](https://github.com/prob-ml/bliss/blob/master/data/coadd_catalog_94_1_12.fits). It would be good to document where this file came from, especially if we want to extend our analysis to different frames.

### LINKED ISSUE #511: investigate mismatch of input flux with galsim galaxy image flux
Is it because of the long tails of the DeVaucauleur component? should we renormalize the image ?

ismael-mendoza:
- Fixes #470 
- Fixes #511 

I simplified some of the code in `reporting.py` that would overwrite the coadd catalog and save it (it's better if we keep the downloaded data as is). Some of the functionality there I when loading the catalog, and other that I hand't used I removed. 

I also added a README for how to download the coadd catalog for future reference.


jeff-regier (approved):
LGTM

ismael-mendoza:
Hi @dereklhansen - I added the `wcs` as an argument to the `CoaddCatalog` so we don't need to overwrite the coadd file I downloaded from CAS. This modified some of the code in your catalog classes, so I just wanted to see if you were ok with these changes or if you have any suggestions before merging.

PR #584, opened 2022-04-27
Delete adjustment probabilities

### LINKED ISSUE #582: delete code for adjusment probabilities

ismael-mendoza:
Closes #582 

Should be merged after #579


ismael-mendoza:
After removing the adjusment probabilities - all the blends are fixed! Except for one that I should investigate where perhaps its close to the border of a tile. 

The precision/recall/f1 is the best I've seen as well.

PR #589, opened 2022-04-29
Use fixed validation set when training

### LINKED ISSUE #581: fix validation dataset when training and make it larger

dereklhansen:
Closes #581.


dereklhansen:
My thinking is we should have a validation set that is at least the size of an entire frame (possibly larger). A back of the envelope calculation suggests we need at least 387 88x88 images in the validation set:
image

The AE currently has 320 images(batch size = 32, n_batches = 10), so slightly less than a full frames' worth. The VAE currently has 5 x 320 images (n_workers =5, batch_size = 32, n_batches = 10), which works out to slightly over 4 frames' worth.

dereklhansen:
image

PR #591, opened 2022-04-29
KDTree neighbor search for matching problem

### LINKED ISSUE #346: KDTrees instead of Hungarian algorithm for matching
@jeff-regier mention this idea where for matching objects it might be possible to use KDTrees rather than the Hungarian Algorithm given that locations have spatial structure. Not urgent to implementbut good to keep in mind if we ever want to make the matching faster.

wangchv:
- Fixes #346 

Replace the Hungarian algorithm (now is JV algorithm) with KDTree neighbor search `match_by_locs_kdtree`. There are two methods in the new matching algorithm:
- `Nearest neighbor search `(default method): for each point, only search for the nearest neighbor (within threshold). If the nearest neighbor are not unique, like two points have same nearest neighbor, then set the further one as unmatched.
- `Radius neighbor search`: for each point, search for neighbors within threshold and match it with the closest and unique point, otherwise set it as unmatched.

Some issues may affect other parts of program:
- If there are no matches, the index return would be empty arrays, and `avg_distance` is Nan.
- `False` in `dist_keep` only represent points at (0,0). Matches with distance larger than threshold have been excluded in the algorithm. 
- `avg_distance` only takes mean on matches within threshold, which is different from the previous version.


dereklhansen:
Hi Cheng! Thank you for working on this!

To make comparisons easier, instead of replacing the existing `match_by_locs` matching algorithm, would you be able to add yours as a separate function `match_by_locs_kdtree`? We will need to have both for comparisons. Down the road, once we know the new matching algorithm is working well, we can remove the old matching code.

wangchv:
@dereklhansen  Of course! Thank you for pointing it out. Just done it!

jeff-regier:
Thanks for working on this @wangchv. The red "X" below indicates that some of checks aren't passing for this pull request. If you click on "Details", you'll see which checks aren't passing. It appears that your code needs to reformatted to match the coding conventions of this repo, which are checked by a program called `black`. The red X will be replaced by a green check once all checks pass.

It may help to run `pre-commit --install` from within the bliss repository. That will make it so that these checks are run locally when you commit (can be disabled with the `--no-verify` flag, i.e., `git commit --no-verify`). It's often easier/faster to get the checks to pass locally, rather than waiting on the CI server to report what's wrong.

jeff-regier:
I updated our project [readme](https://github.com/prob-ml/bliss/blob/master/README.md) so that it mentions our pre-commit hooks in step 7.

wangchv:
@jeff-regier  Thank you for the instruction! I've been working on it since yesterday and the newest version passed.  The converage test failed because I use another function `match_by_locs_kdtree`, which is not included in the program, to represent the new methods for comparison. If there's anything else I need to do, please let me know. Thank you!

jeff-regier:
@wangchv Will you add a unit test that runs the new code and checks that new code performs as expected? That will keep the "codecov/project" check from failing.

wangchv:
@jeff-regier  Do you mean that I should add a same test as the current one on the new code? It seems I need to modify some other files. Or just use some simple cases to check the output of the function in the `test_metrics.py`?

jeff-regier:
I see, just a simple test case or two is enough

wangchv:
@jeff-regier  Hi professor, I just added two test cases in `test_matrics.py` file, and now it passes the checks. I also added a new argument to `match_by_locs_kdtree` to control the methods used for search (0 for radius search and 1 for nearest search). If we finally decide to use one of them, we can just drop this argument.

jeff-regier (commented):
Thanks for your work @wangchv. I think it's premature to merge this PR since the current approach (on master) is fast enough for our uses today, and there's quite a bit more to do to turn this PR into "production" code. I learned from this PR that KDtrees are less of a drop-in replacement for Hungarian algorithm than I had realized, even when the "costs" are all Euclidean distances. Let's return to this PR once we have orders of magnitude more detections to contend with, as may be the case for detecting subhalos.

Issue #587, opened 2022-04-28
Decouple image reconstruction from MAP estimation


dereklhansen:
We currently use the code in `inference.py` to both run the encoder on the entire frame and reconstruct it at the same time. However, I don't even use the whole-image reconstruction in my analysis anymore.

In fact, if we adopt Option 2 in #572, we could just use the Encoder object on the entire frame with a DataLoader to shuffle padded tiles on/off the GPU, which is the only reason we needed the code in `inference.py` in the first place. Then we could delete `inference.py` altogether, which is a complex beast.

As a bonus, we could report a faster runtime if we only produced the TileCatalog on the entire frame.

PR #596, opened 2022-05-02
Use single dimension for ptiles in encoder (LocationEncoder, BinaryEncoder, GalaxyEncoder)

### LINKED ISSUE #572: Keep number of dimensions consistent throughout LocationEncoder
Following discussion in #570, we shouldn't need to flatten/unflatten throughout the LocationEncoder. I see two possibilities:

1. The LocationEncoder keeps the full dimensions throughout (`n_samples x batch_size x n_tiles_h x n_tiles_w x max_sources x D`).
2. The LocationEncoder only operates on padded tiles (`n_samples x n_ptiles x max_sources x D`) throughout.

I'm starting to think Option 2 might be better, because it would make the implementation of LocationEncoder simpler. With this change, methods higher up in the call stack (like `Encoder.max_a_post()` or `.training_step()` from PyTorch Lightning) would be responsible for:
- Creating a valid TileCatalog object with the correct arrangement of tiles (i.e. `"n_ptiles -> batch_size n_tiles_h n_tiles_w"`
- Transforming the input image and background into padded tiles

dereklhansen:
Closes #572.

This is toward #587 as well. Having a single ptile dimension will make it easier to move chunks on/off the GPU when the frame is too big to fit in GPU memory

### DIFF
--- a/bliss/catalog.py
+++ b/bliss/catalog.py
@@ -54,6 +54,13 @@ class TileCatalog(UserDict):
         assert x.shape[:4] == (self.batch_size, self.n_tiles_h, self.n_tiles_w, self.max_sources)
         assert x.device == self.device
 
+    @classmethod
+    def from_flat_dict(cls, tile_slen: int, n_tiles_h: int, n_tiles_w: int, d: Dict[str, Tensor]):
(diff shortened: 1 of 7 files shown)


jeff-regier:
Great!

PR #597, opened 2022-05-02
Allow Encoder to handle entire frame and remove ChunkedScene

### LINKED ISSUE #587: Decouple image reconstruction from MAP estimation

dereklhansen:
Closes #587

### DIFF
--- a/bliss/encoder.py
+++ b/bliss/encoder.py
@@ -116,20 +116,18 @@ class Encoder(nn.Module):
         )
 
     def _make_ptile_loader(self, image: Tensor, background: Tensor, n_tiles_h: int, n_tiles_w: int):
-        img_background = torch.cat((image, background), dim=1)
+        img_bg = torch.cat((image, background), dim=1).to(self.device)
         n_rows_per_batch = self.batch_size // n_tiles_w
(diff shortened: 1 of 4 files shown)


dereklhansen:
@ismael-mendoza I managed to remove the ChunkedScene code and just use the Encoder directly for performing inference on the whole frame. This lets us remove a lot of confusing lines from the codebase... Can you check that it works on your case study?

ismael-mendoza:
Yup, let me try it. Thanks! this looks great

ismael-mendoza (approved):
Looks good, I pushed a small fix and all the figures I have stayed the same 😃

PR #601, opened 2022-05-04
AE case study: plot bin errors of individual galaxies reconstruction measurements as function of SNR

ismael-mendoza:
The plots look reasonable know and ready to be put in the paper. The only thing is that g1 and g2 differ which I don't have an explanation for. Could the AE be biased for low SNR images in only the g1 direction? 

Screen Shot 2022-05-04 at 9 34 09 AM


ismael-mendoza:
The plots improved so I'm merging this PR, BLISS still does well for high SNR galaxies in both cases. I'm still confused why AE performs differently for each ellipticity as a function of SNR so I'll create an issue.

PR #605, opened 2022-05-04
Fix documentation errors by pinning git push action

dereklhansen:
It looks like the recent crashes in our documentation were due to an update from the Github Action I was using to force-push to master-sphinx. I pinned this action at the last good commit.


ismael-mendoza (approved):
Looks good!

PR #606, opened 2022-05-05
Make variational mode a special case of LocationEncoder.sample()

dereklhansen:
This refactors LocationEncoder to use the same logic for both sampling and variational mode (with the latter now a special case of the former).


jeff-regier (approved):
Wonderful!

PR #620, opened 2022-05-13
Add sampling capability to Encoder

dereklhansen:


dereklhansen:
So far I've laid the groundwork for enabling sampling by allowing for BinaryEncoder and GalaxyEncoder to handle input locations of the shape "n_samples x n_ptiles x max_sources x 2". The MAP still gives the same output, but I still need to enable sampling and test it out.

PR #622, opened 2022-05-15
Retrain AE with fixed validation dataset and larger batches

ismael-mendoza:
Closes #516


ismael-mendoza:
Overall results improved and/or stayed the same after retraining with a fix validation dataset with more images per epoch. 
All the of the results of hand-picked blends look better and global statistics results on precision/classification stayed mostly the same. The precision decreased a little bit due to predicting more stars at the high magnitude end. We also classified some before galaxies as stars now. But this is above 22 magnitude (see prob scatter plot) where uncertainty should be high

Screen Shot 2022-05-18 at 14 53 11
Screen Shot 2022-05-18 at 14 53 03
Screen Shot 2022-05-18 at 14 52 54

Issue #508, opened 2022-03-30
Make minimum of pareto distribution 23


ismael-mendoza:

dereklhansen:
Leaving this issue open since I didn't change any settings for @ismael-mendoza's paper in #516

PR #630, opened 2022-05-29
lowered threshold for test_binary

jeff-regier:


ismael-mendoza:
I'm not sure why the tolerance of `m2` changed, but I believe this one changed because I made the simulated objects for training dimmer in #622 which was related to issue #508

PR #632, opened 2022-05-30
homogenization unit-tests

wangchv:
I added two tests to verify that homogenization preserves the image flux:
- The first test uses SDSS frame that is same with `case_studies/psf_homogenization/main.py`.
- The second test uses 1 galaxy image created with galsim.

I tested the error to be less than 1% level (It actually could reach 0.5%).


ismael-mendoza (approved):
The test looks great, thanks @wangchv !

PR #634, opened 2022-05-30
Run encoder on batches

ismael-mendoza:
Allows the encoder to process multiple images in batches when doing `variational_mode`, which is required for doing inference on multiple images of simulated blends simultaneously.

If I tried to do this before on 10k galsim blends, the GPU would run out of memory.


ismael-mendoza:
Hi @dereklhansen - I took your suggestion on how to improve the encoder to process multiple images in this PR. I'm merging now since I know you are in vacation, but if at some point you have additional suggestions on this PR and you could give me a quick review that would be great. Thanks!

dereklhansen (commented):
Looks great to me!

ismael-mendoza:
Thanks!

PR #637, opened 2022-06-01
Add metrics for ellipticity and magnitude on galsim simulated blends as function of SNR and blendedness

### LINKED ISSUE #456: plots and/or metrics for evaluating BLISS on simulated galaxy blends
Should be done after #455 

Something along the lines flux/ellipticity erros as a function of blendedness rate (w/ closes, total), SNR. See Arcelin et al. 2020 Figure 6 for inspiration

- [ ] identify BLISS biases on simulated images (maybe start with 2 galaxies only so easier to do that?) 
- [ ] Consider using `REGAUSS` to measure the shape of individual galaxies (in case current method is weird?) 
- [ ] add functionality to get true/predicted/residual image for random blends from test dataset

ismael-mendoza:
Closes #456


ismael-mendoza:
Done! Results look promising I think (first bin in SNR has 5 objects, so ignore for now) 

image

PR #640, opened 2022-06-06
Im/full galsim decoder

ismael-mendoza:
This PR creates a new decoder that relies only on galsim. This should be an easier decoder for @wangmallory and @wangchv to customize the decoder in `decoder.py`. Since we are already using galsim in `decoder.py` I'm not sure that this would be slower than what we already have. 

- [x] example notebook
- [x] check if we can train encoders with new decoder (as with the simulated dataset) 
- [x] speed? - NOTE: speed for training is reasonable given batch_size = 32 with num_workers = 32
- [x] add unit test
- [x] pass all tests
- [x] coverage looks ok 
- [x] check get same results with metrics for blends plot using this new dataset and generate
- [ ] ~~is detection encoder training performance on new dataset comparable to main branch?~~ -> check this later once we implement stars
- [x] check that 10 workers is enough for training fast 
- [x] is the cpu and gpu utilization high ?


ismael-mendoza:
Hi @dereklhansen - I created this "full galsim decoder" that does not use tiles (like the `decoder.py`) to generate images for training from galsim. I think it should be much simpler to extend and use than the other one but I'm open to suggestions on it if you have sometime. I also think it should have comparable runtime to your decoder when using galsim as the galaxy model. I haven't added stars to it but that should be an easy extension. 

I review on this would be really helpful, @jeff-regier thinks that we could replace the original decoder with this one, but that would be in a future PR once stars are added and performance on real data looks the same.

bliss/models/galsim_decoder.py
```diff
@@ -0,0 +1,291 @@
...
+        sample = super().sample()
+        n_sources = sample.pop("n_sources")
+        flux = sample["galaxy_params"][:, 0].reshape(-1)
+        idx_order = torch.argsort(-flux)
+        reordered_sample = {k: v[idx_order] for k, v in sample.items()}
+        reordered_sample["locs"][0, :] = 0.5  # centered
+        return {"n_sources": n_sources, **reordered_sample}
+
+
+class GalsimGalaxiesDecoder:
```
dereklhansen, line 220: I assume this is where the magic happens

dereklhansen, line 220: I assume this is the new decoder (i.e. meant to replace the "ImageDecoder")

bliss/models/galsim_decoder.py
```diff
@@ -0,0 +1,291 @@
...
+    def __init__(
+        self, single_galaxy_decoder: SingleGalsimGalaxyDecoder, slen: int, bp: int
+    ) -> None:
+        self.single_decoder = single_galaxy_decoder
+        self.slen = slen
+        self.bp = bp
+        assert self.slen + 2 * self.bp >= self.single_decoder.slen
+
+    def __call__(self, full_cat: FullCatalog):
```
dereklhansen, line 229: I like that it takes a FullCatalog as input - maybe we could just call this a FullCatalogDecoder?

bliss/models/galsim_decoder.py
```diff
@@ -0,0 +1,291 @@
...
+        return self.render_galaxies(full_cat, self.single_decoder.psf)
+
+    def render_galaxies(self, full_cat: FullCatalog, psf: galsim.GSObject):
```
dereklhansen, line 232: Call this "render_catalog" instead? Might as well if you are extending to stars eventually

dereklhansen (changes requested):

dereklhansen:
I think the new decoder looks a lot simpler that what we had before and makes sense to eventually replace what we have in `bliss/decoder.py` (especially if there's not a computation hit).

| There is a regression in performance of BLISS run on simulated galsim blends from this dataset compared to last time

It should be possible to compare the two decoders on the same catalog, correct? e.g. in pseudocode

```
tile_cat: TileCatalog
tile_decoder: ImageDecoder
full_decoder: GalsimGalaxiesDecoder # should be renamed to FullCatalogDecoder or something
full_cat = tile_cat.get_full_catalog()

image1 = tile_decoder.render_images(tile_cat)
image2 = tile_decoder.render_images(full_cat)

assert torch.allclose(image1, image2)

bliss/models/galsim_decoder.py
ismael-mendoza: That's right!

ismael-mendoza:
> It should be possible to compare the two decoders on the same catalog, correct? e.g. in pseudocode
Thanks for looking into this so quickly! I actually found out what the issue was ( I was rendering the center brightest galaxy in this [prior](https://github.com/prob-ml/bliss/pull/640/files#diff-d56e08046009da07e785efa607e1825f430a169e7ca3d882fd98b88a4d9eca8aR208) exactly in the middle of two tiles). The datasets were equivalent after all

bliss/models/galsim_decoder.py
ismael-mendoza: and here is the associated [dataset](https://github.com/prob-ml/bliss/pull/640/files#diff-1a8faf476915ed024d8eedd8cbd2d8c94bfc57ba4b79c57280776194793d54f1R77) that combines the prior and the decoder

ismael-mendoza:
@jeff-regier one last thign I wanted to check was that 10 workers was indeed sufficient for fast training (it is) and the cpu and gpu utilization were high. I

Sometimes the CPU utilization would oscillate and not be 100% throughout, perhaps this suggest some improvement that can be made or increase the batch_size? In anycase I think we can merge this since the performance seems reasonable to me. The GPU utilization was high throughout training

Screen Shot 2022-06-16 at 09 01 13
Screen Shot 2022-06-16 at 09 00 52

jeff-regier:
Looks good!

dereklhansen:
>  I modified catalog.py to relax the assumption that when converting a FullCatalog to a TileCatalog there are at most max_sources_per_tile (see the to_tile_params method) . If there are more source in a given tile than max_sources_per_tile, I just arbitrarily pick one of them for the tile _catalog. This was necessary because the FullCatalogDecoder has no constrains on the the number of sources in each tile.

@ismael-mendoza This is dangerous to have as the default behavior. I think it should raise an error in the case there are more than "max_sources" in a tile. 

If this behavior makes sense for your use-case, it should be a keyword argument to `to_tile_params()` whose default argument is `False`. However, maybe a better approach is to just bump up the number of max sources (i.e. expand each tensor along that dimension by one).

ismael-mendoza:
Good point, I will fix this in a new PR

ismael-mendoza:
> However, maybe a better approach is to just bump up the number of max sources (i.e. expand each tensor along that dimension by one).
Unfortunately, the new decoder just places galaxies randomly in the image, so in theory the number of galaxies per tile is can go up to infinity, of course even 2 is an unlikely outcome, but it can happen. I think I will add a keyword argument to make this behavior explicit for now

PR #643, opened 2022-07-05
add DeCals catalog for truth

### LINKED ISSUE #623: use DECaLS for ground truth instead of SDSS coadd
See: https://www.legacysurvey.org/dr9/files/#sweep-brickmin-brickmax-fits

ismael-mendoza:
Closes #623 

## Tasks
- [x] create new class (a la PhotoFullCatalog) to read catalog into FullCatalog
- [x] filter to range of stripe 82 frame we have been using (r > 336.5825) & (r < 336.7461) & (d > -1.044) & (d < -0.8189)
- [x] determine correct cut in magnitude (what is the maximum magnitude detected by PHOTO in frame?) ==> `< 23`
- [x] use interactive plotting to check reasonable locations and sources are predicted
- [x] check magnitudes agree with BLISS more or less or PHOTO
- [x] recreate detection metrics but with this new catalog
- [x] understand mismatch with some bright objects

## Resources

- Catalog containing stripe 82 data we have been using: https://portal.nersc.gov/cfs/cosmo/data/legacysurvey/dr9/south/sweep/9.0/
- Description of sweep catalogs: https://www.legacysurvey.org/dr9/files/#sweep-catalogs-region-sweep (subset of tractor catalogs) 
- Convert flux to mags: https://www.legacysurvey.org/dr5/description/#photometry

## Code

```bash
# create path to 'output-icml-2022'
# relevant table will be in  ./output-icml-2022/sdss_figures/tex
./main.py mode=reconstruct paths.output=./output-icml-2022 +reconstruct=sdss_figures 
```


ismael-mendoza:
Hmm I might need to look more into the WCS, since I can't use the SDSS one. Maybe this comment will help? 

> Image stacks are on tangent-plane (WCS TAN) projections, 3600 × 3600 pixels, at 0.262 arcseconds per pixel.

case_studies/sdss_galaxies_vae/reconstruction.py
```diff
@@ -398,7 +398,7 @@ def create_figure_at_point(
 
     if est_catalog is not None:
         tile_est_catalog = est_catalog.to_tile_params(
-            tile_map_cropped.tile_slen, tile_map_cropped.max_sources
+            tile_map_cropped.tile_slen, tile_map_cropped.max_sources, ignore_extra_sources=True
```
ismael-mendoza, line 401: Hmm the `est_catalog` (is it from BLISS?) cannot possible throw this error. Need to investigate

ismael-mendoza:
@dereklhansen @jeff-regier 

I investigated some bright objects for which BLISS get's it easily but it seems the DeCals catalog is way off. 
The conclusion: I have no idea what is going on 🤷 . Initially I thought, the skyviewer and DeCals tractor catalog I downloaded were different, now I confirmed that they are indeed the same and have the same (ra, dec) and `objid` for the objects I was looking into. 

I carefully look into a pair of bright objects. Both of which BLISS gets it right (if you look at the SDSS image) but for one of them the tractor catalog appears way off. **However**, if you look at the skyviewer image it looks like DeCals is spot on for both. Initially I thought this might be a WCS error, but as you can see in the second pair of images, there are sources for which tractor and BLISS are both spot on the SDSS image and the Skyviewer image. Again, it's important the `objid` match to make sure we are referring to the same object, and they do.

So I'm out of leads for now... 

Note: The legend should read 'galaxy decals' instead of 'galaxy coadd'

Screen Shot 2022-07-09 at 11 57 41
Screen Shot 2022-07-09 at 11 54 03
Screen Shot 2022-07-09 at 11 54 14
Screen Shot 2022-07-09 at 12 06 58

jeff-regier:
For the second object you mention, SDSS Sky Navigator seems at odds with DECaLS too:
Screen Shot 2022-07-11 at 11 00 18 AM

jeff-regier:
The Gaia catalogs agree with DECaLS. The DESI spectrograph catalog also agree with DECaLS, reporting the direction as 336.6499, -0.8339.

This may be an SDSS alignment issue.

jeff-regier:
The second object is a moving star! I asked Dustin Lang about it. He writes

> SDSS was taken around year 2000, and the DECam data was taken 2013-2018, and the star has moved in between!  From the Gaia-EDR3 data it has a measured proper motion of about (100, -100) milli-arcseconds per year (and a significant parallax), so in 15 years it moves a visible amount!

ismael-mendoza:
I decided to merge and keep everythign the same, later we can remove SDSS coadds catalog once the issues are fixed and we all agree on this (didn't want to change Derek's code)

PR #644, opened 2022-07-14
[Strong Lensing]

yashpatel5400:
The commits are split up to be both functionally independent and hopefully easier to review. Commit titles have the information about the overall goals, but for details:

- Add lensing to decoder
  - Adds SIE-based lensing grid sampler to the GalsimDecoder
  - Adds support for PSF parameter loading to GalsimDecodoer (moved the functionality from StarTileDecoder)
- Add Galsim galaxy encoder
  - Encoder specific to Galsim parameters: note that the plotting hasn't really been tested
  - The abstraction of sampling/loss calculations is to allow for their use in the LensEncoder (next diff)
- Add lensing encoder
  - Encoder for lensing: note that the plotting hasn't really been tested again
  - There is likely some room to abstract both the Galsim and Lens encoders, since they're pretty similar
- Integrate lensing into overall encoder
  - Adds ability to use binary encoder to classify between lensed and unlensed galaxies
  - Adds properties to the catalog
  - Adds encoders specific to lensing to the overall encoder
- Add SDSS data for strong lensing
  - Nothing really
- Add strong lensing case study w/ models
  - Training scripts/configs for lensing case study: *very* similar to SDSS case study
- Add lensing tests
- Added scripts for posterior coverage
  - One script for binary posterior coverage (i.e. galaxy/lens indicators)
  - One for CI of continuous variables
  - Note: running these scripts requires some modification to the overall encoder (to condition on the truth), which I do locally but am not checking in. We could add this in if it is of interest for other runs


dereklhansen:
I don't think I'll have time to check this right away. Most of the diff looks like adding the new lensing functionality and case study, so perhaps it's less scary than it seems.

Since there are significant changes to files like `decoder.py`, we need to be careful that the results of the existing case studies stay the same. In short:

- [ ] Make sure all Github CI tests pass (currently failing on black)
- [ ] Make sure that `case_studies/sdss_galaxies` gets the same result as before @ismael-mendoza 
- [ ] Make sure that `case_studies/sdss_galaxies_vae`` gets the same result as before @dereklhansen

bliss/models/binary.py
```diff
@@ -31,6 +31,7 @@ def __init__(
         spatial_dropout: float,
         dropout: float,
         optimizer_params: dict = None,
+        catalog_tag: str = "galaxy_bools",
```
dereklhansen, line 34: I gather this is to accommodate "galaxy_bools" and "lensed_galaxy_bools". This change makes it hard to understand the code since it requires knowing this context.
It would be better to have a separate Module for the lensed_galaxy_bools. It's better to violate the DRY principle a bit for clarity and to keep things disentangled.

bliss/encoder.py
```diff
@@ -32,7 +34,9 @@ def __init__(
         self,
         detection_encoder: DetectionEncoder,
         binary_encoder: Optional[BinaryEncoder] = None,
-        galaxy_encoder: Optional[GalaxyEncoder] = None,
+        galaxy_encoder: Optional[Union[GalaxyEncoder, GalsimEncoder]] = None,
+        lensing_binary_encoder: Optional[BinaryEncoder] = None,
```
dereklhansen, line 38: Let's make LensingBinaryEncoder a new module. Better to repeat ourselves a little now than to entangle these two separate parts of BLISS

dereklhansen (changes requested):

dereklhansen:
I took a look at the hotspots where existing code is getting changed. I only noticed one high-level issue that should be easy to change. It would be cleaner to make a separate `LensedBinaryEncoder` module than to force the `BinaryEncoder` to handle multiple cases and require use of `self.catalog_tag`. It makes sense to abstract the functionality of encoding a binary variable on a tile, but we are currently mixing different levels of abstraction in the BinaryEncoder - in some spots 'self.catalog_tag' is used right next to "star_bools", but having a "star_bools" variable only makes sense if we are encoding galaxies.

It's fine to repeat ourselves a bit for now - we can figure out the ideal abstraction later.

dereklhansen:
Other than my one comment, I feel good about this being merged once it passes CI and we confirm that the existing case studies still work.  Since @yashpatel5400 already organized the changes into neat commits, I suggest doing a regular merge (or a rebase) rather than squashing everything.

bliss/models/lens_encoder.py
```diff
@@ -0,0 +1,356 @@
...
+            pure_lens_params,
+            pure_lens_pred,
+        ) 
+        
+        loss_lensed_galaxy = get_galsim_params_nll(
+            batch["lensed_galaxy_bools"],
+            lensed_galaxy_params,
+            lensed_galaxy_pred,
+        )
+
```
ismael-mendoza, line 177: Is the lens encoder in charge of both predicting the lens params and the galaxy params at the center of the lens? I wonder if the galaxy encoder could be in charge of the latter and the lens encoder of the former. I wonder if that could improve performance too.

bliss/models/lens_encoder.py
```diff
@@ -0,0 +1,356 @@
...
+        pure_lens_dim = self.latent_dim - self.latent_dim_split
+        for param_idx in range(pure_lens_dim): # theta_E, center_x, center_y, e1, e2 = pure_lens_params
+            dist_mean = pure_lens_params[...,2 * param_idx]
+            dist_logvar = pure_lens_params[...,2 * param_idx + 1]
+            dist_sd = (dist_logvar.exp() + 1e-5).sqrt()
+
+            param = Normal(dist_mean, dist_sd).rsample()
+            if param_idx in [0]: # theta_E is log-normal
+                param = param.exp()
+            elif param_idx in [3, 4]: # e1, e2 (3, 4) need to invert: torch.logit((e1 + 1) / 2): TODO: change to angle to constrain relation
```
ismael-mendoza, line 107: is this ellipticity? what is the parametrization you are using ?

bliss/models/galsim_encoder.py
```diff
@@ -0,0 +1,347 @@
...
+    params = params.view(-1, params.shape[-1]).transpose(0, 1)
+
+    # only compute loss where lensing is present
+    true_params = true_params[:, galaxy_bools > 0]
+    params = params[:, galaxy_bools > 0]
+
+    # if no lenses present, skip loss calculation
+    log_prob = torch.zeros(1, requires_grad=True).to(device=true_params.device)
+    if true_params.shape[-1] == 0:
+        return -log_prob
```
ismael-mendoza, line 36: Not sure I understand this, don't you want the galaxy encoder to work even if no lenses are present?

ismael-mendoza (changes requested):
Hi @yashpatel5400, thanks for organizing the commits and this PR. This is great progress! 
I had a bunch comments , ranging from minor/refactoring and a couple which might take a bit longer/require more discussion.

In general, I agree with @dereklhansen that it would be great to separate the functionality of the lens/lens binary encoder as much as possible from other encoders. The same goes from the GalsimEncoder which both Derek and me are looking to use separately(so it would be good to make it independent of the lens functionality). 

Please let me know if you have any questions about my comments, and hopefully they are helpful!

yashpatel5400:
Gonna split up this PR into a bunch of smaller ones (re: a bunch of the comments above). Closing this one!

PR #645, opened 2022-07-20
Mallory/align

wangmallory:
This branch only contains align_single_exposures function and signal_noise_ratio function. The test file is under tests.case_studies.coadds folder.


wangmallory:
Found an error in the function I need to fix.

PR #647, opened 2022-07-29
Add stars to full catalog decoder

ismael-mendoza:


ismael-mendoza:
Thanks @wangchv for working on this. It's a little hard to review because of the rebasing I was saying ( I think you will need to copy the new code to a different branch and start over the PR). 

From what I gathered in the diff, it looks like you are in the right track. You just need to remove my original classes and replace with yours. Let's talk more about exactly what changes to make in our meeting

PR #660, opened 2022-08-03
change name of fluxes to star fluxes where applicable

ismael-mendoza:


bliss/inference.py
```diff
@@ -218,11 +218,9 @@ def __init__(
             full_coadd_cat.plocs = full_coadd_cat.plocs + 0.5
             max_sources = dataset.image_prior.max_sources
             tile_catalog = full_coadd_cat.to_tile_params(self.tile_slen, max_sources)
-            tile_catalog.set_all_fluxes_and_mags(dataset.image_decoder)
```
ismael-mendoza, line 221: @dereklhansen I suspect you don't need this line, since the coadd catalog already has full fluxes and mags set? Let me know if I missed somethign

PR #663, opened 2022-08-03
Add lensing encoder

yashpatel5400:
What the title says!


jeff-regier (approved):
Looks good to me. (The duplicate code is temporary and the test coverage will improve with #664 and #665)

PR #664, opened 2022-08-03
Strong Lensing case study

yashpatel5400:
Adds case study with training scripts, data, and frozen models


jeff-regier (approved):
Looks good (these changes are all in case_studies/strong_lensing)

PR #665, opened 2022-08-03
Encoder/decoder integration

yashpatel5400:
This is the PR where all the other lensing encoders and decoders are integrated, so this partially touches the core pipeline. To address some questions about the choice for the lensing decoder, here's some context about rendering lensed galaxies as we currently support it: lensed galaxies are assumed to be lensed by some other galaxy (i.e. we don't support lensing due to unseen dark matter or anything of that sort). This means, whenever we're rendering a lensed galaxy, we are rendering *both* the lensed galaxy *and* the lensing galaxy. So, in reality, when we have a "lensed galaxy" in the code, it's actually 2 galaxies. This is why, when we have a lensing event, we have 7 parameters for the *lensing* galaxy and 12 for the *lensed* galaxy (7 to describe the galaxy itself and 5 for how it is to be distorted).

As for why to integrate it into the GalsimDecoder directly, it felt natural to say, if we have a galaxy that is involved with lensing, we can first rendering the galaxy and then render the galaxy it's lensing at the same time instead of doing a "second pass" to just render the lensed galaxies. So, I kept that the same as before.

Otherwise, I think I addressed most of the issues from the first PR! I didn't change some of the formatting stuff, cause I just ran black and kept the result of that. But I guess we can override that if we wanna

### DIFF
--- a/bliss/catalog.py
+++ b/bliss/catalog.py
@@ -38,6 +38,9 @@ class TileCatalog(UserDict):
         "matched",
         "mismatched",
         "detection_thresholds",
+        "lensed_galaxy_bools",
+        "lensed_galaxy_probs",
+        "lens_params",
     }
 
     def __init__(self, tile_slen: int, d: Dict[str, Tensor]):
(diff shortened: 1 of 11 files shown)


yashpatel5400:
Closing this, since I cleaned it up in another PR

PR #667, opened 2022-08-08
Encoder/decoder integration

yashpatel5400:
(Same as the previous PR, but cleaned up and now with all the tests passing)


bliss/encoder.py
```diff
@@ -163,20 +180,51 @@ def _encode_ptiles(self, image_ptiles: Tensor, n_samples: Optional[int]):
...
+                lensed_galaxy_probs *= is_on_array.unsqueeze(-1)
+
+                if deterministic:
+                    lensed_galaxy_bools = (
+                        lensed_galaxy_probs > 0.5
+                    ).float() * is_on_array.unsqueeze(-1)
+                else:
+                    lensed_galaxy_bools = (
+                        torch.rand_like(lensed_galaxy_probs) > 0.5
+                    ).float() * is_on_array.unsqueeze(-1)
```
jeff-regier, line 203: Black auto formatting isn't so great for long lines. It's often clearer to create a temporary/intermediate variable to reduce the length of long lines

bliss/encoder.py
```diff
@@ -163,20 +180,51 @@ def _encode_ptiles(self, image_ptiles: Tensor, n_samples: Optional[int]):
...
 
+            if self.lensing_binary_encoder is not None:
+                assert not self.lensing_binary_encoder.training
+                lensed_galaxy_probs = self.lensing_binary_encoder.forward(image_ptiles, locs)
+                lensed_galaxy_probs *= is_on_array.unsqueeze(-1)
+
+                if deterministic:
+                    lensed_galaxy_bools = (
+                        lensed_galaxy_probs > 0.5
+                    ).float() * is_on_array.unsqueeze(-1)
```
ismael-mendoza, line 199: better to split these lines into multiple statements (one per line)

bliss/encoder.py
```diff
@@ -163,20 +180,51 @@ def _encode_ptiles(self, image_ptiles: Tensor, n_samples: Optional[int]):
...
+                else:
+                    lensed_galaxy_bools = (
+                        torch.rand_like(lensed_galaxy_probs) > 0.5
+                    ).float() * is_on_array.unsqueeze(-1)
+
```
ismael-mendoza, line 204: same here

bliss/encoder.py
```diff
@@ -163,20 +180,51 @@ def _encode_ptiles(self, image_ptiles: Tensor, n_samples: Optional[int]):
...
+                lensed_galaxy_bools *= (
+                    galaxy_bools  # currently only support lensing where galaxy is present
+                )
```
ismael-mendoza, line 207: Better to put the comment in a separate line to avoid black formatting

bliss/encoder.py
```diff
@@ -163,20 +180,51 @@ def _encode_ptiles(self, image_ptiles: Tensor, n_samples: Optional[int]):
...
+
+                tile_samples.update(
+                    {
+                        "lensed_galaxy_bools": lensed_galaxy_bools,
+                        "lensed_galaxy_probs": lensed_galaxy_probs,
+                    }
+                )
```
ismael-mendoza, line 214: you might consider saving some lines by doing e.g. `tile_samples['lensed_galaxy_bools'] = lensed_galaxy_bools` instead

bliss/models/psf_decoder.py
```diff
@@ -0,0 +1,153 @@
...
+        sdss_bands: Bands to retrieve from PSF.
+        n_bands: Number of bands to retrieve from PSF.
+    """
+
+    def __init__(
+        self,
+        n_bands: int = 1,
+        psf_image_file: Optional[str] = None,
+        psf_params_file: Optional[str] = None,
+        psf_slen: Optional[int] = None,
```
ismael-mendoza, line 47: is this really optional?

bliss/models/prior.py
```diff
@@ -271,3 +303,35 @@ def _sample_galaxy_params(self, galaxy_bools):
...
+            batch_size,
+            n_tiles_h,
+            n_tiles_w,
+            max_sources,
+            n,
+            device=lensed_galaxy_bools.device,
+        )
+
+        lens_params = generate_lens_shape_tensor(5, torch.rand)
+        if self.prob_lensed_galaxy > 0.0:
```
ismael-mendoza, line 320: you don't need this `if` statement right? if `prob_lensed_galaxy == 0` then `lensed_galaxy_bools` will be all zeroes right?

bliss/models/prior.py
```diff
@@ -271,3 +303,35 @@ def _sample_galaxy_params(self, galaxy_bools):
...
+        generate_lens_shape_tensor = lambda n, dist: dist(
+            batch_size,
+            n_tiles_h,
+            n_tiles_w,
+            max_sources,
+            n,
+            device=lensed_galaxy_bools.device,
+        )
+
+        lens_params = generate_lens_shape_tensor(5, torch.rand)
```
ismael-mendoza, line 319: why do you use `torch.rand` rather than just create a vector of all zeroes?

bliss/models/prior.py
```diff
@@ -271,3 +303,35 @@ def _sample_galaxy_params(self, galaxy_bools):
...
+        """Sample latent galaxy params from GalaxyPrior object."""
+        batch_size, n_tiles_h, n_tiles_w, max_sources, _ = lensed_galaxy_bools.shape
+        generate_lens_shape_tensor = lambda n, dist: dist(
+            batch_size,
+            n_tiles_h,
+            n_tiles_w,
+            max_sources,
+            n,
+            device=lensed_galaxy_bools.device,
+        )
```
ismael-mendoza, line 317: I would avoid the use of lambda functions in this way (this one in particular takes up ~7 lines) and instead create a private method / function

bliss/models/prior.py
```diff
@@ -209,7 +235,13 @@ def _sample_n_galaxies_and_stars(self, is_on_array):
         galaxy_bools = uniform < self.prob_galaxy
         galaxy_bools = (galaxy_bools * is_on_array.unsqueeze(-1)).float()
         star_bools = (1 - galaxy_bools) * is_on_array.unsqueeze(-1)
-        return galaxy_bools, star_bools
+
+        # currently only support lensing where galaxy is present
+        lensed_galaxy_bools = uniform < self.prob_lensed_galaxy
+        lensed_galaxy_bools = (
+            lensed_galaxy_bools * galaxy_bools * is_on_array.unsqueeze(-1)
+        ).float()
```
ismael-mendoza, line 243: split into multiple statements

bliss/models/prior.py
```diff
@@ -142,23 +153,38 @@ def sample_prior(
...
+            )
+            pure_lens_params = self._sample_lens_params(lensed_galaxy_bools)
+            lens_params = torch.cat((lensed_galaxy_params, pure_lens_params), dim=-1)
+
+            catalog_params.update(
+                {
+                    "lensed_galaxy_bools": lensed_galaxy_bools,
+                    "lens_params": lens_params,
+                }
+            )
```
ismael-mendoza, line 185: save some lines by `catalog_params['lens_params'] = lens_params`

bliss/models/prior.py
```diff
@@ -142,23 +153,38 @@ def sample_prior(
         is_on_array = get_is_on_from_n_sources(n_sources, self.max_sources)
         locs = self._sample_locs(is_on_array)
 
-        galaxy_bools, star_bools = self._sample_n_galaxies_and_stars(is_on_array)
-        galaxy_params = self._sample_galaxy_params(galaxy_bools)
+        galaxy_bools, star_bools, lensed_galaxy_bools = self._sample_n_galaxies_and_stars(
+            is_on_array
+        )
```
ismael-mendoza, line 158: I would make a new function `_sample_n_lens` or something like that which takes `is_on_array` and `galaxy_bool` and returns `lensed_galaxy_bools`

bliss/models/galsim_encoder.py
```diff
@@ -190,7 +190,7 @@ def encode(self, image_ptiles: Tensor, tile_locs: Tensor) -> Tensor:
             ms=max_sources,
         )
 
-    def sample(self, image_ptiles: Tensor, tile_locs: Tensor):
+    def sample(self, image_ptiles: Tensor, tile_locs: Tensor, deterministic: Optional[bool]):
```
ismael-mendoza, line 193: why did you add this `deterministic` flag?

bliss/models/decoder.py
```diff
@@ -195,6 +195,15 @@ def _render_ptiles(self, tile_catalog: TileCatalog) -> Tensor:
             tile_catalog["star_fluxes"], "b nth ntw s band -> (b nth ntw) s band"
         )
 
+        if "lens_params" in tile_catalog:
```
ismael-mendoza, line 198: we should talk about avoiding this `if` statements scattered throughout the code, in general tihs makes the code harder to read and maintain. I was imagining separate `lens_decoder` just like we have for the stars would be cleaner to maintain. Ofc it would have to be conditioned on the galaxy_bools or something. Could you tell me if this is possible ?

bliss/models/decoder.py
```diff
@@ -580,7 +503,15 @@ def _render_single_galaxies(self, galaxy_params, galaxy_bools):
 
         # forward only galaxies that are on!
         # no background
-        gal_on = self.galaxy_decoder(z[b == 1])
+        if lens_params is not None:
```
ismael-mendoza, line 506: another `if` statement on `lens_params`, we should talk about whether we can remove this...

ismael-mendoza (changes requested):
Thanks @yashpatel5400 the code looks really good. My biggest suggestion is on the `if` statements scattered in the `decoder` related to the `lens_params` which I think will make the code harder to read/maintain in the long run. In some cases they can also prevent code optimization. 

Do you think there is a way we can get rid of them (I was imagining a separate `LensTileDecoder` but maybe this is unfeasible)? Let's discuss over slack or in-person if you are around west hall when you have a chance. 

The other comments are pretty minor and related to black formatting mostly

bliss/models/galsim_encoder.py
yashpatel5400: This was just to match the other GalaxyEncoder (since we support both in encoder.py)

bliss/models/prior.py
yashpatel5400: yeah, this was just to be explicit, but that's right

yashpatel5400:
alright, think I've addressed all the comments from above!

jeff-regier (approved):
Great!

PR #682, opened 2022-09-07
Refactor decoder.py

dereklhansen:


bliss/models/decoder.py
```diff
@@ -471,11 +464,10 @@ def __init__(self, tile_slen, ptile_slen, n_bands, galaxy_model: GalaxyModel):
...
-    def forward(self, locs, galaxy_params, galaxy_bools):
+    def forward(self, galaxy_params: Tensor, galaxy_bools: Tensor):
         """Renders galaxy tile from locations and galaxy parameters."""
         # max_sources obtained from locs, allows for more flexibility when rendering.
-        n_ptiles = locs.shape[0]
-        max_sources = locs.shape[1]
+        n_ptiles, max_sources, _ = galaxy_bools.shape
         n_galaxy_params = galaxy_params.shape[-1]
 
         galaxy_params = galaxy_params.view(n_ptiles, max_sources, n_galaxy_params)
```
ismael-mendoza, line 473: some of these assertion can be removed I think since they are redundant with the `.view`

bliss/models/decoder.py
```diff
@@ -471,11 +464,10 @@ def __init__(self, tile_slen, ptile_slen, n_bands, galaxy_model: GalaxyModel):
             raise TypeError("galaxy_model is not a valid type.")
         self.galaxy_decoder = galaxy_decoder
 
-    def forward(self, locs, galaxy_params, galaxy_bools):
+    def forward(self, galaxy_params: Tensor, galaxy_bools: Tensor):
```
ismael-mendoza, line 467: do you think we need the separate `_render_single_galaxies` method? `forward` does very little besides calling this method

bliss/models/decoder.py
```diff
@@ -223,21 +222,26 @@ def _render_ptiles(self, tile_catalog: TileCatalog) -> Tensor:
...
         # draw stars and galaxies
-        stars = self.star_tile_decoder(locs, star_fluxes, star_bools)
+        centered_stars = self.star_tile_decoder.forward(star_fluxes, star_bools)
+        stars = self.tiler.forward(locs, centered_stars)
         galaxies = torch.zeros(img_shape, device=locs.device)
         lensed_galaxies = torch.zeros(img_shape, device=locs.device)
 
         if self.galaxy_tile_decoder is not None:
-            galaxies = self.galaxy_tile_decoder(locs, tile_catalog["galaxy_params"], galaxy_bools)
+            centered_galaxies = self.galaxy_tile_decoder.forward(
```
ismael-mendoza, line 231: you changed the class name to just `GalaxyDecoder` but you still call the object `galaxy_tile_decoder`.

bliss/models/decoder.py
```diff
@@ -377,67 +374,63 @@ def _render_one_source(self, locs: Tensor, source: Tensor):
         grid_loc = local_grid - locs_swapped
         return F.grid_sample(source, grid_loc, align_corners=True)
 
-    def _expand_source(self, source: Tensor):
-        """Pad the source with zeros so that it is size ptile_slen."""
-        assert len(source.shape) == 3
 
-        slen = self.ptile_slen + ((self.ptile_slen % 2) == 0) * 1
-        assert len(source.shape) == 3
+def fit_source_to_ptile(source: Tensor, ptile_slen: int):
```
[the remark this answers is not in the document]
ismael-mendoza: since `fit_source_to_ptile` is not a method of the tiler anymore, you can consider renaming `fit_source_to_slen` since it doesn't necessarily has to be a padded tile we are fitting it to. It's a more general function

bliss/models/decoder.py
```diff
@@ -453,14 +446,14 @@ def forward(self, locs, fluxes, star_bools):
...
+        return sources
 
 
-class GalaxyTileDecoder(nn.Module):
+class GalaxyDecoder(nn.Module):
     def __init__(self, tile_slen, ptile_slen, n_bands, galaxy_model: GalaxyModel):
         super().__init__()
         self.n_bands = n_bands
-        self.tiler = Tiler(tile_slen, ptile_slen)
+        self.tiler = TileRenderer(tile_slen, ptile_slen)
```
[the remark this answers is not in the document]
ismael-mendoza: since you are changing `GalaxyTileDecoder` to `GalaxyDecoder` it will also make sense to replace `ptile_slen` with `slen`, then there is no reference to tiles for any of these three clases. I think this will make the classes conceptually cleaner.

ismael-mendoza (commented):
Looks good to me @dereklhansen ! Definitely an improvement toward making this module simpler compared to what we had. I had a few comments and suggestions, let me know what you think

bliss/models/decoder.py
ismael-mendoza: (not part of your PR so feel free to ignore this suggestion)

dereklhansen: Yes let's simplify the asserts below

bliss/models/decoder.py
dereklhansen: Good catch. Now we have `galaxy_decoder.galaxy_decoder` in a few places, which makes me question whether we need a separate `GalaxyDecoder` class at all. I'll raise a refactoring issue

bliss/models/decoder.py
dereklhansen: We can probably do away with this class altogether in a separate PR

PR #685, opened 2022-09-09
refactor PSF Decoder

ismael-mendoza:
Finishing up what I started during the hackathon.

Summary: 
- Remove option to use `.npy` file for `psf_params_file` since it is not currently used in repo
- combine myriad of methods in `PSFDecoder` to get psf from params file into a single function`_get_psf_from_params`
- normalization was done twice for no reason so that was removed


bliss/models/decoder.py
```diff
@@ -439,7 +439,7 @@ def forward(self, locs, fluxes, star_bools):
         n_ptiles = locs.shape[0]
         max_sources = locs.shape[1]
 
-        psf = self.forward_adjusted_psf()
+        psf = torch.from_numpy(self.psf.copy()).to(fluxes.device)
```
ismael-mendoza, line 442: unsure whether this is best replacement

ismael-mendoza:
Currently re-running all my case study to make sure all is working as before

bliss/models/psf_decoder.py
```diff
@@ -60,48 +59,43 @@ def __init__(
...
         raise NotImplementedError("Please extend this class and implement forward()")
 
-    def psf_forward(self):
-        psf = self._get_psf()
-        init_psf_sum = reduce(psf, "n m k -> n", "sum").detach()
-        norm = reduce(psf, "n m k -> n", "sum")
-        psf *= rearrange(init_psf_sum / norm, "n -> n 1 1")
+    def forward_psf_from_params(self):
+        # get psf in each band
+        psf_list = []
```
ismael-mendoza, line 81: add `self.params = None` and check this function and only be run if `self.params is not None`

ismael-mendoza:
All the results in my case_studies folder with real data are looking good, so I'm going to go ahead and merge this PR which greatly simplifies `PSFDecoder`. @dereklhansen if at some point you want to re-run your case studies with these changes and maybe the changes in #682 that might be a good idea

PR #686, opened 2022-09-14
figures for coadd neurips submission

ismael-mendoza:
Main push request adding functionality to produce results for BLISS on coadds targeting the  Neurips 2022 submission

- Make `apply_param_bin` constant time
- shell files for training all models 
- files to produce dataset and results 
- consistent seed identifying experiment
- flux posterior calibration
- plots for neurips submission and poster


ismael-mendoza:
all results from neurips can be reproduced with the code now added, and a README provides instructions. 

All tests are passing in CPU and my case studies folder `sdss_galaxies` works just like before so I'm merging. 

There is one more change I want to make for the posterior calibration plot which is to condition on true locations as mentioned in #696, will resolve this in a separate PR once #696 is closed

PR #695, opened 2022-11-08
Add deterministic sampling for Galsim and lens; case study scripts

yashpatel5400:
The GalsimEncoder and LensEncoder were missing a MAP (deterministic) sampling option, which is added here. I'm also checking in some scripts I used for generating the NeurIPS model .


yashpatel5400:
Ah boy, hang on, gotta fix these formatting issues

yashpatel5400:
Wait, ignore this for now -- will clean up and PR again

Issue #697, opened 2022-11-17
Separate `StarEncoder` class


ismael-mendoza:
It's not clear to me why we should treat stars differently than galaxies/lenses and have their flux be estimated in the `DetectionEncoder`. It would make more sense (I think) to have a separate `StarEncoder` that also centers the star in their padded tile before estimating their flux. 

I suspect we would gain two benefits from this (1) The detection encoder will have an easier time with the optimization since it can now only focus on detections rather than having to juggle (somewhat arbitrarily) the star flux optimization (2) The flux estimation for stars will be improved, since stars will now always be centered making the problem easier for the new encoder. 

This more or less hinges on deprecating StarNet support (#691) which I think we are more or less agreeing that it would be a good idea?

dereklhansen:
> (1) The detection encoder will have an easier time with the optimization since it can now only focus on detections rather than having to juggle (somewhat arbitrarily) the star flux optimization

With FAVI, I don't think this a big issue. In the loss function, the fluxes are optimized separately from the detections. There is some coupling in the architecture, since fluxes and locations come from the final layer of the same neural network.

dereklhansen:
That said, I can see how it could be better to have a separate `StarEncoder`

PR #699, opened 2022-11-19
U/im/coadd flux posterior calibration

ismael-mendoza:
Improvement over ml4science neurips results by conditioning (all) detection encoders on true counts per tiles of all stars for flux measurement. Thus allowing for comparison between encoders on flux measurement on source by source basis. 
Before, the sample was based on matched locations, so the sample that each encoder was tested on was different since the detections were different depending on number of coadds used (making comparison unfair). 

This is also in the spirit of evaluating the flux measurement independently of detections.

It makes sense to condition on star locations that might below detection threshold for some encoders because (1) The flux residual plot is per bin, so some bins can be ignored if desired (2) We want to actually compare each star between encoders, we want to test if the coadd encoders can extract more information from the coadd and give a better estimate of the flux. The only way to do this is if we test all the encoders on the same stars. 

This PR also demonstrates that is not completely straightforward to condition detection encoder on true locations. This suggest some refactoring is necessary or perhaps working on #697. 

I made it clear in the README how results from neurips submission can be reproduced (pointed to a commit). The main conclusions stayed the same (see figures below).

- [x] extend `to_tile_params` to allow for multiple batches
- [x] use detection encoder conditioned on true detections (counts per tile) to allow 1-1 comparison of fluxes for posterior calibration results and residual flux metric plot.

### DIFF
--- a/bliss/catalog.py
+++ b/bliss/catalog.py
@@ -402,33 +402,34 @@ class FullCatalog(UserDict):
             ValueError: If the number of sources in one tile exceeds `max_sources_per_tile` and
                 `ignore_extra_sources` is False.
         """
-        assert self.batch_size == 1, "Currently only supported for a single image"
-        tile_coords_fp = torch.div(self.plocs, tile_slen, rounding_mode="trunc")
-        tile_coords = tile_coords_fp.to(torch.int).squeeze(0)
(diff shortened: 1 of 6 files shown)


ismael-mendoza:
calibration

money

PR #717, opened 2023-03-02
refactoring in preparation to remove tiling

jeff-regier:
I made these refactoring changes while I was comparing encoders architectures. I think it's clearest if I put these refactoring changes in their own PR and merge it before making substantive changes to the encoder.

### DIFF
--- a/bliss/models/detection_encoder.py
+++ b/bliss/models/detection_encoder.py
@@ -11,6 +11,7 @@ from torch import Tensor, nn
 from torch.distributions import Categorical, Normal, Poisson
 from torch.nn import functional as F
 from torch.optim import Adam
+from torch.optim.lr_scheduler import MultiStepLR
 
 from bliss.catalog import TileCatalog, get_images_in_tiles, get_is_on_from_n_sources
 from bliss.models.encoder_layers import (
@@ -44,6 +45,7 @@ class DetectionEncoder(pl.LightningModule):
         annotate_probs: bool = False,
         slack: float = 1.0,
         optimizer_params: Optional[dict] = None,
+        scheduler_params: Optional[dict] = None,
     ):
         """Initializes DetectionEncoder.
 
@@ -60,7 +62,8 @@ class DetectionEncoder(pl.LightningModule):
             hidden: TODO (document this)
             annotate_probs: Annotate probabilities on validation plots?
             slack: Slack to use when matching locations for validation metrics.
-            optimizer_params: Optimizer for training.
+            optimizer_params: arguments passed to the Adam optimizer
+            scheduler_params: arguments passed to the learning rate scheduler
         """
         super().__init__()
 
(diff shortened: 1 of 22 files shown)


jeff-regier:
The most major change here is the addition of a new method called `encode_batch()` to the `DetectionEncoder`. Unlike `encode()` (which I've renamed to `encode_tiled()`) , the method `encode_batch()` takes a batch of *untiled* images as input. The current implementation of `encode_batch()` just tiles these images and runs them through the encoder, like we've always done. However, by overriding `encode_batch()` in a subclass of `DetectionEncoder`, I've been able to test a new encoder architecture that doesn't create padded tiles, which seems to work much better than our current encoder.

ismael-mendoza:
HI @jeff-regier - thanks for this PR. I will review it soon. 

> I've been able to test a new encoder architecture that doesn't create padded tiles, which seems to work much better than our current encoder.

I'm curious, what do you mean by "work much better"? What tests did you do? Thanks!

jeff-regier:
I'll open another PR soon with the new encoder but without removing the old one, so we can run both in parallel for a bit before deciding whether to switch. The new one is 4 times faster per iteration (it's fully convolutional, with skip connections, so it doesn't need to form padded tiles) and, at least at detection in synthetic data, it achieves significantly lower training/validation loss. But more validation is needed.

That said, the changes in this PR should be beneficial regardless of whether we adopt the new encoder.

pyproject.toml
```diff
@@ -28,8 +28,8 @@ hydra-core = ">=1.0.4"
 matplotlib = ">=3.3.3"
 nflows = ">=0.14"
 numpy = ">=1.18.5"
-python = "^3.8"
-pytorch-lightning = ">=1.5.1"
+python = "^3.10"
```
ismael-mendoza, line 31: does this mean you want to drop support for python 3.8 and 3.9 ? I'm OK with this but I was curious of your rationale?

bliss/models/detection_encoder.py
```diff
@@ -44,6 +45,7 @@ def __init__(
         annotate_probs: bool = False,
         slack: float = 1.0,
         optimizer_params: Optional[dict] = None,
+        scheduler_params: Optional[dict] = None,
```
ismael-mendoza, line 48: What do you use the scheduler for?

ismael-mendoza (approved):
Thanks @jeff-regier - I really like the simplification of `detection_encoder.py` you propose. I only have a few comments and questions so I think we can merge. Have you run the tests on the GPU?

bliss/models/detection_encoder.py
jeff-regier: I've been using it to drop the learning rate from 1e-3 to 1e-4 after 10 or so epochs

jeff-regier: I'm also really interested in trying out the [OneCycleLR](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.OneCycleLR.html) scheduler.

PR #718, opened 2023-03-08
fully convolutional detection encoder

jeff-regier:
This PR adds a case study that demonstrates a new fully convolutional encoder architecture. Because it doesn't form padding tiles, this encoder is 4 times than our current detection encoder (>6 training iterations per second, whereas the old detection encoder performed about 1.5 iterations / second). The graphs below suggest that it is also more accurate. The old encoder, which was trained for 24 hours, appears in grey. The new encoder, which was trained for 6 hours, appears in red.

Screenshot 2023-03-08 at 10 04 28 AM

Screenshot 2023-03-08 at 10 04 37 AM

### DIFF
--- a/bliss/train.py
+++ b/bliss/train.py
@@ -32,7 +32,7 @@ def train(cfg: DictConfig):
     dataset = instantiate(cfg.training.dataset)
 
     # setup model
-    model = instantiate(cfg.training.model, optimizer_params=cfg.training.optimizer_params)
+    model = instantiate(cfg.training.model)
 
     # setup trainer
     logger = setup_logger(cfg, paths)
@@ -125,7 +125,6 @@ def log_hyperparameters(config, model, trainer) -> None:
     # choose which parts of hydra config will be saved to loggers
     hparams["mode"] = config["mode"]
     hparams["training"] = config["training"]
-    hparams["optimizer"] = config["training"]["optimizer_params"]
 
     # save number of model parameters
     hparams["model/params_total"] = sum(p.numel() for p in model.parameters())
--- a/case_studies/new_detection_encoder/config/config.yaml
+++ b/case_studies/new_detection_encoder/config/config.yaml
@@ -1,0 +1,163 @@
+defaults:
+    - _self_
+    - training:
+    - override hydra/job_logging: stdout
+
+# completely disable hydra logging
+# https://github.com/facebookresearch/hydra/issues/910
+hydra:
+    output_subdir: null
+    run:
+        dir: .
+
+mode: train
+
+paths:
+    root: ${oc.env:BLISS_HOME}
+    sdss: ${paths.root}/data/sdss
+    data: ${paths.root}/data
+    project: ${paths.root}/case_studies/new_detection_encoder
+    output: ${paths.project}/output
+
+datasets:
+    simulated:
+        _target_: bliss.datasets.simulated.SimulatedDataset
+        prior: ${models.prior}
+        decoder:
+            _target_: bliss.models.decoder.ImageDecoder
+            n_bands: ${models.decoder.n_bands}
+            tile_slen: ${models.decoder.tile_slen}
+            ptile_slen: ${models.decoder.ptile_slen}
+            border_padding: ${models.decoder.border_padding}
+            psf_slen: ${models.decoder.psf_slen}
+            sdss_bands:
+                - 2
+            psf_params_file: ${models.decoder.psf_params_file}
+            galaxy_model:
+        background:
+            _target_: bliss.datasets.simulated.SimulatedSDSSBackground
+            sdss_dir: ${paths.sdss}
+            run: 94
+            camcol: 1
+            field: 12
+            bands:
+                - 2
+        n_tiles_h: 8
+        n_tiles_w: 8
+        n_batches: 128
+        batch_size: 64
+        valid_n_batches: 50
+        generate_device: "cpu"
+        num_workers: 5
+        fix_validation_set: true
+
+models:
+    decoder:
+        _target_: bliss.models.decoder.ImageDecoder
+        n_bands: 1
+        tile_slen: 4
+        ptile_slen: 52
+        border_padding: 24
+        psf_slen: 25
+        sdss_bands:
+            - 2
+        psf_params_file: ${paths.data}/sdss/94/1/12/psField-000094-1-0012.fits
+
+    detection_encoder:
+        _target_: case_studies.new_detection_encoder.main.NewDetectionEncoder
+        #  _target_: bliss.models.detection_encoder.DetectionEncoder
+        input_transform:
+            _target_: bliss.models.detection_encoder.ConcatBackgroundTransform
+        n_bands: ${models.decoder.n_bands}
+        tile_slen: ${models.decoder.tile_slen}
+        ptile_slen: 52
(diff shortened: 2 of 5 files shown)


bliss/train.py
```diff
@@ -32,7 +32,7 @@ def train(cfg: DictConfig):
     dataset = instantiate(cfg.training.dataset)
 
     # setup model
-    model = instantiate(cfg.training.model, optimizer_params=cfg.training.optimizer_params)
+    model = instantiate(cfg.training.model)
```
jeff-regier, line 35: It seems simpler to include the optimizer parameter in `cfg.training.model`, particularly now that I've introduced parameters for the scheduler too. But please lmk if I'm missing something.

pyproject.toml
```diff
@@ -35,6 +35,7 @@ scipy = ">=1.4.1"
 seaborn = ">=0.11.2"
 torch = ">=1.9"
 torchmetrics = ">=0.5.1"
+yolov5 = "^7.0.9"
```
ismael-mendoza, line 38: did you run `poetry update` or just `poetry add yolo5`? and in the last PR?

case_studies/sdss_galaxies_vae/config/config.yaml
```diff
@@ -132,23 +134,31 @@ models:
...
+        optimizer_params:
+            lr: 1e-3
     galaxy_vae:
         _target_: bliss.models.vae.galaxy_vae.OneCenteredGalaxyVAE
         slen: 53
         latent_dim: 32
         hidden: 256
         n_bands: 1
+        optimizer_params:
+            lr: 1e-4
```
ismael-mendoza, line 161: I think this should be `1e-3`

case_studies/sdss_galaxies/config/config.yaml
```diff
@@ -99,6 +101,8 @@ models:
         psf_slen: 25
         sdss_bands:
             - 2
+        optimizer_params:
+            lr: 1e-3
```
ismael-mendoza, line 105: I think the `ImageDecoder` does not require `optimizer_params`?

ismael-mendoza: maybe you meant to include it within the `galaxy_model` block? (Although note that the `OneCenteredGalaxyAE` is not trained when within the decoder, it's kept fixed with the `ckpt` given)

case_studies/new_detection_encoder/main.py
```diff
@@ -0,0 +1,54 @@
...
+        super().__init__(**ode_args)
+
+        # the checkerboard sizes below only make sense for images composed of an
+        # 8x8 grid of 4x4 tiles. It's straightforward to generalize it to arbitrary
+        # image/tile sizes, but I done so because I haven't determined yet whether
+        # this checkerboard is helpful/necessary
+        shape = (20, 20)
+        base_checkerboard = np.indices(shape).sum(axis=0) % 2
+        self.checkerboard = torch.from_numpy(base_checkerboard).unsqueeze(2).repeat([1, 4, 4])
+        self.checkerboard = self.checkerboard.view(1, 1, 80, 80).float().cuda()
```
ismael-mendoza, line 28: I'm struggling to understand this checker board and what you are doing here. But maybe you can explain in the meeting. 

Also what do you mean by "images composed of an 8x8 grid of 4x4 tiles."?

case_studies/new_detection_encoder/main.py
```diff
@@ -0,0 +1,54 @@
...
+
+        arch_dict = OmegaConf.to_container(architecture)
+        self.model = DetectionModel(cfg=arch_dict, ch=3)
+
+    def do_encode_batch(self, images_with_background):
+        b = images_with_background.size(0)
+        batch_checkerboard = self.checkerboard.expand([b, 1, -1, -1])
+        x = torch.cat([images_with_background, batch_checkerboard], dim=1)
+        self.model.model[-1].training = True
```
ismael-mendoza, line 37: Why do you turn this on every time you encode a batch? What does `model.model[-1]` represent?

ismael-mendoza (changes requested):
Thanks @jeff-regier for this PR. In terms of changes, there is only very minor ones in one of the config files. The other comments are just questions which you can feel free to delay until the meeting today.

case_studies/new_detection_encoder/config/config.yaml
```diff
@@ -0,0 +1,163 @@
...
+                [[-1, 4, 5], 1, Concat, [1]],  # cat backbone P3
+                [-1, 3, C3, [256, false]],  # 17 (P3/8-small)
+
+                [[17], 1, Detect, [nc, anchors]],  # Detect(P3, P4, P5)
+            ]
+    prior:
+        _target_: bliss.models.prior.ImagePrior
+        n_bands: 1
+        max_sources: 1
+        mean_sources: 0.6931471805599453  # 50% no source, 50% one source
```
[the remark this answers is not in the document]
jeff-regier: It is more like m2. I just picked it because I thought it would train faster than if most tiles were empty. But indeed, we should test the new encoder in more realistic settings too.

case_studies/new_detection_encoder/config/config.yaml
```diff
@@ -0,0 +1,163 @@
...
+        dropout: 0.0
+        hidden: 128
+        annotate_probs: true
+        slack: 1.
+
+        optimizer_params:
+            lr: 1e-3
+        scheduler_params:
+            milestones: [8, 24]
+            gamma: 0.1
```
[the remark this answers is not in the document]
jeff-regier: Decreasing it twice seems to be overkill, but once seems to be helpful. (although I need to view more replicates to be sure)

pyproject.toml
jeff-regier: I ran `poetry update` in the previous PR, which unfortunately seem to have broken the `case studies` workflow. 

In this PR I just ran `poetry add yolov5`. We don't really use a lot of yolov5, so I also could have copied/pasted some of the yolov5 code into the bliss repo. We may want to do that in the future so it's easier to modify.

case_studies/sdss_galaxies/config/config.yaml
jeff-regier: oops. fixed.

case_studies/new_detection_encoder/main.py
jeff-regier: It's a hack that lets us use a function in yolov5 for a purpose it wasn't exactly indented for. Otherwise, we'd need to copy the relevant code from yolov5 to our repo, so we can modify it, which may ultimately be good idea.

ismael-mendoza (approved):
@jeff-regier  I think you can merge now. Feel free to delete the checkerboard later or not depending on your experiments.

jeff-regier:
The checkerboard wasn't helpful/necessary after all. Deleted. I fixed the config files too.

jeff-regier:
I improved the learning rate schedule for the new architecture. (Dropped one of the milestones.) Here are results for the old (grey) and new (red) encoders, both trained for 24 hours. The new encoder finishes 4x as many iterations during that time, but even per iteration it appears to have an edge.
Screenshot 2023-03-10 at 9 39 22 AM

Screenshot 2023-03-10 at 9 39 57 AM

Issue #721, opened 2023-04-19
speed up the ImageDecoder


jeff-regier:
`ImageDecoder` is quite slow currently. Even with 32 workers, it can't keep up with the encoder. GalSim's `ScatteredImageBuilder` may less us write batches of light sources more efficiently. It could also be useful to look at how [imSim](https://github.com/LSSTDESC/imSim) interacts with GalSim.

If we can't speed up `ImageDecoder` by at least 10x, then as an alternative we can generate simulated training images ahead of time and write them to disk. We may also want to use data augmentation in this case to make more of each image we generate: apply random 90-degree rotations and small translations. Such data augmentation would need to be reflected in the tile catalog too.

Steps to using cached images:
- [x] Modify `case_studies/summer_template/main.py` to support a new mode: `generate`
- [x] Create a new file named `bliss/generate.py` that is in some sense analogous to `predict.py` and `train.py`. It would contain a function called `generate(...)` that takes `cfg` as an argument
- [x] When called `generate(cfg)` should create a `SimulatedDataset` object (using the `instantiate` function provided by hydra, as in `train.py`), generate a lot of data, serialized the data, and write the data to a file whose name is specified in the `cfg` object.
- [x] `case_studies/summer_template/config.yaml` should probably have a new top-level entry called `cached_simulator` (analogous to `simulator`, but with many fewer fields). This is where we'd store that path the filename (or directory) that contains the cache images
- [x] `simulated_dataset.py` should contain an additional class called `CachedSimulatedDataset`, with a constructor that takes a filename (or directory path) of the cached images and loads the file into memory.
- [x] `CachedSimulatedDataset` won't use any workers because it's just looking up loaded data. It will provide minibatches that are sampled at random from the available cached images.

PR #738, opened 2023-05-03
R/W generated dataset to/from disk

zhixiangteoh:
Two-step process.

1. Generate; e.g.,
```sh
python case_studies/721_decoder_speedup/main.py mode=generate cached_simulator.file_data_capacity=100 simulator.n_batches=5
```
`generator.py` instantiates and uses a `SimulatedDataset` object to render images in `cfg.simulator.prior.batch_size` minibatches, then processes these generated minibatches by concatenation and flattening, before writing out to `.pkl` files via `pickle`. 

2. Train; e.g.,
```sh
python case_studies/721_decoder_speedup/main.py mode=train training.use_cached_simulator=true cached_simulator.file_data_capacity=100 simulator.n_batches=5
```
`training.use_cached_simulator=true` config causes the training step to use `datamodule=instantiate(cfg.cached_simulator)` in PyTorch Lightning's `trainer.fit`.

Addresses #721.


case_studies/721_decoder_speedup/config.yaml
```diff
@@ -0,0 +1,150 @@
...
+        accelerator: "gpu"
+        devices: 1
+        limit_train_batches: 1.0
+        limit_val_batches: 1.0
+        check_val_every_n_epoch: 10
+        log_every_n_steps: 10  # corresponds to n_batches
+    testing:
+        file: null
+        batch_size: 32
+        num_workers: 0  # why not increase this?
```
jeff-regier, line 135: Is `num_workers` here still needed now that we're loading images from disk?

bliss/simulator/simulated_dataset.py
```diff
@@ -87,3 +89,77 @@ def val_dataloader(self):
...
+            self,
+            batch_size=self.simulator.image_prior.batch_size,
+            num_workers=self.simulator.num_workers,
+        )
+
+    def val_dataloader(self):
+        return DataLoader(
+            self,
+            batch_size=self.simulator.image_prior.batch_size,
+            num_workers=self.simulator.num_workers,
```
jeff-regier, line 157: Seems like `num_workers` here should be zero, not the number used for generating new data

bliss/simulator/simulated_dataset.py
```diff
@@ -87,3 +89,77 @@ def val_dataloader(self):
 
     def test_dataloader(self):
         return DataLoader(self, batch_size=None, num_workers=self.num_workers)
+
+
+class CachedSimulatedDataset(pl.LightningDataModule, Dataset):
+    def __init__(
+        self,
+        simulator: SimulatedDataset,
```
jeff-regier, line 97: Better to just pass `batch_size` to the constructor, rather than passing it the whole simulator. We may want to use a different `batch_size` when simulating data than when loading cached data. And anyway, it seems cleaner to decouple `CachedSimulatedDataset` from `SimulatedDataset`

bliss/generate.py
```diff
@@ -0,0 +1,98 @@
...
+            assert len(flat_data["images"]) == file_data_size
+            assert len(flat_data["background"]) == file_data_size
+            for i in range(file_data_size):
+                file_datum: FileDatum = {}  # type: ignore
+                for key, _ in flat_data.items():
+                    file_datum[key] = (  # type: ignore
+                        {k: flat_data[key][k][i] for k in flat_data[key].keys()}
+                        if key == "tile_catalog"
+                        else flat_data[key][i]
+                    )
```
jeff-regier, line 93: the previous line is a bit hard to make sense of. Can it be simplified?

bliss/generate.py
```diff
@@ -0,0 +1,98 @@
...
+    simulated_dataset = simulator.train_dataloader().dataset
+    assert isinstance(
+        simulated_dataset, IterableDataset
+    ), "simulated_dataset must be IterableDataset"
+
+    # create cached_data_path if it doesn't exist
+    if not os.path.exists(cached_data_path):
+        os.makedirs(cached_data_path)
+
+    if len(data_files) != n_files:
```
jeff-regier, line 48: When is this false? Only if `n_files == 0`? If so, it seems clearer to just use `assert n_files > 0`

jeff-regier: (also, then everything below wouldn't have to be indented as much)

bliss/generate.py
```diff
@@ -0,0 +1,98 @@
...
+            tile_catalog_flattened = {}
+            for key in batch_data[0]["tile_catalog"].keys():
+                tile_catalog_flattened[key] = (
+                    torch.flatten(
+                        torch.cat([data["tile_catalog"][key] for data in batch_data]),
+                        start_dim=0,
+                        end_dim=1,
+                    )
+                    if len(batch_data) > 1
+                    else batch_data[0]["tile_catalog"][key]
```
jeff-regier, line 72: the black autoformatted if-else blocks aren't so easy to read. Can it be written in a way that's easier to understand?

bliss/generate.py
```diff
@@ -0,0 +1,98 @@
...
+                )
+                flat_data["tile_catalog"] = tile_catalog_flattened
+            for key in batch_data[0].keys():
+                if key != "tile_catalog":
+                    flat_data[key] = (
+                        torch.flatten(
+                            torch.cat([data[key] for data in batch_data]), start_dim=0, end_dim=1
+                        )
+                        if len(batch_data) > 1
+                        else batch_data[0][key]
```
jeff-regier, line 82: same here

jeff-regier (changes requested):
Looks good! Please see requested changes inline.

jeff-regier:
Also, please make sure training iterations / second is at least 6.

case_studies/721_decoder_speedup/config.yaml
```diff
@@ -0,0 +1,150 @@
...
+        run: 94
+        camcol: 1
+        field: 12
+        bands: [2]
+
+cached_simulator:
+    _target_: bliss.simulator.simulated_dataset.CachedSimulatedDataset
+    simulator: ${simulator}
+    file_data_capacity: 5000
+    cached_data_path: ${paths.data}/cached_dataset
```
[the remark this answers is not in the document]
zhixiangteoh: Done - see #744

bliss/generate.py
zhixiangteoh: Done.

bliss/generate.py
zhixiangteoh: Done.

bliss/generate.py
zhixiangteoh: Done.

bliss/generate.py
zhixiangteoh: Done.

bliss/train.py
```diff
@@ -29,7 +29,11 @@ def train(cfg: DictConfig):
     pl.seed_everything(cfg.training.seed)
 
     # setup dataset.
-    simulator = instantiate(cfg.simulator)
+    dataset = (
+        instantiate(cfg.cached_simulator)
+        if cfg.training.use_cached_simulator
+        else instantiate(cfg.simulator)
+    )
```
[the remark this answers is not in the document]
zhixiangteoh: Done.

bliss/generate.py
```diff
@@ -0,0 +1,98 @@
...
+
+    file_data_capacity = cfg.cached_simulator.file_data_capacity
+    cached_data_path = cfg.cached_simulator.cached_data_path
+
+    # largest `batch_size` multiple <= `file_data_capacity`
+    file_data_size = (
+        file_data_capacity // simulator.image_prior.batch_size
+    ) * simulator.image_prior.batch_size
+
+    assert file_data_size >= simulator.image_prior.batch_size, "file_data_capacity too small"
```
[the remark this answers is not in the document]
zhixiangteoh: Would this be too restrictive? Assuming the workers do the work completely in parallel—for now, do we not want to just abstract away when/how the workers are generating the images and when/how these `batch_size` minibatches of images are combined into files?

For example, if 4 workers are generating images in parallel, but `file_data_size == 2 * batch_size`, if `n_files == 1` then we know for sure that 2 workers end up doing unnecessary work. But I'm not sure what happens if, say, `n_files == 2`—perhaps there's no unnecessary generation in this case?

bliss/simulator/simulated_dataset.py
zhixiangteoh: Done.

bliss/generate.py
```diff
@@ -0,0 +1,98 @@
...
+def generate(cfg: DictConfig):
+    simulator = instantiate(cfg.simulator)
+
+    file_data_capacity = cfg.cached_simulator.file_data_capacity
+    cached_data_path = cfg.cached_simulator.cached_data_path
+
+    # largest `batch_size` multiple <= `file_data_capacity`
+    file_data_size = (
+        file_data_capacity // simulator.image_prior.batch_size
+    ) * simulator.image_prior.batch_size
```
jeff-regier, line 28: Would you write this a bit more simply, e.g., 
```
    bs = simulator.image_prior.batch_size
    file_data_size = (file_data_capacity // bs) * bs
```

bliss/generate.py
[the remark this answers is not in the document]
jeff-regier: Ah I think I misunderstood the meaning of `file_data_size`.

bliss/generate.py
zhixiangteoh: Done.

bliss/generate.py
```diff
@@ -0,0 +1,98 @@
...
+    # use SimulatedDataset to generate data in minibatches iteratively,
+    # then concatenate before caching to disk via pickle
+    simulated_dataset = simulator.train_dataloader().dataset
+    assert isinstance(
+        simulated_dataset, IterableDataset
+    ), "simulated_dataset must be IterableDataset"
+
+    # create cached_data_path if it doesn't exist
+    if not os.path.exists(cached_data_path):
+        os.makedirs(cached_data_path)
```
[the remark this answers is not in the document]
zhixiangteoh: Done.

bliss/simulator/simulated_dataset.py
zhixiangteoh: Done.

zhixiangteoh:
> Also, please make sure training iterations / second is at least 5.5.

Currently, averaging to about 4 training iterations / second (in each epoch, starts out slower, then speeds up to > 5 it/s).

bliss/simulator/simulated_dataset.py
```diff
@@ -87,3 +91,74 @@ def val_dataloader(self):
...
+                self.data += self.read_file(f"{self.cached_data_path}/{filename}")
+        assert self.data, "No cached data loaded; run `generate.py` first"
+
+    def read_file(self, filename: str) -> List[FileDatum]:
+        with open(filename, "rb") as f:
+            return pickle.load(f)
+
+    def get_batch(self, batch_idx) -> Dict:
+        batch_data = self.data[batch_idx * self.batch_size : (batch_idx + 1) * self.batch_size]
+        images = torch.stack([datum["images"] for datum in batch_data])
```
jeff-regier, line 133: If we need this code to be more efficient (i.e., to keep up with the GPU), we should consider storing the images together in memory by putting them all one giant tensor. (Same for the background and the tile catalogs.) As is, we're storing references to the images consecutively, but not the images themselves. So we still end up accessing memory in a fairly random order. Which may or may not be fast enough.

zhixiangteoh: I see, yea this makes sense. You mean to say upon shuffling (the references), the accesses are spatially distant, and we are not exploiting spatial locality when we could be?

jeff-regier: Yes, that's what I had in mind.

Another idea: we could leave it as is an try using multiple workers after all. Not a lot, maybe 2 or 4. Given all the 4 loops in `get_batch` we may benefit from a few workers after all.

jeff-regier: It would be more elegant to exploit spatial locality, but whatever keeps the GPU busy is fine :-)

zhixiangteoh:
> Also, please make sure training iterations / second is at least 5.5.

With configuration as in `case_studies/721_diskrw_gsimages/config.yaml`, running

```
python case_studies/721_diskrw_gsimages/main.py 'mode=train'
```

gives median 6.1 it/s.

bliss/simulator/simulated_dataset.py
```diff
@@ -87,3 +91,76 @@ def val_dataloader(self):
...
+
+        # assume cached image files exist, read from disk
+        for filename in os.listdir(self.cached_data_path):
+            if "valid" in filename:
+                continue
+            if filename.startswith("dataset") and filename.endswith(".pkl"):
+                self.data += self.read_file(f"{self.cached_data_path}/{filename}")
+        assert self.data, "No cached data loaded; run `generate.py` first"
+
+        if self.fix_validation_set:
```
jeff-regier, line 128: Probably better to assume that we're always using a fixed validation set, if we're using a fixed training set.

bliss/simulator/simulated_dataset.py
```diff
@@ -87,3 +91,76 @@ def val_dataloader(self):
...
+            yield self.get_batch(batch_idx)
+
+    def train_dataloader(self):
+        return DataLoader(self, batch_size=None, num_workers=0)
+
+    def val_dataloader(self):
+        return DataLoader(self.valid, batch_size=None, num_workers=0)
+
+    def test_dataloader(self):
+        return DataLoader(self, batch_size=None, num_workers=0)
```
jeff-regier, line 166: Is the test set the same as the training set? That seems like a problem. We should probably have separate files that contain the test set.

jeff-regier:
> With configuration as in `case_studies/721_diskrw_gsimages/config.yaml`, running
> 
> ```
> python case_studies/721_diskrw_gsimages/main.py 'mode=train'
> ```
> 
> gives median 6.1 it/s.

Excellent!

tests/testing_config.yaml
```diff
@@ -18,6 +18,10 @@ simulator:
         _target_: bliss.simulator.background.ConstantBackground
         background: [865.0]
 
+cached_simulator:
+    file_data_capacity: 50
+    cached_data_path: ${paths.data}/test_cached_dataset
```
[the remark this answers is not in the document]
zhixiangteoh: Yep that sounds good; I manually checked that the tmp path does get created by pytest fixture, and this fail-safe is probably extraneous/useless.

bliss/simulator/simulated_dataset.py
zhixiangteoh: Done.

bliss/simulator/simulated_dataset.py
zhixiangteoh: Done.

tests/testing_config.yaml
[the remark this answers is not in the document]
zhixiangteoh: Done.

jeff-regier (approved):
Looks great!

PR #744, opened 2023-05-05
Add base hydra config to be extended

### LINKED ISSUE #739: shared base config file
Currently each case study has its own hydra config file and the unit tests have their own config file, and a lot of settings are duplicated. To add an additional attribute to a class, for example, you have to change all the config files separately. There must be a better way, perhaps through Hydra's [extending configs](https://hydra.cc/docs/patterns/extending_configs/) functionality. Perhaps we can start a `conf` directory that contains a `base_config.yaml` file, which gets used (potentially with modifications/extensions) everywhere.

zhixiangteoh:
Reuse / extend `{BLISS_HOME}/conf/base_config.yaml` anywhere. Given following directory structure:

```sh
├── bliss
├── case_studies
│   └── summer_template
│       └── config.yaml
├── conf
│   └── base_config.yaml
└── tests
    └── basic_config.yaml
```

we can set defaults in `conf/base_config.yaml` like so:

```yml
# {BLISS_HOME}/conf/base_config.yaml
---
...
mode: train

paths:
    root: ${oc.env:BLISS_HOME}
    data: ${paths.root}/data
    sdss: ${paths.data}/sdss
    output: ${paths.root}/output
    pretrained_models: ${paths.data}/pretrained_models

...
```

and override/augment variables in `case_studies/summer_template/config.yaml` like so:

```yml
# {BLISS_HOME}/case_studies/summer_template/config.yaml
---
defaults:
    - ../../conf@_here_: base_config
    - _self_
    - override hydra/job_logging: stdout

paths:
    project: ${paths.root}/case_studies/summer_template
    output: ${paths.project}/output
    pretrained_models: ${paths.data}/pretrained_models

...
```

where `paths.output: ${paths.project}/output` here overrides the corresponding default value from `conf/base_config.yaml`. 

Note:
1. relative path to default config is used, as in `../../conf@_here_: base_config`
2. the order of defaults is important; here, we 'apply' the defaults from `base_config` before overriding values defined in this config (`_self_`)

Fixes #739.

### DIFF
--- a/case_studies/summer_template/config.yaml
+++ b/case_studies/summer_template/config.yaml
@@ -1,57 +1,15 @@
 ---
 defaults:
+    - ../../conf@_here_: base_config
     - _self_
     - override hydra/job_logging: stdout
 
-hydra:
-    output_subdir: null
-    run:
-        dir: .
-
-mode: train
-
 paths:
-    root: ${oc.env:BLISS_HOME}
-    data: ${paths.root}/data
-    sdss: ${paths.data}/sdss
     project: ${paths.root}/case_studies/summer_template
     output: ${paths.project}/output
     pretrained_models: ${paths.data}/pretrained_models
 
 simulator:
-    _target_: bliss.simulator.simulated_dataset.SimulatedDataset
-    n_batches: 128
-    valid_n_batches: 10  # 256
-    fix_validation_set: true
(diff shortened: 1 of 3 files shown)


jeff-regier (approved):
Excellent!

PR #746, opened 2023-05-08
functional test

### LINKED ISSUE #722: round-trip functional tests
Two things to check with Pytest:

- [x] If we simulate small images with various combinations of one, two, or three stars and galaxies, the encoder with pretrained weights (currently stored in `data/pretrained_models`) correctly predicts the locations and properties of these sources.
- [x] If we predict the properties of the light sources in an SDSS image using this encoder, and then reconstruct the image based on these properties, the residual between the original SDSS image and the reconstruction is low (e.g., < 1% of total flux).

XinyueLi1012:
closes #722


bliss/predict.py
```diff
@@ -11,17 +11,13 @@ def prepare_image(x, device):
     return x[:, :, :height, :width]
 
 
-def predict(cfg):
+def predict(cfg, image, background):
```
jeff-regier, line 14: Nice! This generalization of predict will help with many use cases.

tests/test_end_to_end.py
```diff
@@ -11,4 +13,7 @@ def test_train(self, cfg):
         train(cfg)
 
     def test_predict(self, cfg):
-        predict(cfg)
+        sdss = instantiate(cfg.predict.dataset)
+        prepare_img = prepare_image(sdss[0]["image"], cfg.predict.device)
+        prepare_bg = prepare_image(sdss[0]["background"], cfg.predict.device)
+        predict(cfg, prepare_img, prepare_bg)
```
jeff-regier, line 19: It'd be good to avoid duplicating these 4 lines of code (both here and in main.py). Perhaps make a function called `predict_sdss(cfg)` in `predict.py` that is called here and in main.py, so that code isn't duplicated

tests/test_sdss_reconstruct.py
```diff
@@ -0,0 +1,36 @@
...
+
+class TestSdssReconstrust:
+    def test_sdss_reconstrst(self, cfg):
+        sdss = instantiate(cfg.predict.dataset)
+        true_img = sdss[0]["image"][:, 160:320, 160:320]
+        true_bg = sdss[0]["background"][:, 160:320, 160:320]
+        prepare_img = prepare_image(true_img, cfg.predict.device)
+        prepare_bg = prepare_image(true_bg, cfg.predict.device)
+        est_full = predict(cfg, prepare_img, prepare_bg)
+        est_tile = est_full.to_tile_params(cfg.encoder.tile_slen, cfg.simulator.prior.max_sources)
```
jeff-regier, line 15: Perhaps `predict_sdss(cfg)` could be called here too? If so, we'd need the config file include the ranges 160:320 and 160:320.

jeff-regier (changes requested):
Looks good. Please make the changes above and rebase on master.

jeff-regier (approved):
Looks good! Thanks for your work.

PR #748, opened 2023-05-08
bug fix and remove risky cropping code

jeff-regier:
This PR fixes the star flux bug and includes a newly pretrained `sdss.pt` model in `data/pretrained_models/`.

@aakashdp6548 Please lmk if this fixes the issue with the test you are writing.

@XinyueLi1012 Please lmk if you can increase the threshold in your test following this PR.


aakashdp6548:
Looks good to me. The `star_fluxes` values are more reasonable, and taking the top 100 gives me pretty good results:
cat_comparison_top100

Also fwiw, the number of sources is also a little bit better - 197 predicted compared to 200 in SDSS.

Issue #737, opened 2023-05-02
replace linear_sum_assignment with min_weight_full_bipartite_matching


jeff-regier:
In metrics.py, this call to `linear_sum_assignment` can be slow:
https://github.com/prob-ml/bliss/blob/5dae3d7d784ac6e92ddf015bc36bcc2b855f0097/bliss/metrics.py#L202

But we can replace it with this sparse version:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.min_weight_full_bipartite_matching.html#scipy.sparse.csgraph.min_weight_full_bipartite_matching

The similarity between predicted/true light sources more than `slack` pixels apart can be set to zero.

Issue #723, opened 2023-04-19
improve metrics and plotting


jeff-regier:
- [x] Profile metrics.py to figure out why some metrics are so slow; speed the metrics calculation up and/or identify if there’s something inherent to task that is slow.  If there is an inherently slow piece, we can run it less frequently.
- [x] Identify metrics that aren't generally useful and delete them
- [x] for the remaining metrics,  log them all to the tensorboard metrics by default
- [x] ditto for plotting -- delete unnecessary plotting routines and have plots from the rest included in tensorboard by default every x epochs

PR #749, opened 2023-05-08
Issues #737, #723 - Optimizing/cleaning metrics, removing unnecessary plotting functions

sawanp813:


bliss/metrics.py
```diff
@@ -40,50 +43,60 @@ def __init__(
             fp: false positives = # of predicted sources not matched with true source
             avg_distance: Average l-infinity distance over matched objects.
             total_true_n_sources: Total number of true sources over batches seen.
-            total_correct_class: Total # of correct classifications over matched objects.
-            total_n_matches: Total # of matches over batches.
-            conf_matrix: Confusion matrix (galaxy vs star)
             disable_bar: Whether to show progress bar
+            gal_tp: true positives = # of sources correctly classified as a galaxy
```
jeff-regier, line 47: Could we rename `tp` and `fp` something like `detection_tp` and `detection_fp` to better distinguish them from `gal_tp` and `gal_fp`?

bliss/metrics.py
```diff
@@ -40,50 +43,60 @@ def __init__(
...
         for b in tqdm(range(true.batch_size), desc=desc, disable=self.disable_bar):
             ntrue, nest = true.n_sources[b].int().item(), est.n_sources[b].int().item()
             tlocs, elocs = true.plocs[b], est.plocs[b]
+            tgbool, egbool = true["galaxy_bools"][b].reshape(-1), est["galaxy_bools"][b].reshape(-1)
             if ntrue > 0 and nest > 0:
                 mtrue, mest, dkeep, avg_distance = match_by_locs(tlocs, elocs, self.slack)
-                tp = len(elocs[mest][dkeep])  # n_matches
-                true_galaxy_bools = true["galaxy_bools"][b][mtrue][dkeep]
-                tp_gal = true_galaxy_bools.bool().sum()
+                tp = len(elocs[mest][dkeep])
```
jeff-regier, line 79: Is the `if` statement above OK? It seems like if, say, `ntrue == 0`, we could still have some false positives, for example

bliss/metrics.py
```diff
@@ -94,70 +107,14 @@ def compute(self) -> Dict[str, Tensor]:
...
-        """Calculate misclassification accuracy, and confusion matrix."""
-        return {
-            "n_matches": self.total_n_matches,
-            "n_matches_gal_coadd": self.total_coadd_gal_matches,
-            "class_acc": self.total_correct_class / self.total_n_matches,
-            "conf_matrix": self.conf_matrix,
+            "n_matches": self.gal_tp + self.gal_fp + self.gal_tn + self.gal_fn,
+            "n_matches_gal_coadd": self.gal_tp + self.gal_fn,
+            "class_acc": (self.gal_tp + self.gal_tn)
+            / (self.gal_tp + self.gal_fp + self.gal_tn + self.gal_fn),
```
jeff-regier, line 113: Tough to parse. Could you define the part of this calculation, e.g., the denominator immediately before the return statement?

bliss/metrics.py
```diff
@@ -188,8 +145,13 @@ def match_by_locs(true_locs, est_locs, slack=1.0):
     assert true_locs.shape[-1] == est_locs.shape[-1] == 2
     assert isinstance(true_locs, torch.Tensor) and isinstance(est_locs, torch.Tensor)
 
-    locs1 = true_locs.view(-1, 2)
-    locs2 = est_locs.view(-1, 2)
+    # reshape
+    locs1 = true_locs.view(-1, 2)  # 11x2 - coordinates for true light sources
+    locs2 = est_locs.view(-1, 2)  # 7x2 - coordinates for estimated light sources
```
jeff-regier, line 150: Are the 11 and 7 here referring to a test case? If so, better to delete them

bliss/metrics.py
```diff
@@ -188,8 +145,13 @@ def match_by_locs(true_locs, est_locs, slack=1.0):
...
+
+    # remove non-existent estimated/true light sources
+    locs1 = locs1[torch.abs(locs1).sum(dim=1) != 0]
+    locs2 = locs2[torch.abs(locs2).sum(dim=1) != 0]
```
jeff-regier, line 154: Can you use `n_sources` to remove off light sources rather than relying on locs summing to zero? Even if locs summing to zero works consistently, it's sort of hard from someone to understand

bliss/metrics.py
```diff
@@ -40,50 +43,60 @@ def __init__(
...
 
     # pylint: disable=no-member
     def update(self, true: FullCatalog, est: FullCatalog) -> None:  # type: ignore
-        """Update the internal state of the metric including tp, fp, total_true_n_sources, etc."""
+        """Update the internal state of metrics including tp, fp, total_coadd_gal_matches, etc."""
         assert true.batch_size == est.batch_size
 
         count = 0
-        desc = "Detection Metric per batch"
+        desc = "Bliss Metric per batch"
```
jeff-regier, line 72: Can/should we get rid of tqdm on the line before this one? Presumably the loop will be extremely fast -- almost instant

bliss/metrics.py
```diff
@@ -94,70 +107,14 @@ def compute(self) -> Dict[str, Tensor]:
             "recall": recall,
             "f1": f1,
             "avg_distance": self.avg_distance,
-            "n_galaxies_detected": self.tp_gal,
```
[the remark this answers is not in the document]
jeff-regier: It'd be nice to remove "tp" and "fp" from this return statement, so that they aren't logged; it's enough to know precision and recall.

bliss/metrics.py
sawanp813: Will do! It's definitely an unclear naming convention as-is.

bliss/metrics.py
sawanp813: Right. The if-statement was pre-existing but I didn't think of that case - will fix.

bliss/metrics.py
sawanp813: Sure!

bliss/metrics.py
sawanp813: Yes they're referring to a test case, I caught that a little bit ago. Will fix.

bliss/metrics.py
sawanp813: Definitely, sounds good. Think it's also more logical to handle it like you suggested.

bliss/metrics.py
sawanp813: Sure - I wasn't sure if the progress bar was needed for some reason. Sounds good!

bliss/metrics.py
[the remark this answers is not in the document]
sawanp813: Right, I can rename those metrics. And agreed, 'fp' and 'tp' aren't necessary once precision and recall are calculated.

bliss/metrics.py
```diff
@@ -198,215 +160,20 @@ def match_by_locs(true_locs, est_locs, slack=1.0):
     # Penalize all pairs which are greater than slack apart to favor valid matches.
     locs_err = locs_err + (locs_err_l_infty > slack) * locs_err.max()
 
+    # convert light source error matrix to CSR
+    csr_locs_err = csr_matrix(locs_err.detach().cpu())
+
     # find minimal permutation and return matches
-    row_indx, col_indx = sp_optim.linear_sum_assignment(locs_err.detach().cpu())
+    row_indx, col_indx = min_weight_full_bipartite_matching(csr_locs_err)
```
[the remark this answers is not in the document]
sawanp813: Aakash and I were actually just talking about how the 'min_weight_full...' matching has some strange effects on the metrics in comparison to the 'linear_sum_assignment' function matching. We can talk about it more tomorrow morning but just wanted to give you a heads up!

bliss/metrics.py
jeff-regier: Yep, thanks, that's what I had in mind

aakashdp6548:
Hey FYI, the new metrics don't work on gpu. Running `pytest tests/test_end_to_end -k test_train --gpu` fails on this line in BlissMetrics.update with "can't convert cuda:0 device type tensor to numpy":
`conf_matrix = confusion_matrix(tgbool, egbool, labels=[1, 0])`

bliss/metrics.py
```diff
@@ -36,128 +39,80 @@ def __init__(
...
+            detection_tp = len(elocs[mest][dkeep])
+            detection_fp = nest - detection_tp
+            tgbool = tgbool[mtrue][dkeep].reshape(-1)
+            egbool = egbool[mest][dkeep].reshape(-1)
+            assert fp >= 0
+            self.detection_tp += detection_tp
+            self.detection_fp += detection_fp
+            self.avg_distance += avg_distance
+            self.total_true_n_sources += ntrue  # type: ignore
+            conf_matrix = confusion_matrix(tgbool.cpu(), egbool.cpu(), labels=[1, 0])
```
sawanp813, line 88: Confusion_matrix converts the iterable inputs to Numpy arrays, so **_tgbool_** and **_egbool_** have to be allocated on the CPU since Numpy arrays are not supported on the GPU.

jeff-regier (approved):
LGTM!

PR #754, opened 2023-05-10
Fix metrics bug when 0 sources

### LINKED ISSUE #752: Fix metrics behavior when either true_locs or est_locs is empty
`match_by_locs()` throws a runtime error if either true_locs or est_locs are empty, which can happen if the priors are too small. If both are empty we can just continue in `update()`, since there are no TP/FP/TN/FN. If one is empty and the other isn't, we probably need to update the values manually.

```
# Penalize all pairs which are greater than slack apart to favor valid matches.
>       locs_err = locs_err + (locs_err_l_infty > slack) * locs_err.max()
E       RuntimeError: max(): Expected reduction dim to be specified for input.numel() == 0. Specify the reduction dim with the 'dim' argument.
```

aakashdp6548:
Fixes #752. Metrics didn't correctly handle the case when either the number of true or estimated sources in a batch was 0, and would raise an exception. Now we manually check before calling `match_by_locs` and increment false positives / false negatives accordingly. I don't think there's any good way of measuring "average distance" in this case since there are no matches to compute the distance between, so that only gets updated in batches that have both true and estimated sources.

Also added `test_no_sources` in test_metrics that checks this functionality.

### DIFF
--- a/bliss/metrics.py
+++ b/bliss/metrics.py
@@ -72,6 +72,17 @@ class BlissMetrics(Metric):
             ntrue, nest = true.n_sources[b].int().item(), est.n_sources[b].int().item()
             tlocs, elocs = true.plocs[b], est.plocs[b]
             tgbool, egbool = true["galaxy_bools"][b].reshape(-1), est["galaxy_bools"][b].reshape(-1)
+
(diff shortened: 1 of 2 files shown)


jeff-regier (approved):
Makes sense & looks good

PR #755, opened 2023-05-11
Decouple generate config from training

zhixiangteoh:
So that `cfg.cached_simulator` is only needed/used for training, and `cfg.generate` (with `cfg.simulator`) defines how data is generated at generate step.

### DIFF
--- a/bliss/generate.py
+++ b/bliss/generate.py
@@ -17,7 +17,7 @@ FileDatum = TypedDict(
 
 
 def generate(cfg: DictConfig):
-    file_data_capacity = cfg.cached_simulator.file_data_capacity
+    file_data_capacity = cfg.generate.file_data_capacity
     cached_data_path = cfg.cached_simulator.cached_data_path
 
     # largest `batch_size` multiple <= `file_data_capacity`
--- a/bliss/generate.py
+++ b/bliss/generate.py
@@ -17,17 +17,17 @@ FileDatum = TypedDict(
 
 
 def generate(cfg: DictConfig):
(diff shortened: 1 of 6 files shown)


bliss/generate.py
```diff
@@ -17,7 +17,7 @@
 
 
 def generate(cfg: DictConfig):
-    file_data_capacity = cfg.cached_simulator.file_data_capacity
+    file_data_capacity = cfg.generate.file_data_capacity
     cached_data_path = cfg.cached_simulator.cached_data_path
```
aakashdp6548, line 21: I think it would also be good to decouple `cached_data_path`, like defining it under `cfg.generate` and pointing `cfg.cached_simulator.cached_data_path = ${cfg.generate.cached_data_path}` if desired.

jeff-regier: Could we rename `file_data_capacity` to something like `images_per_file`? (Is that what it represents?) Otherwise it's not clear what the units are (file capacity is often measured in number of bytes).

bliss/generate.py
```diff
@@ -17,7 +17,7 @@
 
 
 def generate(cfg: DictConfig):
-    file_data_capacity = cfg.cached_simulator.file_data_capacity
+    file_data_capacity = cfg.generate.file_data_capacity
```
jeff-regier, line 20: Could we rename `file_data_capacity` to something like `images_per_file`? (Is that what it represents?) Otherwise it's not clear what the units are (file capacity is often measured in number of bytes).

jeff-regier (approved):
Please merge once checks are passing and comments are addressed

PR #760, opened 2023-05-12
Make CachedSimulatedDataset mappable Dataset

zhixiangteoh:
Partially fix bug where each epoch runs `num_workers * n_batches + valid_n_batches` iterations when it should be `n_batches + valid_n_batches` iterations. Partial because bug would remain for `SimulatedDataset`.

Associated sanity-check test results ensuring similar validation losses / speed of the two approaches:
| dataset type                         | mappable | mappable | iterable | iterable |
| - | - | - | - | - |
| num workers                          | 0        | 32       | 0 | 32 |
| training loss (after 250 iterations) | 1.82     | 1.82     | 1.92 | 1.9 |
| val loss                             | 1.8513   | 1.8513   | 1.90651 | \- |
| speed (median it/s)                  | 4.98     | 1.4      | 4.92 | 3.98 |

### DIFF
--- a/bliss/generate.py
+++ b/bliss/generate.py
@@ -28,9 +28,6 @@ def generate(cfg: DictConfig):
     # in <= `images_per_file`-image files
     n_files = -(cfg.generate.n_batches * bs // -images_per_file)  # ceil division
 
-    # stores details of the written image files - { filename: string, data }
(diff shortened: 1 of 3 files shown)


jeff-regier (approved):
Looks good!

PR #764, opened 2023-05-15
Multi-process single-threaded data gen

### LINKED ISSUE #761: single-threaded data generation
The rumor is that our multithreaded data generation routine isn’t scaling well in the number of threads. If true, we could get rid of threading (i.e., set `num_workers = 0`) and just launch multiple independent processes to generate data. Each process would write one output file. To launch these processes we could use a bash command line the following:
```
for i in {0 .. 31}; do bliss generate generate.name=train generate.filenum=$i & done
```

zhixiangteoh:
Either 
```
{BLISS_HOME}/generate_data_in_parallel.sh 32 0 {CONFIG_PATH}
```
or 

```
bliss -cp {CONFIG_PATH} 'mode=generate' 'generate.n_processes=32' 'generate.n_workers_per_process=0'
```

Some stats:
32 processes * 6 batches * 64 images/batch = **12,288 images** | **32 processes** | **0 workers/process**
- bash: 12454.65s user 113.11s system 2074% cpu 10:05.89 total
- python multiprocessing: 10617.55s user 74.27s system 934% cpu 19:03.68 total

1 process * 192 batches * 64 images/batch = **12,288 images** | **1 process** | **32 workers/process**
- bash: 7936.11s user 33.56s system 114% cpu 1:55:54.86  total

Fixes #761.

### DIFF
--- a/conf/base_config.yaml
+++ b/conf/base_config.yaml
@@ -56,13 +56,13 @@ simulator:
 
 cached_simulator:
     _target_: bliss.simulator.simulated_dataset.CachedSimulatedDataset
-    n_batches: ${generate.n_batches}
+    train_n_batches: ${generate.n_batches}
     batch_size: ${generate.batch_size}
     num_workers: 0
     cached_data_path: ${generate.cached_data_path}
-    train_file_prefix: ${generate.train_file_prefix}
-    val_split_file_idxs: null
-    test_split_file_idxs: null
+    file_prefix: ${generate.file_prefix}
+    val_split_file_idxs: null # inclusive range
+    test_split_file_idxs: null # inclusive range
 
 encoder:
     _target_: bliss.encoder.Encoder
@@ -109,14 +109,13 @@ encoder:
         ]
 
 generate:
-    splits: [train, valid, test]
+    n_workers_per_process: 0
     n_batches: ${simulator.n_batches}
     batch_size: ${simulator.prior.batch_size}
-    valid_n_batches: ${simulator.valid_n_batches}
-    test_n_batches: ${simulator.valid_n_batches}
     max_images_per_file: 5000
     cached_data_path: ${paths.data}/cached_dataset
-    train_file_prefix: dataset
+    file_prefix: dataset
+    files_start_idx: 0
 
 training:
     name: null
--- a/conf/base_config.yaml
+++ b/conf/base_config.yaml
@@ -60,6 +60,9 @@ cached_simulator:
     batch_size: ${generate.batch_size}
     num_workers: 0
     cached_data_path: ${generate.cached_data_path}
+    train_file_prefix: ${generate.train_file_prefix}
+    val_split_file_idxs: null
+    test_split_file_idxs: null
 
 encoder:
     _target_: bliss.encoder.Encoder
@@ -113,6 +116,7 @@ generate:
     test_n_batches: ${simulator.valid_n_batches}
     max_images_per_file: 5000
     cached_data_path: ${paths.data}/cached_dataset
+    train_file_prefix: dataset
 
 training:
     name: null
--- a/bliss/generate.py
+++ b/bliss/generate.py
@@ -52,26 +52,10 @@ def generate(cfg: DictConfig):
                 images_per_file // bs, simulated_dataset, "Simulating images in batches for file"
             )
             file_data = itemize_data(batch_data)
-            with open(f"{cached_data_path}/dataset_{file_idx}.pt", "wb") as f:
+            with open(
+                f"{cached_data_path}/{cfg.generate.train_file_prefix}_{file_idx}.pt", "wb"
+            ) as f:
                 torch.save(file_data, f)
-
-    if "valid" in cfg.generate.splits:
-        valid = generate_data(
-            cfg.generate.valid_n_batches,
-            simulated_dataset,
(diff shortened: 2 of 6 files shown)


zhixiangteoh:
Need input on

conf/base_config.yaml
```diff
@@ -107,12 +107,15 @@ encoder:
...
     splits: [train, valid, test]
+    n_processes: 1
+    n_workers_per_process: 0
     n_batches: ${simulator.n_batches}
     batch_size: ${simulator.prior.batch_size}
     valid_n_batches: ${simulator.valid_n_batches}
     test_n_batches: ${simulator.valid_n_batches}
     max_images_per_file: 5000
     cached_data_path: ${paths.data}/cached_dataset
+    train_file_prefix: dataset
```
aakashdp6548, line 118: You'll need to change the loading in CachedSimulatedDataset to use this too, since it's hardcoded as "dataset_" right now

aakashdp6548:
Based on the timing stats, the bash script sounds like a better option to me. The Python method isn't just slower, but it also seems more inefficient: 10617 / 60 / 32 + 1 = 6.5 is way less than the total time of 19 minutes, so that suggests that it's spending most of the time idle/sleeping (for comparison, with the bash script it's 12454 / 60 / 32 + 2 = 8.48 compared to 10 total).

Since this script is just a utility to run the generation multiple times, I think it's fine to not have it built in to the bliss program itself, and maybe just add a note in the readme or in the docs (when we actually update them lol) indicating how to use the script. FWIW in theory we shouldn't need to run it too many times anyway once we have the background/psf/band changes worked out.

jeff-regier (approved):
I appreciate you trying out both possibilities. bash seems better to me as it is faster in wall-clock time. I'm not really sure where the script should go though. Perhaps we should have a new top-level `scripts` directory?

conf/base_config.yaml
zhixiangteoh: Done

aakashdp6548: Also need to update the filenames for the validation and test datasets: 
https://github.com/prob-ml/bliss/blob/201ca1f7594c140490b19b863e6cd85e7f5b4788/bliss/simulator/simulated_dataset.py#L132

zhixiangteoh: Revamped design / use case of data generation/training. ptal at new commit.

zhixiangteoh:
@jeff-regier @aakashdp6548 Data generation is now split-blind (i.e., does not care about train/val/test splits). At training time, `cached_simulator` uses `cfg.cached_simulator.(val/test)_split_file_idxs` to determine which images to use for train, val, and test splits.

jeff-regier (approved):
Looks really good

aakashdp6548 (approved):
I'm not a huge fan of all the manual computation involved with specifying data, especially with this new change. First we need to do `generate.n_batches` * `generate.batch_size` to get the total number of images. Then we divide by `generate.max_images_per_file` and ceil to get the total number of files to specify the indices for the val/test sets. If we want to know the percentage of the data used for each split, that might just be # of indices / number of files, but depending on batch_size we might be using less training data than was generated (but same amount of validation/test) so that would need more manual calculations.

I feel like it would be a lot cleaner to abstract away all of the calculations and just let the user specify how much data they want, similar to what Huggingface does - when loading a dataset (even from disk), you specify what splits you want and what percentage of that split: https://huggingface.co/docs/datasets/v1.4.1/splits.html I'm not saying we need that same degree of flexibility, but something a little bit more streamlined would be easier to use than what we have right now.

I'll go ahead and approve this since we've already talked about this, but I think at some point we definitely should clean it up a little.

zhixiangteoh:
> I'm not a huge fan of all the manual computation involved with specifying data, especially with this new change. First we need to do `generate.n_batches` * `generate.batch_size` to get the total number of images. Then we divide by `generate.max_images_per_file` and ceil to get the total number of files to specify the indices for the val/test sets. If we want to know the percentage of the data used for each split, that might just be # of indices / number of files, but depending on batch_size we might be using less training data than was generated (but same amount of validation/test) so that would need more manual calculations.

As discussed before, there are conflicting goals right now between maintaining parity with `SimulatedDataset`, and streamlining the generation -> training process. This explains the existence of `cfg.cached_simulator.train_n_batches`. Certainly once we have a much better idea of how/when we want to generate / load data, we can streamline this process more, and in so doing sacrifice more parity with `SimulatedDataset`.

> I feel like it would be a lot cleaner to abstract away all of the calculations and just let the user specify how much data they want, similar to what Huggingface does - when loading a dataset (even from disk), you specify what splits you want and what percentage of that split: https://huggingface.co/docs/datasets/v1.4.1/splits.html I'm not saying we need that same degree of flexibility, but something a little bit more streamlined would be easier to use than what we have right now.

IMO, this change is actually a step in the right direction towards this vision—it is just that now, instead of specifying split percentage, we manually specify the file indices used for val/test splits, in `cfg.cached_simulator.(val/test)_split_file_idxs`. Right now this gives more flexibility in terms of which files exactly you want to use for val/test splits, but it is true that it requires the user to inspect the `hparams.yaml` to determine how many batches each file is, and also it only allows train/val/test splits at the file-level, so it might not be as fine-grained as we might potentially want it to be.

We'll merge this and continue experimenting / streamlining this process.

PR #767, opened 2023-05-16
Add early stopping

aakashdp6548:
Add option to stop training early if validation loss hasn't decreased for a set number of epochs. This is configurable using the `enable_early_stopping` and `patience` flags in `config.trainer`.

I removed the `min_epochs` flag from trainer because that overrides early stopping. I also suppressed a pylint warning for too-many-statements in `train` - we should maybe consider factoring some of the stuff that isn't directly related to training, like loading best weights, into a separate function.

### DIFF
--- a/bliss/train.py
+++ b/bliss/train.py
@@ -2,19 +2,20 @@ import datetime as dt
 import json
 from pathlib import Path
 from time import time_ns
-from typing import Any, Dict, Optional
+from typing import Any, Dict
 
 import pytorch_lightning as pl
 import torch
 from hydra.utils import instantiate
 from omegaconf import DictConfig, OmegaConf
 from pytorch_lightning.callbacks import ModelCheckpoint
+from pytorch_lightning.callbacks.early_stopping import EarlyStopping
 from pytorch_lightning.loggers import TensorBoardLogger
 from pytorch_lightning.profiler import AdvancedProfiler
 from pytorch_lightning.utilities import rank_zero_only
 
 
-def train(cfg: DictConfig):  # pylint: disable=too-many-branches
+def train(cfg: DictConfig):  # pylint: disable=too-many-branches, too-many-statements
     # setup paths and seed
     paths = OmegaConf.to_container(cfg.paths, resolve=True)
     assert isinstance(paths, dict)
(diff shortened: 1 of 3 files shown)


bliss/train.py
```diff
@@ -44,10 +45,15 @@ def train(cfg: DictConfig):  # pylint: disable=too-many-branches
...
+    checkpoint_callback = setup_checkpoint_callback(cfg)
+    early_stopping = setup_early_stopping(cfg)
     profiler = setup_profiler(cfg)
 
-    callbacks = [] if checkpoint_callback is None else [checkpoint_callback]
+    callbacks = []
+    if checkpoint_callback:
+        callbacks.append(checkpoint_callback)
+    if early_stopping:
+        callbacks.append(early_stopping)
```
zhixiangteoh, line 56: Perhaps add a test for early stopping? Both to silence this warning, and also as good practice.

bliss/train.py
```diff
@@ -111,11 +118,18 @@ def setup_callbacks(cfg) -> Optional[ModelCheckpoint]:
...
 
 
+def setup_early_stopping(cfg):
+    early_stopping = None
+    if cfg.training.enable_early_stopping:
+        early_stopping = EarlyStopping(
+            monitor="val/loss", mode="min", patience=cfg.training.patience
+        )
+    return early_stopping
+
```
zhixiangteoh, line 131: Ditto.

conf/base_config.yaml
```diff
@@ -129,7 +131,6 @@ training:
         check_val_every_n_epoch: 10
         log_every_n_steps: 10  # corresponds to n_batches
         max_epochs: ${training.n_epochs}
-        min_epochs: ${training.n_epochs}
```
zhixiangteoh, line 132: Is there a way to keep `min_epochs` while having the option to enable early stopping? Ideally we would want to still be able to train for exactly >= `min_epochs`?

zhixiangteoh (changes requested):
Thanks! If it's straightforward to add a test for early stopping, that'd be great

conf/base_config.yaml
aakashdp6548: We can keep it, but we'd have to make it a separate config parameter - with it just using training.n_epochs, it would always train for exactly n_epochs, even with early stopping enabled.

bliss/train.py
aakashdp6548: Imo it would be tricky to actually test early stopping itself, since it would need to run for at least a couple of epochs and would cause tests to take a lot longer. Instead, I just toggled `enable_early_stopping` in `testing_config` so that the callback initialization gets hit in `test_train`.

bliss/train.py
aakashdp6548: Fixed by enabling early stopping in tests.

jeff-regier (approved):
LGTM!

zhixiangteoh (approved):
Looks good

PR #769, opened 2023-05-17
Refactor decoder to simplify PSF construction

aakashdp6548:
Cleaned up the code in `decoder.py` to simplify the functions used to generate the PSF, and add more documentation about what each function is doing.

Main changes:

- combining `forward_psf_from_params` and some of the processing in the constructor into a single function, `_get_psf`
- making `_get_mgrid` no longer static, and removing the scaling factor at the end (since it was un-scaled immediately in the constructor, canceling it out) 
- changing self.params from a Parameter to a buffer

Comparing the PSF generated by the current changes to the one generated by the old version, there are minor differences (on the order of 10^-12), but I think those are just due to imprecision with floating point operations. 

Please take a look and let me know if these changes seem reasonable - I'm only planning on merging into the psf branch for now, so I'll do additional testing there too before merging to master.

```
old_psf = <psf generated from old version of decoder>
new_decoder = instantiate(cfg.simulator.decoder, psf_params_file="/data/sdss/94/1/12/psField-000094-1-0012.fits")
new_psf = new_decoder.psf_galsim._original._image._array

old_psf == new_psf  # False - small differences
np.allclose(old_psf, new_psf)  # True
```

### DIFF
--- a/bliss/simulator/decoder.py
+++ b/bliss/simulator/decoder.py
@@ -22,43 +22,31 @@ class ImageDecoder(nn.Module):
     ) -> None:
         super().__init__()
         assert n_bands == 1, "Only 1 band is supported"
+        assert Path(psf_params_file).suffix == ".fits"
+        assert len(sdss_bands) == n_bands
 
         self.n_bands = n_bands
         self.pixel_scale = pixel_scale
-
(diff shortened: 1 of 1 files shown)


jeff-regier (approved):
Looks much better

PR #771, opened 2023-05-22
Decoder enhancements

sawanp813:


bliss/encoder.py
sawanp813, line 1: These changes to the encoder ensure that execution still occurs in the single-band case. The encoder doesn't yet support five-dimensional inputs, for example, which I'll be exploring more this week. For now, multiple-band encoding will execute when the input is four-dimensional as it is here.

bliss/simulator/decoder.py
```diff
@@ -14,55 +14,49 @@
...
-        assert n_bands == 1, "Only 1 band is supported"
 
-        self.n_bands = n_bands
+        self.n_bands = len(sdss_fields["bands"])
         self.pixel_scale = pixel_scale
+        self.psf_slen = psf_slen
+        self.psf_galsim = {}  # Dictionary indexed by (run, camcol, field) tuple
 
-        self.params = None  # psf params from fits file
+        sdss_dir = sdss_fields["dir"]
```
sawanp813, line 28: Here, we retrieve the relevant information for computing the relative flux values for sampled light sources from arbitrary SDSS catalog rcf's.

bliss/simulator/decoder.py
```diff
@@ -14,55 +14,49 @@
...
-        assert len(sdss_bands) == n_bands
-        psf_params = self._get_fit_file_psf_params(psf_params_file, sdss_bands)
-        self.params = nn.Parameter(psf_params.clone(), requires_grad=True)
+        for field_params in sdss_fields["field_list"]:
+            run = field_params["run"]
+            camcol = field_params["camcol"]
+            fields = field_params["fields"]
 
-        self.psf_slen = psf_slen
+            for field in fields:
```
sawanp813, line 36: L36-43 results in the construction of the `self.psf_galsim` dictionary, which is indexed by `(<run>, <camcol>, <field>)` and returns the PSFs for that particular field. Particularly, only the requested PSFs are generated based on the bands specified in `sdss_bands`.

bliss/simulator/prior.py
```diff
@@ -86,18 +99,73 @@ def __init__(
...
+            return
+        for field_params in sdss_fields["field_list"]:
+            run = field_params["run"]
+            camcol = field_params["camcol"]
+            fields = field_params["fields"]
+            for field in fields:
+                sdss_path = Path(self.sdss)
+                camcol_dir = sdss_path / str(run) / str(camcol) / str(field)
+                po_path = camcol_dir / f"photoObj-{run:06d}-{camcol:d}-{field:04d}.fits"
+                po_fits = fits.getdata(po_path)
```
aakashdp6548, line 113: A lot of the codecov warnings can probably be fixed by adding a test that uses SDSS backgrounds drawn from multiple fields. I can work on writing one this afternoon/evening.

conf/base_config.yaml
```diff
@@ -84,7 +91,20 @@ encoder:
...
+            # [-1, 3, Conv, [64, 1, 1]],
+            # [-1, 1, Conv, [128, 3, 2]],
+            # [-1, 1, Conv, [128, 3, 1]],
+            # [-1, 1, Conv, [256, 3, 2]],
+            # [-1, 6, C3, [256]],
+            # [-1, 1, Conv, [512, 3, 2]],
+            # [-1, 9, C3, [512]],
+            # [-1, 1, Conv, [1024, 3, 2]],
+            # [-1, 3, C3, [1024]],
+            # [-1, 1, SPPF, [1024, 5]],
```
aakashdp6548, line 106: @sawanp813  Can we remove these commented lines in this PR?

bliss/simulator/prior.py
```diff
@@ -95,18 +99,72 @@ def __init__(
         self.galaxy_a_scale = galaxy_a_scale
         self.galaxy_a_bd_ratio = galaxy_a_bd_ratio
 
-    def sample_prior(self) -> TileCatalog:
+        # load all star, galaxy ratios required for sampling
+        if self.n_bands == 1:
+            return
+        for field_params in sdss_fields["field_list"]:
```
sawanp813, line 105: (L105 - 137) Constructs `self.rcf_stars_rest` and `self.rcf_gals_rest`, which contain the relative flux values across all bands for a given RCF depending on whether the source is a star or galaxy. This change ensures that all files are read in the constructor up-front as opposed to having a random file read whenever a batch is generated. Additionally, it fixes the issue with the rejection-sampling scheme from earlier. This avoids the issues with variable runtime.

bliss/simulator/prior.py
```diff
@@ -95,18 +99,72 @@ def __init__(
...
             `(batch_size, self.n_tiles_h, self.n_tiles_w)`.
             The remaining dimensions are variable-specific.
         """
         locs = self._sample_locs()
-        galaxy_params = self._sample_galaxy_prior()
-        star_fluxes = self._sample_star_fluxes()
+        select_gal_rcfs = []
+        select_star_rcfs = []
+
+        for rcf in batch_rcfs:
```
sawanp813: (L155 - 167) The relative flux ratios corresponding to the selected RCFs for this particular batch are collected and apssed to `self._sample_galaxy_prior` and `self._sample_star_fluxes`. Each image that is generated samples flux ratios from one particular FITs file.

bliss/simulator/prior.py
```diff
@@ -148,53 +206,33 @@ def _sample_n_galaxies_and_stars(self, n_sources):
...
-        thing_id = column_to_tensor(po_fits, "thing_id").numpy()
-
-        # TODO: Rewrite so star and galaxy sources are separated from SDSS file
-        # to avoid variable runtime in while loop
-        # pick random light source from SDSS image
-        obj = np.random.randint(0, 999)
-        while objc_type[obj] != obj_id or thing_id[obj] == -1 or torch.any(fluxes[obj] < 0):
-            obj = np.random.randint(0, 999)
-
+    def multiflux(self, rcf_ratios, flux_base) -> Tensor:
```
sawanp813, line 224: `multiflux` is changed to account for the upstream changes. Now, there is no checking to see whether a sampled light source from a FITS file is a star or a galaxy as this is known beforehand when `multiflux` is called. `multiflux` computes and returns the flux values for all remaining bands scaled according to the draw from the Pareto distribution for the first band specified in the config.

bliss/simulator/prior.py
```diff
@@ -216,7 +254,8 @@ def _sample_galaxy_prior(self):
             self.galaxy_alpha, self.galaxy_flux_min, self.galaxy_flux_max, latent_dims
         )
 
-        total_flux[:, :, :, :, 1:] = self.multiflux(3, total_flux[:, :, :, :, 0])
+        if self.n_bands > 1:
```
sawanp813, line 258: Condition checks if the number of bands is greater than 1 because if only single-band inference is requested, there is no need for any sampling from the SDSS catalog.

bliss/simulator/prior.py
```diff
@@ -148,53 +206,33 @@ def _sample_n_galaxies_and_stars(self, n_sources):
...
         )
-        # draw value for g-band from pareto for each light source
+        # sample value for first band from pareto for each light source
         fluxes[:, :, :, :, 0] = self._draw_truncated_pareto(
             self.star_flux_alpha, self.star_flux_min, self.star_flux_max, latent_dims
         )
-        # Compute r-band via g-band value and ratio from SDSS catalog
-        fluxes[:, :, :, :, 1:] = self.multiflux(6, fluxes[:, :, :, :, 0])
+        # Compute remaining band fluxes via ratio from SDSS catalog and sampled values
+        if self.n_bands > 1:
```
sawanp813, line 220: Condition checks if the number of bands is greater than 1 because if only single-band inference is requested, there is no need for any sampling from the SDSS catalog. Also, we pass `multiflux` the `star_ratios` as computed above which refers to the particular RCFs which will be sampled to render the stars for each image in the batch.

bliss/simulator/prior.py
jeff-regier: Yep, it'd be good add to a test to keep our coverage from decreasing

bliss/encoder.py
```diff
@@ -93,6 +93,13 @@ def dist_param_groups(self):
         }
 
     def encode_batch(self, batch):
+        img_dims = list(batch["images"].size())
+        images_with_background = torch.zeros(
+            img_dims[0],
+            img_dims[1] * 2,
+            img_dims[2],
+            img_dims[3],
+        )
```
jeff-regier, line 102: Is this debugging code that got left in? (Why are we setting `images_with_background` here if we're just going to redefine it on right away?)

bliss/simulator/background.py
```diff
@@ -1,48 +1,64 @@
...
+    def __init__(self, background: Tuple[float, ...], **_kwargs):
         super().__init__()
-        bg: Tensor = torch.tensor(background)
-        bg = rearrange(bg, "c -> 1 c 1 1")
+        bg: Tensor = torch.tensor([background])
+        bg = rearrange(bg[0], "c -> 1 c 1 1")
         self.register_buffer("background", bg, persistent=False)
 
-    def sample(self, shape) -> Tensor:
+    def sample(self, shape, **_kwargs) -> Tensor:
```
jeff-regier, line 21: Why are we adding `_kwargs` here and above? It doesn't look like they are used

bliss/encoder.py
sawanp813: Yes it was - I added the `torch.cat` line afterwards to fix the issue. I'll remove the first definition of `images_with_background`.

bliss/simulator/background.py
```diff
@@ -1,48 +1,64 @@
...
+
+        # Add all backgrounds from specified run/col/field to list
+        backgrounds = []
+        for param_obj in field_list:
+            params: Dict = OmegaConf.to_container(param_obj)  # type: ignore
+            sdss = SloanDigitalSkySurvey(sdss_dir=sdss_dir, bands=bands, **params)
+            backgrounds.extend([field["background"] for field in sdss])  # type: ignore
+
+        background = torch.from_numpy(np.stack(backgrounds, axis=0))
+        if len(background.size()) == 3:  # convert to 4d tensor if only one background
```
jeff-regier, line 42: If there's only one background, can we require that background is already a 4d tensor with a singleton first dimension? (It's nice to avoid conditional expressions when possible.)

jeff-regier: Oh I see, background isn't provided as an argument to this function. Still, is there a way to avoid treating 1 background as a special case?

aakashdp6548: That's what the next line is doing! :) Before there was only one background so this was always executed, now since we might have multiple backgrounds already we need to check first.
https://github.com/prob-ml/bliss/blob/95e3087cf69e40da628e33060245e09b851d1d5e/bliss/simulator/background.py#L42-L43

bliss/simulator/background.py
jeff-regier: Got it

bliss/simulator/background.py
jeff-regier: I was hoping to avoid even this one-line special case, but perhaps that's not possible (?) Why does the dimension of background vary here?

jeff-regier: Is this if condition ever true now? Even if you call `np.stack` with a singleton list, it still adds a dimension:
```
>>> two_d_array = np.ones((4,5))
>>> np.stack([two_d_array,], axis=0).shape
(1, 4, 5)
```

bliss/simulator/background.py
```diff
@@ -1,48 +1,64 @@
...
+
+        # select region to sample from (same for all images in batch for simplicity)
         h_diff, w_diff = self.height - hlen, self.width - wlen
         h = 0 if h_diff == 0 else np.random.randint(h_diff)
         w = 0 if w_diff == 0 else np.random.randint(w_diff)
-        bg = self.background[:, :, h : (h + hlen), w : (w + wlen)]
-        return bg.expand(batch_size, -1, -1, -1)
+
+        # sample region from random background for each image in batch
+        if rcf_indices is not None:
```
jeff-regier, line 59: Any way to get rid of this conditional statement? i.e., do we definitely need both cases? Code with extra branching is a bit harder to test and for other developers to understand

bliss/simulator/decoder.py
```diff
@@ -14,55 +14,49 @@
...
-        psf_params_file: str,
         psf_slen: int,
-        sdss_bands: Tuple[int, ...],
+        sdss_fields: DictConfig,
     ) -> None:
         super().__init__()
-        assert n_bands == 1, "Only 1 band is supported"
 
-        self.n_bands = n_bands
+        self.n_bands = len(sdss_fields["bands"])
```
jeff-regier, line 23: Great. Yeah, there was no reason before to require a caller to specify `n_bands` when we already had list of the bands.

bliss/simulator/decoder.py
```diff
@@ -14,55 +14,49 @@
...
+        for field_params in sdss_fields["field_list"]:
+            run = field_params["run"]
+            camcol = field_params["camcol"]
+            fields = field_params["fields"]
 
-        self.psf_slen = psf_slen
+            for field in fields:
+                # load raw params from file
+                field_dir = f"{sdss_dir}/{run}/{camcol}/{field}"
+                filename = f"{field_dir}/psField-{run:06}-{camcol}-{field:04}.fits"
```
jeff-regier, line 39: This is all quite specific to SDSS, even though it's in `decoder.py`, which isn't supposed to be specific to any one survey. But I see this as an issue to predates this PR, so perhaps it's fine for now. At some point though it'd be good to introduce more abstraction so we can move any SDSS-specific code to `sdss.py`.

bliss/simulator/prior.py
```diff
@@ -139,11 +207,35 @@ def _sample_n_galaxies_and_stars(self, n_sources):
...
+            fluxes[:, :, :, :, 1:] = self.multiflux(star_ratios, fluxes[:, :, :, :, 0])
+        return fluxes
+
+    def multiflux(self, rcf_ratios, flux_base) -> Tensor:
+        """Generate flux values for remaining bands based on draw."""
+        flux_rest = torch.zeros(
+            self.batch_size, self.n_tiles_h, self.n_tiles_w, self.max_sources, self.n_bands - 1
+        )
```
jeff-regier, line 228: Why is so much of this code not run by our existing tests? Could we change the code so that this code does get exercised by our existing tests? i.e., is possible, I'd like to treat single-band generation as special case of multi-band generation that doesn't require different control flow.

jeff-regier: It's harder for bugs to hide is there aren't two ways to do it: one for single-band and one for multi-band. I'd like the same multi-band generation code to work for any number of bands: such as 5, 3, or even 1.

conf/base_config.yaml
```diff
@@ -120,7 +140,7 @@ generate:
 training:
     name: null
     version: null
-    n_epochs: 50
+    n_epochs: 30
```
jeff-regier, line 143: Probably better to exclude changes to `base_config.yaml` from this PR unless they are required for multiband decoding

bliss/simulator/simulated_dataset.py
```diff
@@ -52,17 +79,45 @@ def _apply_noise(images_mean):
...
+            A dictionary of the simulated TileCatalog, and (batch_size, bands, height, width)
+            tensors for images and background.
+        """
+        rcfs, rcf_indices = (
+            self._get_random_rcf(self.image_prior.batch_size) if not independent else None
+        )
         with torch.no_grad():
-            tile_catalog = self.image_prior.sample_prior()
-            images, background = self.simulate_image(tile_catalog)
+            tile_catalog = self.image_prior.sample_prior(rcfs)  # TODO: add rcf
```
jeff-regier, line 119: I don't think I understand the TODO comment. Maybe open a github issue instead?

jeff-regier (changes requested):
I'm excited it works, but a bit concerned that multi-band and single-band generation aren't using all of the same code. We should avoid extra conditional statements / branching whenever possible. Ultimately, I don't see use ever wanting to generate single-banded data except perhaps for unit testing.

bliss/simulator/decoder.py
aakashdp6548: Sawan and I were discussing this on Friday, and we were thinking the best way to make this more general (and cleaner!) would probably be better to give Simulator a catalog-style object that contains all of the possible backgrounds/psfs/ratios to use, instead of it having to load them itself. But that would require some more general changes (and probably some new data structures) so I think it would be better to save it for a different PR.

bliss/simulator/simulated_dataset.py
aakashdp6548: Forgot to remove this, it was just a reminder to pass rcf to sample_prior. Will be fixed in the next commit.

bliss/simulator/decoder.py
jeff-regier: Sounds good

bliss/simulator/prior.py
sawanp813: I think in the config located on the remote repo the number of bands is specified as just 1 which is probably why CodeCov didn't check any of the lines where `n_bands` was greater than 1. When testing locally prior to committing, the tests hit the unchecked code when I was debugging, so I might change the base and testing configs to specify multiple bands. I definitely agree that the focus should be multi-band generation and that single-band is a special case. 

I thought to add the conditional statements because we don't have a reason to do any sampling from the SDSS catalogs for ratios if we are only using a single band. I'm not sure how we could exactly combine the multi-band and single-band cases for that reason.

bliss/simulator/background.py
aakashdp6548: Huh you're right, it already is 1 x c x h x w for a single background already, good catch! We can just remove this check then.
image

bliss/simulator/prior.py
jeff-regier: Even in the single band case, if that single band isn't the r band, we'll still need to use this sampling code. I'm not sure if this, but perhaps we should sample all five bands each time, and then only use the fluxes for subset of the bands that are specified in the config file.

sawanp813: Oh. So is the Pareto distribution we sample from only parameterized to reflect the r-band flux value?

bliss/simulator/background.py
aakashdp6548: Since we were unsure last week of whether we should choose backgrounds and psfs independently or specify which R/C/F to use for both, I left it as an option in case we change our mind later while still finalizing the model. Right now it's set up to always use the RCF - should I remove the option completely, or would it be better to just document the function better?

bliss/simulator/prior.py
jeff-regier: Yes, I think so

bliss/simulator/background.py
jeff-regier: Probably better if pick one way

bliss/simulator/prior.py
aakashdp6548: @sawanp813 Could we just always load the ratios for all five bands in the constructor and return all five bands in `multiflux`, and then just slice by the bands we want in `sample_prior`? So `sample_star_fluxes` and `sample_galaxy_fluxes` would always return a tensor with last dimension 5 for all 5 bands, and then in sample_prior we can do `self._sample_galaxy_prior(select_gal_rcfs)[:, :, :, :, self.bands]`
https://github.com/prob-ml/bliss/blob/95e3087cf69e40da628e33060245e09b851d1d5e/bliss/simulator/prior.py#L167-L169

It would be a bit inefficient for the one-band case, but it's only during generation and probably not that likely to be used.

bliss/simulator/background.py
aakashdp6548: Got it, I'll make it always require the RCF then. Maybe when we restructure it to take catalog-agnostic input we can add back the option of sampling independently

bliss/simulator/prior.py
sawanp813: Right, that makes sense if the Pareto distribution only reflects the r-band. We can definitely make that change.

sawanp813: It would be inefficient for the one-band case but it doesn't seem like we'll be using single-band inference going forward anyways. I feel like that would be the best way to handle it.

bliss/simulator/prior.py
```diff
@@ -209,33 +207,38 @@ def _sample_n_galaxies_and_stars(self, n_sources):
...
+        # select specified bands
+        select_flux = torch.zeros(
+            self.batch_size, self.n_tiles_h, self.n_tiles_w, self.max_sources, self.n_bands
         )
-        # Compute remaining band fluxes via ratio from SDSS catalog and sampled values
-        if self.n_bands > 1:
-            fluxes[:, :, :, :, 1:] = self.multiflux(star_ratios, fluxes[:, :, :, :, 0])
-        return fluxes
+        for i, bnd in enumerate(self.bands):
+            select_flux[:, :, :, :, i] = total_flux[:, :, :, :, bnd]
```
aakashdp6548: @sawanp813  instead of a for loop, can we just index by self.bands? Then instead of these lines, all we need is `select_flux = total_flux[..., self.bands]` (we may need to convert self.bands to a numpy array first)

sawanp813: Yeah the for-loop wasn't ideal, and I didn't know that was possible - will do!

bliss/simulator/prior.py
aakashdp6548: Added tests for initializing simulator with multiple fields for background/psf, and with multiple bands

bliss/simulator/background.py
aakashdp6548: Fixed

bliss/simulator/decoder.py
```diff
@@ -123,30 +162,53 @@ def render_galaxy(self, galaxy_params: Tensor):
...
+                gs_img = galsim.Image(array=images[b, bnd], scale=self.pixel_scale)
+                for s in range(n_sources):
+                    if full_cat["galaxy_bools"][b][s] == 1:
+                        gal_with_band = torch.cat(  # noqa: WPS220,WPS317
+                            (
+                                torch.unsqueeze(full_cat["galaxy_params"][b][s][bnd], 0),
+                                full_cat["galaxy_params"][b][s][self.n_bands :],
+                            ),
+                            0,
+                        )
```
jeff-regier, line 195: Would you write the lines above a bit more concisely, i.e. using some temp variables? The way black autoformats some things is quite hard to read.

bliss/simulator/prior.py
```diff
@@ -86,18 +99,35 @@ def __init__(
...
-        galaxy_params = self._sample_galaxy_prior()
-        star_fluxes = self._sample_star_fluxes()
+        select_gal_rcfs = []
+        select_star_rcfs = []
+
+        for rcf in batch_rcfs:
+            rcf = tuple(rcf)
+            if rcf in self.rcf_gals_rest:
+                select_gal_rcfs.append(self.rcf_gals_rest[rcf])  # noqa: WPS529
+            else:  # noqa: WPS529
```
jeff-regier, line 122: Should we disable WPS529 project wide?

bliss/simulator/prior.py
```diff
@@ -139,27 +169,96 @@ def _sample_n_galaxies_and_stars(self, n_sources):
...
+                range(self.max_sources),
+                range(self.max_bands),
+            ):
+                # sample a light source
+                flux_sample = obj_ratios[np.random.randint(0, len(obj_ratios) - 1)]
+                total_flux[b, h, w, s, bnd] = flux_sample[bnd] * flux_base[b, h, w, s]
+
+        return total_flux
+
+    def _sample_ratios(self, sdss_fields):
```
jeff-regier, line 202: Is this function sampling something, or just loading a bunch of data? sampling = random. But it looks like this function is just loading. If so, it'd be good to name it something like `_load_color_distribution`

bliss/simulator/prior.py
sawanp813: Aakash and I think that it would make sense to disable it for now since .get() doesn't exactly fit in with the flow of the code in this section. We are also probably going to have to rewrite the majority of this section (and others) anyways when we generalize the decoder to support several other catalogs. I also think it's more readable in this format, personally.

jeff-regier (approved):
Looks good, please merge when ready

PR #774, opened 2023-05-25
Xl/724 improve predict

XinyueLi1012:
allow predict function draw plots


jeff-regier (changes requested):
Please make sure the various checks are passing, and remove the "draft" status of the PR once you're ready for review/merging.

bliss/predict.py
```diff
@@ -11,7 +17,101 @@ def prepare_image(x, device):
     return x[:, :, :height, :width]
 
 
-def predict(cfg, image, background):
+def crop_plocs(cfg, h, plocs):
+    tiles_to_crop = cfg.encoder.tiles_to_crop
+    tile_slen = cfg.encoder.tile_slen
+    if not cfg.predict.ifSimulated:
```
jeff-regier, line 23: Please use underscores in variables names rather than camel case for consistency

bliss/predict.py
```diff
@@ -11,7 +17,101 @@ def prepare_image(x, device):
...
+        crop = cfg.predict.crop
+        minb = crop[0] + tiles_to_crop * tile_slen
+        maxb = crop[1] - tiles_to_crop * tile_slen
+
+    else:
```
jeff-regier, line 28: Is there anyway to avoid this if-else statement, so that images are cropped in the same way regardless of where they are from? (I'm not sure there is, but it's good to avoid branches whenever possible: they're harder to test and harder to understand.)

jeff-regier: For example, if the simulated data were a little more like the real data, could we avoid having this branch?

bliss/predict.py
```diff
@@ -23,19 +123,45 @@ def predict(cfg, image, background):
         pred = encoder.encode_batch(batch)
         est_cat = encoder.variational_mode(pred)
 
-    print("{} light sources detected".format(est_cat.n_sources.item()))
+    if (cfg.predict.plot.ifPlot) & (true_plocs is not None):
```
jeff-regier, line 126: Would `and` have worked here rather than `&`? If so, I'd prefer `and` because its more common

bliss/predict.py
```diff
@@ -23,19 +123,45 @@ def predict(cfg, image, background):
...
+        ptc = ttc * ts
+        cropped_image = image[0, 0, ptc:-ptc, ptc:-ptc]
+        cropped_background = background[0, 0, ptc:-ptc, ptc:-ptc]
+        plot_predict(cfg, cropped_image, cropped_background, true_plocs, est_cat)
 
     return est_cat
 
 
-def predict_sdss(cfg):
+def decal_plocs_from_sdss(cfg):
```
jeff-regier, line 137: Would you add docstring to these functions explaining what they do? This one is a bit mysterious

conf/base_config.yaml
```diff
@@ -163,4 +163,10 @@ predict:
     encoder: ${encoder}
     weight_save_path: ${paths.pretrained_models}/sdss.pt
     device: "cuda:0"
-    crop: [0, -1]
+    crop: [160, 800]
+    plot:
+        ifPlot: true
```
jeff-regier, line 168: here too, please avoid CamelCase

jeff-regier: (for consistency, nothing inherently wrong with it)

conf/base_config.yaml
```diff
@@ -163,4 +163,14 @@ predict:
...
     weight_save_path: ${paths.pretrained_models}/sdss.pt
     device: "cuda:0"
-    crop: [0, -1]
+    crop:
+        if_crop: true
+        left_upper_corner: [160, 160]
+        width: 640
+        height: 640
+    plot:
+        if_plot: true
```
jeff-regier, line 172: Would it make sense to rename `if_plot` to `show_plot`? The variable name `if_plot` doesn't seem self-explanatory, but `show_plot` makes sense to me right away (if that's indeed what this boolean variable controls)

conf/base_config.yaml
```diff
@@ -163,4 +163,14 @@ predict:
...
+        width: 1000
+        height: 1000
+        out_file_name: predict.html
+    if_simulated: false
```
jeff-regier, line 176: Same here, it's not so self-explanatory what `if_simulated` means. Would it make senes to call it `is_simulated`?

conf/base_config.yaml
```diff
@@ -31,7 +31,7 @@ simulator:
         field_list:  # list of dictionaries, each with run, camcol, and list of fields
           - run: 94
             camcol: 1
-            fields: ${range:12,500,50}
+            fields: [12]
```
jeff-regier, line 34: @aakashdp6548 Is this change to `fields` in `base_config.yaml` OK with you? (Will this affect your code?)

conf/base_config.yaml
```diff
@@ -9,7 +9,7 @@ hydra:
     run:
         dir: .
 
-mode: train
+mode: predict
```
jeff-regier, line 12: Better not to change the default mode project-wide

main.py
```diff
@@ -28,14 +28,14 @@ def make_range(start, stop, step=1):
...
-@hydra.main(config_path="conf", config_name="config", version_base=None)
+@hydra.main(config_path="conf", config_name="base_config", version_base=None)
 def main(cfg):
     if cfg.mode == "generate":
         generate(cfg)
     elif cfg.mode == "train":
         train(cfg)
     elif cfg.mode == "predict":
-        predict_sdss(cfg)
+        predict_sdss(cfg, if_plot=True)
```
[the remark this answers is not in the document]
jeff-regier: It's probably better if `predict_sdss` is called without arguments other than `cfg`

tests/test_sdss_reconstruct.py
```diff
@@ -20,13 +19,14 @@ def test_sdss_reconstruct(self, cfg):
         true_bright = true_img_crop - true_bg_crop
 
         bright_pix_mask = (recon_img - 100) > 0
-        res_bright = recon_img[bright_pix_mask] - torch.tensor(true_bright)[bright_pix_mask]
+        recon_img = recon_img.to(cfg.predict.device)
+        res_bright = recon_img[bright_pix_mask] - true_bright[bright_pix_mask]
 
         recon_img += true_bg_crop
-        res_img = recon_img - torch.tensor(true_img_crop)
+        res_img = recon_img - true_img_crop
```
jeff-regier, line 26: Oh good, this looks simpler

conf/base_config.yaml
jeff-regier: Same for `if_crop`, I think I'd called it `do_crop` or just `crop`.

bliss/predict.py
jeff-regier: Yes, that seems good

conf/base_config.yaml
aakashdp6548: Yup, should be fine! The default in base_config should actually be [12] anyway since the other files may not be downloaded, I made the change in #775 as well

jeff-regier (approved):
Looks good! Thanks for your work

PR #775, opened 2023-05-25
Change default fields in base_config; assert paths exist when loading

aakashdp6548:
Add asserts with more descriptive error messages to check that files exist before loading, and fix wrong default fields in base_config

### DIFF
--- a/bliss/simulator/decoder.py
+++ b/bliss/simulator/decoder.py
@@ -1,3 +1,4 @@
+import os
 from typing import Tuple
 
 import galsim
@@ -98,6 +99,7 @@ class ImageDecoder(nn.Module):
         Returns:
             psf_params: tensor of parameters for each band
         """
+        assert os.path.exists(psf_fit_file), f"{psf_fit_file} does not exist"
         # HDU 6 contains the PSF header (after primary and eigenimages)
         data = fits.open(psf_fit_file, ignore_missing_end=True).pop(6).data
         psf_params = torch.zeros(len(bands), 6)
--- a/bliss/simulator/decoder.py
+++ b/bliss/simulator/decoder.py
@@ -99,7 +99,11 @@ class ImageDecoder(nn.Module):
         Returns:
             psf_params: tensor of parameters for each band
         """
-        assert os.path.exists(psf_fit_file), f"{psf_fit_file} does not exist"
+        msg = (
+            f"{psf_fit_file} does not exist. "
+            + "Make sure data files are available for fields specified in config."
+        )
+        assert os.path.exists(psf_fit_file), msg
         # HDU 6 contains the PSF header (after primary and eigenimages)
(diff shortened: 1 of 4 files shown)


bliss/simulator/decoder.py
```diff
@@ -98,6 +99,7 @@ def _get_fit_file_psf_params(psf_fit_file: str, bands: Tuple[int, ...]):
         Returns:
             psf_params: tensor of parameters for each band
         """
+        assert os.path.exists(psf_fit_file), f"{psf_fit_file} does not exist"
```
zhixiangteoh, line 102: nit: Should we suggest a fix in the error message, e.g. "ensure `fields` in field_list config correspond to available data files" or something along those lines?

bliss/simulator/prior.py
```diff
@@ -213,6 +214,7 @@ def _load_color_distribution(self, sdss_fields):
                 sdss_path = Path(self.sdss)
                 camcol_dir = sdss_path / str(run) / str(camcol) / str(field)
                 po_path = camcol_dir / f"photoObj-{run:06d}-{camcol:d}-{field:04d}.fits"
+                assert os.path.exists(po_path), f"{po_path} doesn't exist"
```
zhixiangteoh, line 217: ditto

bliss/surveys/sdss.py
```diff
@@ -57,6 +58,7 @@ def __init__(
         pf_file = f"photoField-{run:06d}-{camcol:d}.fits"
         camcol_path = self.sdss_path.joinpath(str(run), str(camcol))
         pf_path = camcol_path.joinpath(pf_file)
+        assert os.path.exists(pf_path), f"{pf_path} does not exist"
```
zhixiangteoh, line 61: ditto

zhixiangteoh (approved):
Thanks! For the future, might it be worth it to add a very simple, cursory test_case_studies that just runs very basic bliss mode=generate and mode=train on the case study folder.

The case studies folder could be filtered by if it has a config.yaml)? Or perhaps a testing_config parameter/field that is an array of case studies to run this cursory test on--so that whenever you create a new case study you'd have to manually add it to this array?

bliss/simulator/decoder.py
aakashdp6548: Updated error messages

aakashdp6548:
Adding a test for the case studies sounds like a good idea, but I'm concerned about how much extra time it would take to run `generate` and `train` for each case study every PR compared to just checking manually. I think we can leave it for now, and maybe reconsider if/when we add more case studies?

zhixiangteoh:
> Adding a test for the case studies sounds like a good idea, but I'm concerned about how much extra time it would take to run `generate` and `train` for each case study every PR compared to just checking manually. I think we can leave it for now, and maybe reconsider if/when we add more case studies?

It could be a very small test, by overriding the configs with small values.

Issue #727, opened 2023-04-19
optimize encoder performance


jeff-regier:
- [x] explore whether pretraining with crowded field helps
- [x] tune the optimizer scheduler (try OneCycleLR and higher rates up front)
- [x] Consider different batch and images sizes and various amounts of padding
- [x] consider different anchor sizes
- [x] ~~consider using a different NN head (but same backbone) for GalSim parameters~~
- [x] ~~consider replacing/augmenting batchnorm with group-norm or instance-norm~~
- [ ] ~evaluate whether/which log, sigmoid, coadd, and z-score transformations of images improve encoder performance~ (tracked by new issue #829)
~-[ ] tune the encoder architecture both for accuracy and speed~ (tracked by new issue #822)
- [ ] save the best performing model in the `data/pretrained_models` directory

For all these, [Hydra multi-run](https://hydra.cc/docs/tutorials/basic/running_your_app/multi-run/) should be helpful. We'd like to keep a record or what did/didn't work well.

For comparative benchmarks, runs should be repeated 5 to 10 times per dataset with different seeds to quantify variation due to randomness in the weight initialization. We should also repeat the experiments with 2 or 3 (synthetic) datasets to assess whether our results hold up across datasets.

zhixiangteoh:
WIP - https://docs.google.com/spreadsheets/d/1JTkKxC6ApoJUPlYAn51TnSZ6BG0Neykl7-1bSTlyiR8

jeff-regier:
The remaining tasks are tracked by other issues

PR #776, opened 2023-05-26
Add case study that records configs used to explore encoder optimizations

zhixiangteoh:
Each `case_studies/optimize_encoder_727/exp_*`  folder is meant to be used as the `--config-path` argument to the `bliss` executable program, with minimal command-line overrides, like so:

```sh
bliss --config-path case_studies/optimize_encoder_727/exp_pretrain [--multirun]
```


jeff-regier:
I'm closing this PR because we aren't planning to merge it, but instead to keep it to help with #727.

PR #778, opened 2023-05-31
Miscellaneous fixes.

aakashdp6548:
- Update comment in main to reflect changed default config name
- Clean up background processes in generate script on keyboard interrupt
- Fix error with sampling ratios in prior


bliss/simulator/prior.py
```diff
@@ -120,14 +121,17 @@ def sample_prior(self, batch_rcfs) -> TileCatalog:
...
-                select_gal_rcfs.append(np.random.choice(list(self.rcf_gals_rest.values())))
-            if rcf in self.rcf_stars_rest:
-                select_star_rcfs.append(self.rcf_stars_rest[rcf])  # noqa: WPS529
-            else:
-                select_star_rcfs.append(np.random.choice(list(self.rcf_stars_rest.values())))
+            # get value if key exists, otherwise pick randomly
+            select_gal_rcfs.append(
+                self.rcf_gals_rest[rcf]
+                if rcf in self.rcf_gals_rest
+                else random.choice(list(self.rcf_gals_rest.values()))
```
aakashdp6548, line 128: `numpy.random.choice` apparently doesn't play nice with an iterable of objects (in this case a list of tensors), so we're using `random.choice` instead

jeff-regier (approved):
Looks good!

PR #780, opened 2023-06-02
Add average_keep_distance metric for distance only on good matches

aakashdp6548:
`metrics.avg_distance` is the average distance over all matched objects, not just over those we consider a "good" match (i.e. distance < slack). This can be useful in the case where there are no good matches at all, but also having a measure of how good the good matches are would be useful.

image


sawanp813 (approved):
Looks good to me!

PR #781, opened 2023-06-05
Allow config overrides in parallel generate script

zhixiangteoh:
Allow overriding hydra config parameters in `scripts/generate_data_in_parallel.sh`, and also add graceful exit.

### DIFF
--- a/scripts/generate_data_in_parallel.sh
+++ b/scripts/generate_data_in_parallel.sh
@@ -3,23 +3,12 @@
 # gracefully exit
 trap "cleanup" SIGINT SIGTERM EXIT
 
-# Usage: generate_data_in_parallel.sh <relative config_path> <num_files_per_process> <num_processes> <num_workers_per_process>
+# Usage: generate_data_in_parallel.sh <relative config_path> <num_files_per_process> [<config_overrides> <num_processes> <num_workers_per_process>]
 CONFIG_PATH=$1
 N_FILES_PER_PROCESS=$2
 CONFIG_OVERRIDES=${3:-""}
 NUM_PROCESSES=${4:-32}
 N_WORKERS_PER_PROCESS=${5:-0}
-
-# Gracefully exit when interrupted
-pids=()
-cleanup() {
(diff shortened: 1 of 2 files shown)


scripts/generate_data_in_parallel.sh
```diff
@@ -23,10 +27,22 @@ if [ -z "$CONFIG_PATH" ] || [ -z "$N_FILES_PER_PROCESS" ]; then
...
+for pid in "${pids[@]}"; do
+    wait $pid
+done
+
+cleanup() {
+    echo "Killing bliss data generation processes..."
+    for pid in "${pids[@]}"; do
+        kill $pid
+    done
+}
```
aakashdp6548, line 48: I actually added in the cleanup last week when you were out - can you remove that from this PR?
https://github.com/prob-ml/bliss/blob/95ad90f651765118255762781200bee420e03969/scripts/generate_data_in_parallel.sh#L9-L18

scripts/generate_data_in_parallel.sh
```diff
@@ -1,10 +1,14 @@
 #!/bin/bash
 
+# gracefully exit
+trap "cleanup" SIGINT SIGTERM EXIT
+
 # Usage: generate_data_in_parallel.sh <relative config_path> <num_files_per_process> <num_processes> <num_workers_per_process>
```
aakashdp6548, line 6: `# Usage: generate_data_in_parallel.sh <relative config_path> <num_files_per_process> <num_processes> <num_workers_per_process>`
Can you update this comment to reflect the new usage of the script?

aakashdp6548 (changes requested):

scripts/generate_data_in_parallel.sh
zhixiangteoh: Done.

scripts/generate_data_in_parallel.sh
zhixiangteoh: Fixed.

PR #784, opened 2023-06-05
Metrics for galaxy parameters

aakashdp6548:
Added metrics for point estimates of galaxy parameters to make it easier to interpret model performance. Namely, we compute the median absolute error for `{disk/bulge}_flux`, `{disk/bulge}_q`, `{disk/bulge}_hlr` (half-light radius), and `beta_radians` between the estimated and true catalogs. Note that these metrics are only computed when these parameters are available in the true catalog, i.e. from the simulated data. 

Other changes:

- documented the galsim parameters in `decoder.render_galaxy`
- refactored `metrics.update` into two main functions, `_update_detection_metrics` (which contains the old metrics) and `_update_classification_metrics` (which is new and only gets called when classification metrics can be computed)
- add test for classification metrics
- speed up validation by removing `validation_epoch_end` in encoder (no useful behavior) and only plot images on first batch in validation loop

image

### DIFF
--- a/bliss/metrics.py
+++ b/bliss/metrics.py
@@ -1,5 +1,6 @@
-from typing import Dict, Optional
+from typing import Dict, List, Optional
 
+import numpy as np
 import torch
 from einops import rearrange, reduce
 from scipy.sparse import csr_matrix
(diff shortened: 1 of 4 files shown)


bliss/metrics.py
```diff
@@ -49,29 +61,51 @@ def __init__(
...
+            "avg_distance",
+            "avg_keep_distance",
+            "total_true_n_sources",
+            "gal_tp",
+            "gal_fp",
+            "gal_fn",
+            "gal_tn",
+        ]
+        for metric in self.detection_metrics:
+            self.add_state(metric, default=torch.tensor(0.0), dist_reduce_fx="sum")
```
jeff-regier, line 89: Nice

jeff-regier (approved):
Looks good!

PR #790, opened 2023-06-15
Conditional metrics

### LINKED ISSUE #785: more interpretable metrics
Most of our metrics are computed by first matching a predicted catalog to a ground truth catalog. This seems reasonable enough, and indeed, I can't think of a better way to score a predicted catalog that isn't probabilistic. However, it has some quirks/drawbacks. For example, a predicted catalog with lower recall will tend to have fewer matches, and the matched light sources will tend to have a higher SNR, which makes it easier to predict most other properties (e.g., star flux, galaxy shape params). So if we're trying to compare two methods of estimating star flux, for example, the method that has lower star flux MAE, for example, may be genuinely better at estimating star flux, or it may just have lower recall.

For predicted catalogs that are probabilistic, here's an alternative way to score them: rather than using the median or mode of the variational distribution as the predicted catalog, use the median of the variational distribution *conditioned on* the true locations and source types. In this way, we can consider changes in the MAE of predictions (e.g., galaxy half-light radius) in isolation of changes in detection recall/precision and source-type classification accuracy.

aakashdp6548:
Refactor metrics to compute classification metrics conditioned on true source and source type for TileCatalogs (fixes #785). Now we still compute detection metrics based on matches, but if given a TileCatalog, we compute galaxy classification metrics by examining tiles in the estimated catalog corresponding to tiles in the true catalog containing a galaxy. 

Major changes:
- rewrote most of `metrics.BlissMetrics` to compute metrics over TileCatalogs instead of FullCatalogs. There is a `mode` flag that can enable this new behavior; by default, it is set to `BlissMetrics.Mode.Full` since that's probably a more common use case.
- Updated TileCatalog functions to be public for use in metrics, and added `gather_param_at_tiles` as a generic way of getting tile parameters at specific locations
- Updated `encoder.variational_mode` to only compute the median/mode of the variational distributions, without masking any of the tiles (this is already done in `to_full_params` anyway). Added a parameter to determine whether to return a TileCatalog or a FullCatalog

Minor changes:
- ~~Created `bliss/utils.py` for generic utility functions like `column_to_tensor` or `convert_mag_to_flux`~~ (Reverted)
- Updated `test_metrics.test_classification_metrics` to use TileCatalogs


bliss/utils.py
```diff
@@ -0,0 +1,41 @@
+import numpy as np
+import torch
+from torch import Tensor
+
+from bliss.catalog import FullCatalog
```
jeff-regier, line 5: Probably better to avoid creating a "utils" module, see https://www.yanglinzhao.com/posts/utils-antipattern/

bliss/encoder.py
```diff
@@ -123,15 +123,25 @@ def encode_batch(self, batch):
...
+        Returns:
+            Union[FullCatalog, TileCatalog]: Catalog based on predictions.
+        """
         # the mean would be better at minimizing squared error...should we return that instead?
         tile_is_on_array = pred["on_prob"].mode
         # this is the mode of star_log_flux but not the mean of the star_flux distribution
         star_fluxes = pred["star_log_flux"].mode.exp()  # type: ignore
-        star_fluxes *= tile_is_on_array
-        galaxy_bools = pred["galaxy_prob"].mode * pred["on_prob"].mode  # type: ignore
-        star_bools = (1 - pred["galaxy_prob"].mode) * pred["on_prob"].mode  # type: ignore
```
jeff-regier, line 134: Could this change break other parts of the code? If you don't multiple by `pred["on_prob"].mode` here, do we use `est_cat.n_sources` everywhere else to determine what stars/galaxies have been detected?

bliss/utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def convert_mag_to_flux(mag: Tensor, nelec_per_nmgy=987.31) -> Tensor:
```
jeff-regier, line 8: If we're using this function in non-sdss-specific contexts, we may want to call it something like `convert_mag_to_nmgy` instead. Because a nanomaggy is a specific unit of flux.

Also, I'm not sure if `987.31` is the right constant to use outside of sdss. (it could be)

jeff-regier:
Looks really good! Adding a `mode` flag make sense to me.

bliss/metrics.py
```diff
@@ -1,4 +1,8 @@
-from typing import Dict, List, Optional
+# pylint: disable=E1101
+# mypy: disable-error-code="union-attr"
```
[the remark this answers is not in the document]
aakashdp6548: I personally haven't found it too useful, not sure about the others. I added this module-level suppression since mypy really doesn't play nice with union types - it kept complaining about the `Union[TileCatalog, FullCatalog]` not having certain attributes even though the type is checked in `update` to ensure that the catalog is the type I expect it to be.

bliss/encoder.py
aakashdp6548: `TileCatalog.to_full_params` also masks by `is_on_array`, so doing it here was unnecessary (and also not very useful, since none of the galaxy params were masked anyway). I confirmed that the FullCatalog returned by calling variational_mode is the same before and after the changes. 
https://github.com/prob-ml/bliss/blob/bc0e2df666c9672107e5a03861953ca0c6cf5bc3/bliss/catalog.py#L117-L119

Detection metrics use `n_sources` to slice `true_locs` and `est_locs` (since they're sorted with tiles being on at the top):
https://github.com/prob-ml/bliss/blob/bc0e2df666c9672107e5a03861953ca0c6cf5bc3/bliss/metrics.py#L150-L152

Classification metrics also mask by the tile being on and the source being a galaxy/star in the true catalog:
https://github.com/prob-ml/bliss/blob/bc0e2df666c9672107e5a03861953ca0c6cf5bc3/bliss/metrics.py#L195-L202

I think these are the only places that use the output of variational mode.

bliss/utils.py
aakashdp6548: I just reverted this commit since it wasn't related to the metrics at all, I just had it sitting on the branch.

bliss/utils.py
aakashdp6548: Reverted the commit that moved this to utils - now the function is back in `sdss.py`, but it's currently not used by any code. `DecalsFullCatalog` has its own version of the function without the `nelec_per_nmgy` argument - not sure what the units are meant to be, but this is used when loading the Decals catalog:
https://github.com/prob-ml/bliss/blob/2f16bf3074aa4475b899f4292768c18afd7cafe6/bliss/surveys/decals.py#L22-L24

PR #795, opened 2023-06-15
Test API train on 1 epoch

zhixiangteoh:
Make tests run faster. Part of the issue is because test_api now uses base_config and not testing_config, as the "base config". Ideally we would want to be able to use testing_config as the base config to bliss.api.BlissClient.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -6034,4 +6034,4 @@ tqdm = ">=4.64.0"
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.10"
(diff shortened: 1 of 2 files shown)


aakashdp6548:
The times look a lot better than before :) Setup for test_generate is still kind of long, but I think that's because it's running all of the fixtures first (because of the pytest.mark.usefixtures)

40.18s setup    tests/test_api.py::TestApi::test_generate
6.62s call     tests/test_api.py::TestApi::test_generate

Since this just runs in the background on GitHub anyway, I think ~10 minutes is fine for now, but we should revisit it if we add more tests and it starts to bloat more

aakashdp6548 (approved):

jeff-regier:
Thanks @zhixiangteoh !

PR #800, opened 2023-06-16
avoid create new file when tesing plot

XinyueLi1012:
save predict.html in the temporary directory and delete it after testing


jeff-regier (approved):
Looks good, thanks!

Issue #730, opened 2023-04-19
AstroPy integration


jeff-regier:
- [x] Allow users to download pretrained models, one for each survey, from ~~huggingface~~ our git-lfs repo (without installing git-lfs)
- [x] Allow users to fine-tune pretrained models.
- [x] build an interface somewhat like [photutils](https://photutils.readthedocs.io/en/stable/) has
- [x] API to allow users to specify a bounding box and a survey. Then bliss automatically downloads relevant data and provides a catalog.
- [x] Display the inferred catalog using our Bokeh interface (in `predict.py`)
- [x] Allow the user to export the catalog into a fits file by using [`astropy.table`](https://docs.astropy.org/en/stable/table/index.html) and `astropy.fits`. 
- [x] ~~Allow the user to import their own image data into bliss using the [`CCDData` class](https://docs.astropy.org/en/stable/io/unified.html#getting-started-with-image-i-o)~~
- [x] use as many astropy data structures and affiliated packages as possible, especially [astroquery](https://astroquery.readthedocs.io/en/latest/) 
- [x] tutorials/documentation like other affiliate packages have
- [x] develop a tutorial notebook showcasing our AstroPy integration through a typical user session
- [x] #772 
- [x] #853
- [x] #860
- [ ] #845 
- [ ] documentation that's more like the docs for [photutils](https://photutils.readthedocs.io/en/stable/) (e.g., a "Getting started" section, "User documentation" section, and a "Developer documentation" section)
- [ ] test bliss on mac and windows systems (not just linux)
- [ ] setup tox testing to ensure bliss works with multiple versions of Python
- [ ] [apply to become an astropy affiliate](https://www.astropy.org/affiliated/#becoming-an-affiliated-package)

PR #801, opened 2023-06-16
Encode PSF Variation

aakashdp6548:
Previously, #771 updated the simulator to generate images using a variety of PSFs. Now, we also use this information in the encoder, in the form of the deconvolved image and the parameters used to generate the PSF function. 

Major changes:
- When generating the data, we keep track of the PSF used to convolve the image, and use the [Richardson-Lucy](https://scikit-image.org/docs/stable/api/skimage.restoration.html#skimage.restoration.richardson_lucy) method to obtain the deconvolved image. The deconvolution and the power law model parameters used to generate the PSF are always saved when generating data.
- There are two new config parameters in `encoder.input_transform_params`, `use_deconv_channel` and `concat_psf_params` that determine whether to use the deconvolved image and the PSF params during training. If set to true, these are concatenated onto the input tensors as separate channels per band. `get_input_tensor` in encoder.py handles checking config parameters and deciding which inputs to use. This is also extensible for multi-band support, and also for other input transformations like log-transform, z-score, etc.

Minor changes:
- Added range resolver for OmegaConf to `bliss/___init___.py`. Although it's generally not good practice to put code in `__init__.py`, in this case we want this to run regardless of the entry point (main.py, the new API, or tests) and so this is cleaner than duplicating it in each file.
- Updated tests involving generate/decoder functions since some of those now have different return values
- Decreased number of batches generated in tests to make tests run faster

Known issues:
- TestApi won't pass until #730 is merged due to a bug with an environment variable
- Planning to add a case study demonstrating PSF variability in a later PR


bliss/__init__.py
```diff
@@ -0,0 +1,10 @@
+# flake8: noqa
```
jeff-regier, line 1: Does this disable all flake8 checks? If so, could you instead just disable the one that's problematic?

jeff-regier: Is there anywhere else this command could go? I'm not sure where, but I'm concerned that new developers on the project won't find it if it's here, and therefore won't know about it.

bliss/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+import omegaconf
+
+
+def make_range(start, stop, step):
+    return omegaconf.ListConfig(list(range(start, stop, step)))
+
+
+# resolve ranges in config files (we want this to execute at any entry point)
+omegaconf.OmegaConf.register_new_resolver("range", make_range, replace=True)
```
jeff-regier, line 10: Is there anywhere else this command could go? I'm not sure where, but I'm concerned that new developers on the project won't find it if it's here, and therefore won't know about it.

main.py
```diff
@@ -17,15 +16,6 @@
     environ["BLISS_HOME"] = bliss_home.as_posix()
 
 
-# just returning range has issues with referencing other objects, this is more robust
-def make_range(start, stop, step=1):
-    return omegaconf.listconfig.ListConfig(list(range(start, stop, step)))
-
-
-# resolve ranges in config file
-omegaconf.OmegaConf.register_new_resolver("range", make_range, replace=True)
```
jeff-regier, line 26: Should this file be within the bliss directory, e.g. `bliss/main.py` or the contents should be within `bliss/api.py`? That's might hunch. We'd just have to change the pyproject.toml to point to the new location of `main()`.

tests/conftest.py
```diff
@@ -1,22 +1,12 @@
...
 from hydra.utils import instantiate
 
 
-# just returning range has issues with referencing other objects, this is more robust
-def make_range(start, stop, step=1):
-    return omegaconf.listconfig.ListConfig(list(range(start, stop, step)))
-
-
-# resolve ranges in config file
-omegaconf.OmegaConf.register_new_resolver("range", make_range, replace=True)
```
jeff-regier, line 17: Oh nice, I see you found a way to remove duplicated code.

jeff-regier (approved):
Looks good

bliss/__init__.py
aakashdp6548: I'm not really a huge fan of this either, but I'm not sure where else to put it. Currently we have three entry points to bliss - one in main.py, one from the API functions in api.py, and one in the tests. Instead of duplicating it in each of those locations, just having it in the top-level `__init__.py` ensures that it's loaded no matter the entry point, but does run the risk of it being overlooked.

An alternate option could be to put it in its own script, like `bliss/setup.py`, and import that in `__init__.py` instead. It still requires putting something in the init, but it would be just one import instead of additional logic. Do you think that's a better idea?

main.py
aakashdp6548: I think it's worth keeping it separate for now because I think not everything has been ported over to the new API yet, according to @zhixiangteoh. Once everything is accessible via the API it would probably be good to remove main.py.

bliss/__init__.py
aakashdp6548: There isn't a way to disable a specific flake8 check for the whole file in the file itself - I think you need to modify the command line flags for that. If we pull everything but an import out of the file though (see response to other comment) then there wouldn't be a risk of missing something by suppressing the check.

jeff-regier: Got it, seems fine then

main.py
jeff-regier: Sounds good, I look forward to removing main.py in a future PR

bliss/__init__.py
jeff-regier: I guess it's OK here after all.

jeff-regier: Adding a new file with so little code probably wouldn't simplified things

sawanp813 (approved):

sawanp813:
Looks good 👍🏽

PR #804, opened 2023-06-19
Filter tile catalog by flux and restrict tile catalog to brightest source per tile

aakashdp6548:
Added auxiliary functions to `TileCatalog` to
1) Restrict the catalog to the brightest source per tile (for when `max_sources > 1`), and
2) Filter the tile catalog by a given flux range.

`get_brightest_source_per_tile` returns a new `TileCatalog` where only the brightest "on" source per tile is kept. Effectively, this means that there are two conditions to fulfill: the source must have been "on" in the original catalog, and it must be brighter than the other sources (if any) in that tile. The tensors change in size from (b x nth x ntw x **max_sources** x k) to (b x nth x ntw x **1** x k), where b = batch size, nth x ntw is the dimension of an image in tiles, max_sources is the old maximum sources per tile, and k is the dimension of each parameter (e.g. 2 for locs).

`filter_tile_catalog_by_flux` takes a minimum and maximum flux value and returns a new `TileCatalog` where all "on" sources are within that range. Instead of constructing new tensors, this function just masks out the sources that aren't in the range by setting all parameters for that source to 0.

In the image below, the orange sources are the ones remaining after taking the brightest source per tile and then filtering with flux between 10,000 and 500,000. Yellow are sources dropped by the flux filtering, and red are ones dropped by restricting the sources per tile.
image

### DIFF
--- a/bliss/catalog.py
+++ b/bliss/catalog.py
@@ -199,6 +199,68 @@ class TileCatalog(UserDict):
         idx_to_gather = repeat(indices, "... -> ... k", k=param.size(-1))
         return torch.gather(param, dim=1, index=idx_to_gather)
 
+    def _get_fluxes_of_on_sources(self):
+        """Gets fluxes of "on" sources based on whether the source is a star or galaxy.
+
+        Returns:
+            Tensor: a tensor of fluxes of size (b x nth x ntw x max_sources x 1)
+        """
+        fluxes = torch.where(
+            self["galaxy_bools"], self["galaxy_params"][..., 0, None], self["star_fluxes"]
+        )
+        return torch.where(self.is_on_array[..., None], fluxes, torch.zeros_like(fluxes))
+
+    def get_brightest_source_per_tile(self):
+        """Restrict TileCatalog to only the brightest 'on' source per tile.
+
+        Returns:
+            TileCatalog: a new catalog with only one source per tile
+        """
+        if self.max_sources == 1:
+            return self
+
+        # sort by fluxes of "on" sources to get brightest source per tile
+        on_fluxes = self._get_fluxes_of_on_sources()
+        top_idx = on_fluxes.argsort(dim=3, descending=True)[
+            :, :, :, 0:1, 0
+        ]  # 0:1 keeps dims right for slicing
+
+        d = {}
+        for key, val in self.to_dict().items():
+            if key == "n_sources":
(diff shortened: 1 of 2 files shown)


aakashdp6548:
@XinyueLi1012 let me know if this is what we need for filtering the DC2 catalogs, or if any changes need to be made

aakashdp6548:
@jeff-regier I tried to keep this robust to the coming changes to `galaxy_bools`. I think the only parts you will need to change are:
- in `TileCatalog._get_fluxes_of_on_sources`, instead of `self.galaxy_bools` it should change to `self.source_type == TypeEnum.Galaxy`
https://github.com/prob-ml/bliss/blob/b01c6c6bfe9fc995e3eb9e2f38d3d440eb006b09/bliss/catalog.py#L208-L210
- in `test_catalogs.multi_source_tilecat` change `galaxy_bools` to an int tensor of all galaxy source types
https://github.com/prob-ml/bliss/blob/b01c6c6bfe9fc995e3eb9e2f38d3d440eb006b09/tests/test_catalogs.py#L16

jeff-regier (approved):
Great, thanks!

bliss/catalog.py
```diff
@@ -199,6 +199,68 @@ def gather_param_at_tiles(self, param_name: str, indices: Tensor) -> Tensor:
...
+            if key == "n_sources":
+                d[key] = self.n_sources.bool().int()  # send all positive values to 1, 0 to 0
+            else:
+                param_dim = val.size(-1)
+                idx_to_gather = repeat(top_idx, "... -> ... k", k=param_dim)
+                d[key] = torch.take_along_dim(val, idx_to_gather, dim=3)
+
+        return TileCatalog(self.tile_slen, d)
+
+    def filter_tile_catalog_by_flux(self, min_flux=622, max_flux=1e6):
```
jeff-regier, line 239: Probably better not to set defaults for min/max flux, as the user should be fully aware of what filtering is taking place. Also, there aren't really reasonable defaults that apply across surveys.

aakashdp6548: Would it be better if we make the default range (0, torch.inf)? I was thinking having a default value would be good if you just want to specify either a max or min, like to get all values greater than some threshold.

jeff-regier: Yes (0, torch.inf) seems good

PR #816, opened 2023-06-21
Added psf variation case study

aakashdp6548:
Added case study demonstrating the capabilities of the PSF-aware model compared to a model trained on a single field.


jeff-regier (approved):
Nice!

Issue #815, opened 2023-06-21
Flaky tests due to Git LFS non-authenticated rate limiting


zhixiangteoh:
We get `KeyError: 'sha'` when running [tests in test_api](https://github.com/prob-ml/bliss/actions/runs/5334836380/jobs/9667171690) and `KeyError: 'content'`, due to using GitHub API to download files from our Git LFS, as in change in #809.

PR #817, opened 2023-06-21
Fix flaky tests due to device config, GH api usage

### LINKED ISSUE #814: RuntimeError: Expected all tensors to be on the same device, but found at least two devices
Our tests run with `pytest --gpu` stopped passing following PR #788, due to the following error:
```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument other in method wrapper__equal)
```

I've open a PR (#813) that changes the continuous integration process so that we'll catch issues like these going forward before merging.

zhixiangteoh:
- Use pytest config for device configs
- Call GitHub API with auth headers

Fixes #814, #815.


jeff-regier (approved):
Great 👍

PR #818, opened 2023-06-21
Multi-Band E2E PR

sawanp813:
Prior merge added support for a deconvolution channel and PSF parameter appending to the encoder input. The entire model only considers r-band encoding and cataloging. This PR extends this to all five bands specified by SDSS (u, g, r, i, z). Not only do we predict relevant parameters across all five bands (e.g. flux), but BLISS is now trained with five-bands of information to perform cataloging.

Major changes:
- Modifying encoder parameters to predict all fluxes, adjusts loss computation, and introduces input pre-processing support (via z-score)
- Adjusts metrics to compute across all five bands
- Adding the new five-band model as the standard model used for testing and in API (`z_score_five_band.pt`)
- Rewriting test cases to account for multi-band predictions (e.g. `test_sdss_reconstruct.py`, `test_simulate.py`)
- New `transforms.py` module for any future input pre-processing to be done in `encoder.py`
Minor changes:
- Plotting co-added images as opposed to single r-band

Issues
- Bad predictions for bright light sources force adjustments to `test_sdss_reconstruct`

Notes
- Per-pixel normalization schemes did not work nearly as effectively as whole-image normalization


bliss/conf/base_config.yaml
```diff
@@ -21,13 +21,13 @@ paths:
...
-    valid_n_batches: 10  # 256
+    n_batches: 1
+    valid_n_batches: 0  # 256
     fix_validation_set: true
-    num_workers: 32
+    num_workers: 0
     sdss_fields:  # Used in simulator for sampling backgrounds and PSFs for decoder
-        dir: ${paths.sdss}
-        bands: [2]
+        dir: /data/sdss
```
aakashdp6548, line 29: Were these changed intentionally? We should at least keep `dir: ${paths.sdss}` since there's no guarantee a user will have a /data directory.

bliss/simulator/simulated_dataset.py
```diff
@@ -188,11 +188,13 @@ def __init__(
         self,
         train_n_batches: int,
         batch_size: int,
+        bands: List,
         num_workers: int,
         cached_data_path: str,
         file_prefix: str,
         val_split_file_idxs: List[int],
         test_split_file_idxs: List[int],
+        z_score_transform: bool,
```
aakashdp6548, line 197: Do we still need this if the z-scoring is now being done in the encoder?

bliss/conf/base_config.yaml
```diff
@@ -67,13 +67,17 @@ cached_simulator:
...
     file_prefix: ${generate.file_prefix}
+    bands: [0, 1, 2, 3, 4]
     val_split_file_idxs: null # inclusive range
     test_split_file_idxs: null # inclusive range
+    z_score_transform: True
 
 encoder:
     _target_: bliss.encoder.Encoder
-    bands: [2]
+    z_score: True
```
aakashdp6548, line 79: Similar to other comment, do we still need this here?

bliss/encoder.py
```diff
@@ -37,6 +37,7 @@ def __init__(
         self,
         architecture: DictConfig,
         bands: list,
+        z_score: bool,
```
aakashdp6548, line 40: It would probably be cleaner to include this in the `input_transform_params` dict

bliss/metrics.py
```diff
@@ -43,10 +43,19 @@ class BlissMetrics(Metric):
...
-    disk_flux: List
-    bulge_flux: List
+    # Classification metrics
+    disk_frac: List
+    bulge_frac: List
+    u_flux_s: List
+    g_flux_s: List
+    r_flux_s: List
+    i_flux_s: List
+    z_flux_s: List
```
aakashdp6548, line 53: Are these being computed somewhere? Unless I'm just missing it, it looks like the old star flux calculation was removed but the new ones haven't been added.

main.py
```diff
@@ -17,8 +17,8 @@
 
 
 # config_path should be overriden when running `bliss` poetry executable
-# e.g., `bliss --cp case_studies/summer_template -cn config`
-@hydra.main(config_path="bliss/conf", config_name="base_config", version_base=None)
+# e.g., `bliss --cp case_studies/summer_template`
+@hydra.main(config_path="case_studies/summer_template", config_name="config", version_base=None)
```
aakashdp6548, line 21: This probably shouldn't default to `case_studies/summer_template` since that's just a template config

tests/testing_config.yaml
```diff
@@ -21,9 +21,9 @@ simulator:
         batch_size: 32
         prob_galaxy: 0.5
         star_flux_min: 1e5
-    background:
-        _target_: bliss.simulator.background.ConstantBackground
-        background: [865.0]
+    # background:
+    #     _target_: bliss.simulator.background.SimulatedSDSSBackground
+    #     sdss_fields: ${simulator.sdss_fields}
```
aakashdp6548, line 26: Should these lines be uncommented?

tests/testing_config.yaml
```diff
@@ -56,7 +56,8 @@ training:
 
 predict:
     decals_frame: ${paths.decals}/tractor-3366m010.fits
-    weight_save_path: ${paths.pretrained_models}/sdss.pt
+    weight_save_path: ${paths.pretrained_models}/r_long.pt
```
aakashdp6548, line 59: Did you also intend to add this model to the repo? If not, we should probably keep the path to `sdss.pt`

aakashdp6548 (changes requested):

bliss/conf/base_config.yaml
sawanp813: That was unintentional, thanks for catching that!

bliss/conf/base_config.yaml
sawanp813: I wanted to keep this one optional in case a user wants to do single-band experiments with or without z-scoring. I have it set in encoder so that if multiple bands are specified in encoder, `self.z_score` is automatically set to `True` regardless of what is specified in the config. If single band is specified, the config specifiation is used instead.

bliss/simulator/simulated_dataset.py
sawanp813: If we're set on z-scoring only being done in the encoder then this is definitely not needed. Haven't noticed a decrease in speed as of yet so think it should be fine to remove this.

bliss/metrics.py
sawanp813: I had to take it out because of some failed pre-commit checks. Currently adding them back!

tests/testing_config.yaml
sawanp813: They shouldn't be there at all - I forgot to mention this to @jeff-regier earlier but if we want to remove support for ConstantBackground then this isn't needed in the testing config at all. The reason for removing ConstantBackgrounds is that z-scoring a constant background results in NaN's since the standard deviation of the background would just be 0.

tests/testing_config.yaml
sawanp813: This is temporary so one of the test cases can be debugged on my end (`test_simulate.py`) until I create some new example tile catalogs.

bliss/encoder.py
sawanp813: I believe it is, I just forgot to remove this before committing. Will fix!

sawanp813:
I think I'm going to hold off on including `test_multiband.py` just because there is already a lot going on in the above files. Think it would be better to get all of the multi-band changes included and then fix up that test case to avoid more delays.

bliss/conf/base_config.yaml
```diff
@@ -21,13 +21,13 @@ paths:
 
 simulator:
     _target_: bliss.simulator.simulated_dataset.SimulatedDataset
-    n_batches: 128
-    valid_n_batches: 10  # 256
+    n_batches: 1
+    valid_n_batches: 0  # 256
     fix_validation_set: true
-    num_workers: 32
+    num_workers: 0
```
jeff-regier, line 27: We probably still want 32 workers by default, right?

jeff-regier: And is there a reason `n_batches` and `valid_n_batches` changes?

bliss/encoder.py
```diff
@@ -61,13 +61,20 @@ def __init__(
...
+        self.sdss_bands = ["u", "g", "r", "i", "z"]  # NOTE: SDSS-specific naming conventions
         self.n_bands = len(self.bands)
         self.optimizer_params = optimizer_params
         self.scheduler_params = scheduler_params if scheduler_params else {"milestones": []}
         self.input_transform_params = input_transform_params
 
         self.tile_slen = tile_slen
 
+        # automatically z-score if number of bands is greater than 1.
+        if self.n_bands > 1:
```
jeff-regier, line 73: Seems better to always do what's in the config file (i.e., not do anything automatically)

bliss/encoder.py
```diff
@@ -137,6 +151,16 @@ def get_input_tensor(self, batch):
...
+        if self.input_transform_params.get("z_score"):
+            assert (
+                batch["background"][0, 0].std() > 0
+            ), "Constant backgrounds not supported for multi-band encoding"
+            inputs[0] = (
+                batch["images"] - torch.mean(batch["images"], dim=(2, 3), keepdim=True)
+            ) / torch.std(batch["images"], dim=(2, 3), keepdim=True)
+            inputs[1] = (
+                batch["background"] - torch.mean(batch["background"], dim=(2, 3), keepdim=True)
+            ) / torch.std(batch["background"], dim=(2, 3), keepdim=True)
```
jeff-regier, line 163: The black autoformater strikes again! Will you try writing this with intermediate variables so it's a bit easier to read?

bliss/encoder.py
```diff
@@ -137,6 +151,16 @@ def get_input_tensor(self, batch):
             ), "concat_psf_params specified but psf params not present in data"
             n, _, h, w = batch["images"].size()
             inputs.append(batch["psf_params"].view(n, 6, 1, 1).expand(n, 6, h, w))
+        if self.input_transform_params.get("z_score"):
+            assert (
+                batch["background"][0, 0].std() > 0
+            ), "Constant backgrounds not supported for multi-band encoding"
+            inputs[0] = (
+                batch["images"] - torch.mean(batch["images"], dim=(2, 3), keepdim=True)
+            ) / torch.std(batch["images"], dim=(2, 3), keepdim=True)
```
jeff-regier, line 160: Woah, is the z-scoring computing the standard deviation over the whole image? I thought we were just dividing by the square root of the background brightness at each pixel. Do this normalization scheme work better than just subtracting the background and dividing by the square root of the back intensity?

bliss/encoder.py
```diff
@@ -244,16 +284,22 @@ def _get_loss(self, pred: Dict[str, Distribution], true_tile_cat: TileCatalog):
...
+            star_flux_loss = -pred[f"star_log_flux {bnd}"].log_prob(star_log_fluxes[:, :, :, i])
+            star_flux_loss *= true_star_bools
+            loss += star_flux_loss
+            loss_with_components[f"star_flux_loss {bnd}"] = (
+                star_flux_loss.sum() / true_star_bools.sum()
+            )
 
         # galaxy properties loss
-        galsim_names = ["flux", "disk_frac", "beta_radians", "disk_q", "a_d", "bulge_q", "a_b"]
+        fluxes = [f"flux {bnd}" for bnd in self.sdss_bands]
```
jeff-regier, line 301: `"flux {bnd}" -> "flux_{bnd}"` (for consistency with the other galsim names)

bliss/encoder.py
```diff
@@ -324,6 +370,10 @@ def test_step(self, batch, batch_idx):
 
     def predict_step(self, batch, batch_idx, dataloader_idx=0):
         """Pytorch lightning method."""
+        # NOTE: Debugging predict issue in test_simulate.py - extra dimension
```
jeff-regier, line 373: Can we fix this issue in `test_simulate.py` rather than introducing an extra conditional block in the encoder?

bliss/metrics.py
```diff
@@ -43,10 +43,19 @@ class BlissMetrics(Metric):
...
+    u_flux_g: List
+    g_flux_g: List
+    r_flux_g: List
+    i_flux_g: List
+    z_flux_g: List
```
jeff-regier, line 58: Can we store the fluxes in a tensor? That will help us to generalize the code to work for other surveys, which may have a different number of bands.

bliss/simulator/prior.py
```diff
@@ -229,13 +229,13 @@ def _load_color_distribution(self, sdss_fields):
...
-                        )
+                            fluxes[obj] / fluxes[obj][2]
+                        )  # noqa: WPS220  # flux relative to r-band
                         star_fluxes_rest.append(ratios)  # noqa: WPS220
                     elif objc_type[obj] == 3 and thing_id[obj] != -1 and torch.all(fluxes[obj] > 0):
                         ratios = (  # noqa: WPS220
-                            fluxes[obj] / fluxes[obj][2]  # flux relative to r-band
-                        )
+                            fluxes[obj] / fluxes[obj][2]
+                        )  # noqa: WPS220  # flux relative to r-band
```
jeff-regier, line 238: Maybe put the comment above the code so we have avoid this formatting

main.py
```diff
@@ -17,7 +17,7 @@
 
 
 # config_path should be overriden when running `bliss` poetry executable
-# e.g., `bliss --cp case_studies/summer_template -cn config`
+# e.g., `bliss --cp case_studies/summer_template`
```
jeff-regier, line 20: accidental change?

bliss/encoder.py
sawanp813: I was thinking that this might be better since we get the weird no prediction behavior when using the non-z-scored multi-band input. It might be an easy mistake to specify multiple bands but forget to specify the z-score transform or some other pre-processing. I can just keep it to what is in the config though if you'd prefer that.

bliss/encoder.py
sawanp813: Sounds good!

bliss/encoder.py
sawanp813: This is the z-scoring I've been doing for all of the z-scored models. I haven't tried subtracting the background and dividing by the square root of the background intensity. Should I experiment with that first?

bliss/metrics.py
sawanp813: Sounds good.

main.py
sawanp813: Yes, fixed!

bliss/encoder.py
jeff-regier: How about raising a warning if a user doesn't enable the z-score transform for multiband, instead of automatically turning on z-scoring and ignoring the config file?

bliss/encoder.py
jeff-regier: Yes please. Subtracting the background and dividing by the square root of the background intensity (per pixel) seems a lot less mysterious/complex/hard-to-analyze. Let's hope it works just as well.

bliss/metrics.py
aakashdp6548: I think we still need a list since we want to compute the median error over a variable number of sources per tile, so a fixed-size tensor wouldn't be appropriate. But we can just have it be a single list which stores all five bands, instead of one list per band.

Additional context: torchmetrics.Metric requires state to be either a tensor, or a list of tensors. So before, `star_flux` was a list of `n_d x 1` tensors, where `d = total # of sources in a batch`. Now, we can have it be a list of `n_d x 5` tensors instead, and compute the median over each band separately in `compute()`.

bliss/encoder.py
sawanp813: That's a better idea, sounds good.

bliss/metrics.py
sawanp813: I'll hold off on this for now. I like the idea of having a list of `n_d x 5` tensors instead - it should look more clean.

jeff-regier (approved):
Please merge when ready. We can fix any issues in follow-up PRs.

aakashdp6548 (approved):

PR #824, opened 2023-06-23
simplify generate_data_in_parallel.sh

jeff-regier:


scripts/generate_data_in_parallel.sh
```diff
@@ -1,26 +1,31 @@
...
+
+# gracefully exit on these conditions
+trap "cleanup" SIGINT SIGTERM
+
+NUM_PROCESSES=${1:-""}
+CONFIG_OVERRIDES=${2:-""}
+
+# Error if NUM_PROCESS not specified
+if [ -z "$NUM_PROCESSES" ]; then
+    echo "Usage: generate_data_in_parallel.sh <num_processes> [<config_overrides>]"
```
jeff-regier, line 21: I think this interface is easier to understand/use: the only mandatory argument, num_processes, is used by this script and not passed on the bliss/hydra. The rest of the arguments pass through to bliss/hydra.

bliss/conf/base_config.yaml
```diff
@@ -119,13 +119,14 @@ encoder:
...
+    process_index: 0
     n_workers_per_process: 0
     n_batches: ${simulator.n_batches}
     batch_size: ${simulator.prior.batch_size}
-    max_images_per_file: 5000
+    max_images_per_file: 4096
     cached_data_path: ${paths.data}/cached_dataset
     file_prefix: dataset
-    files_start_idx: 0
+    num_files_per_process: 128
```
zhixiangteoh, line 129: Actually, what if we removed `process_index` and `num_files_per_process` from this base config, to make it cleaner, but append `+generate.process_index` in the calls to the `bliss` executable in the parallel generate script? (See next few comments.) This abstracts away these details about multi-process data generation, since we ship this base config file but not the /scripts directory.

Also, currently `n_batches` together with `batch_size` determines the total number of images, and together with `max_images_per_file` determines the total number of files to be generated (by this process). Meaning either one of `num_files_per_process` or `n_batches` could be unnecessary.

bliss/generate.py
```diff
@@ -24,7 +24,7 @@
 def generate(cfg: DictConfig):
     max_images_per_file = cfg.generate.max_images_per_file
     cached_data_path = cfg.generate.cached_data_path
-    files_start_idx = cfg.generate.files_start_idx
+    files_start_idx = cfg.generate.process_index * cfg.generate.num_files_per_process
```
zhixiangteoh, line 27: Following the above, we could then change this to

```python
# code from few lines below, can be shifted up; or these two new lines shifted down
n_files = -(cfg.generate.n_batches * bs // -images_per_file) # note: this is technically "n_files for this process"

process_index = cfg.generate.process_index or 0
files_start_idx = process_index * n_files
```

scripts/generate_data_in_parallel.sh
```diff
@@ -1,26 +1,31 @@
...
     exit 1
 fi
 
 pids=()
 for ((i=0; i<$NUM_PROCESSES; i++));
 do
-    files_start_idx=$((i * N_FILES_PER_PROCESS))
-    bliss -cp $CONFIG_PATH 'mode=generate' "simulator.num_workers=$N_WORKERS_PER_PROCESS" "generate.files_start_idx=$files_start_idx" $CONFIG_OVERRIDES &
+    bliss 'mode=generate' "generate.process_index=$i" $CONFIG_OVERRIDES &
```
zhixiangteoh, line 28: Following the above, this could then be modified slightly to:

```sh
bliss 'mode=generate' "+generate.process_index=$1" $CONFIG_OVERRIDES &
```

zhixiangteoh (commented):
Thanks! Do we want to abstract away the multi-process generation details from base config that gets shipped to user?

bliss/generate.py
jeff-regier: Seems good but probably better done as a separate PR

jeff-regier: Oh I see, you're proposing hiding `process_index` from the user?

jeff-regier: That's easy enough. I got confused by the `n_files = ...` line but I see that's already in the code.

zhixiangteoh: Yup, thanks!

scripts/generate_data_in_parallel.sh
zhixiangteoh: Agreed

zhixiangteoh (approved):
Thanks!

bliss/api.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
```
zhixiangteoh, line 1: I believe we need this because of executing `main` from this file as `bliss` executable?

jeff-regier: Still works fine for me because there's a shebang in the file named `bliss` that poetry generates:
```
#!/home/regier/bliss/.venv/bin/python
import sys
from bliss.api import main

if __name__ == '__main__':
    sys.exit(main())
```

Issue #820, opened 2023-06-22
sample_star_fluxes uses galaxy parameters


aakashdp6548:
`_sample_star_fluxes` should use `self.star_flux_min`, `self.star_flux_max`, and `self.star_flux_alpha` instead of the galaxy versions of the parameters.
https://github.com/prob-ml/bliss/blob/e2a09160dae464215a25bca466936e6334ce01db/bliss/simulator/prior.py#L171-L175

Issue #812, opened 2023-06-20
Bug: Metrics computation fails for FullCatalogs if "galaxy_params" available


aakashdp6548:
If we try to compute metrics between two FullCatalogs that have "galaxy_params", the classification metrics fail. This wasn't implemented in #790 because the idea was you would just set `mode=Mode.Tile` if you have the galaxy params, but we shouldn't rely on that assumption. Instead, we can just compute the classification metrics the old way (based on the location matching).

PR #832, opened 2023-06-23
Miscellaneous fixes to catalogs and metrics

### LINKED ISSUE #810: Get properties from catalog using bracket operator
Instead of having some parameters accessible through the dot operator (`cat.n_sources`) and others through the bracket operator (`cat["star_fluxes"]`), maybe it would be better to add additional logic in `__getitem__` to just return the corresponding property? So something like
```
def __getitem__(self, key: str):
    if hasattr(self, key):
        return getattr(self, key)
    return super.__getitem__(key)
```
This way everything would be accessible through the bracket operator. Otherwise it seems like there's not much point having them extend UserDict.

_Originally posted by @aakashdp6548 in https://github.com/prob-ml/bliss/pull/807#discussion_r1235377019_

aakashdp6548:
Fixes #810, #812, #820. Also moved the catalog filtering functions in `test_metrics.py` to the appropriate catalogs, since they can be more generally useful.

### #810 - Get catalog properties using bracket operator
Now we can access attributes like `locs` and also properties like `galaxy_bools` from `TileCatalog` and `FullCatalog` using the bracket operator, like `tile_cat["galaxy_bools"]` for consistency with the actual keys of the dictionary.

I also moved `metrics.BlissMetrics.Mode` -> `metrics.MetricsMode` to avoid issues with comparisons with relative imports.

Alternatives considered: adding `locs` and `n_sources` as keys to the dict, instead of keeping them separate. These are being used in a lot of places though, so it would be a massive change. Also this alone wouldn't help with properties like "star_bools", "galaxy_bools", and "is_on_mask" (i.e. you still have to know beforehand which are keys and which are properties).

### #812: Fix classification metrics computation with FullCatalogs
Computing metrics between FullCatalogs that have `galaxy_params` available will now use the old matching-based approach to compute them, instead of crashing.

### #820 - Fix star flux sampling parameters
Now we use the the star flux parameters for sampling star flux instead of the galaxy parameters. This shouldn't have affected any previous results since the default values are the same.

### Filtering catalogs by RA/DEC
Refactored `test_metrics._get_photo_cat` into `PhotoFullCatalog.restrict_by_ra_dec`, and `test_metrics._get_decals_cat` into `DecalsFullCatalog.from_file`, since the behavior of these functions is more generally useful than just in tests.


jeff-regier (approved):
Excellent!

Issue #838, opened 2023-06-26
api.fullcat_to_astropy_table can crash when there is only one source in a catalog


aakashdp6548:
`api.fullcat_to_astropy_table` squeezes the dimensions of the catalog as part of the conversion from a FullCatalog to a table. If the tensor is all one-dimensional (e.g. 1x1x1), which occurs when there is only one batch, one source, and a one-dimensional parameter (e.g. `star_log_fluxes`), `squeeze` removes all of the dimensions, so we get an IndexError.
https://github.com/prob-ml/bliss/blob/e076f32770a5bd74fa088277886cb164d02499ec/bliss/api.py#L222-L224

```
>>> x = torch.tensor(1).reshape(1, 1, 1)
>>> x.shape
torch.Size([1, 1, 1])
>>> x.squeeze().shape
torch.Size([])
>>> x.squeeze().shape[0]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IndexError: tuple index out of range
```

PR #839, opened 2023-06-26
Rewrite API tests to mock downloads, simplify test behavior

aakashdp6548:
This PR simplifies the behavior in `test_api` to not require downloading any files from a remote server. Instead, we mock `requests.get` and `requests.post` to mimic the expected behavior, allowing us to still test functions like `download_git_lfs_file` without needing to actually download anything.

In the new `test_load_pretrained_weights` and `test_load_survey`, we don't actually need the data for anything, so we can just check that the mocked data was saved to the right filenames. In `test_predict_sdss_default_rcf` we do actually need the data to make predictions, so we just update the paths to use the `$BLISS_HOME/data` paths instead of the temporary directory.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1,12 +1,14 @@
+import bz2
+import gzip
 import os
-import shutil
 from pathlib import Path
 
 import pytest
+import requests
+from mock_tests import mock_get, mock_post, mock_train
 from omegaconf import OmegaConf
 
-from bliss.api import BlissClient
-from bliss.utils.download_utils import download_git_lfs_file
+from bliss import api
 
 
 @pytest.fixture(scope="session")
@@ -16,7 +18,7 @@ def cwd(tmpdir_factory):
 
 @pytest.fixture(scope="class")
 def bliss_client(cwd, cfg):
-    client = BlissClient(str(cwd))
+    client = api.BlissClient(str(cwd))
     # Hack to apply select conftest.py overrides, since client.base_cfg should be private
     overrides = {
         "training.trainer.accelerator": cfg.training.trainer.accelerator,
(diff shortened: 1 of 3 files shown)


tests/test_api.py
```diff
@@ -30,139 +32,80 @@ def bliss_client(cwd, cfg):
 @pytest.fixture(scope="class")
 def cached_data_path_api(bliss_client):
     bliss_client.cached_data_path = bliss_client.cwd + "/data/cached_dataset"
-    bliss_client.generate(n_batches=3, batch_size=8, max_images_per_file=16)
+    bliss_client.generate(n_batches=2, batch_size=5, max_images_per_file=5)
```
zhixiangteoh, line 35: Why 5? I think we should stick to at least 8. And reason for `max_images_per_file` being 2 x `batch_size` is also as a sanity check that two batches get "flattened"/itemized and saved as expected.

tests/test_api.py
```diff
@@ -16,7 +18,7 @@ def cwd(tmpdir_factory):
 
 @pytest.fixture(scope="class")
 def bliss_client(cwd, cfg):
-    client = BlissClient(str(cwd))
+    client = api.BlissClient(str(cwd))
```
zhixiangteoh, line 21: Could we stick to

```python
from bliss.api import BlissClient

client = BlissClient(...)
```
?

tests/test_api.py
```diff
@@ -30,139 +32,80 @@ def bliss_client(cwd, cfg):
...
-@pytest.fixture(scope="class")
-def weight_save_path(bliss_client, pretrained_weights_filename):
+def weight_save_path(bliss_client, cfg):
+    """Train model for 1 epoch and return path to saved model."""
     weight_save_path = "tutorial_encoder/0_fixture.pt"
     bliss_client.train_on_cached_data(
         weight_save_path=weight_save_path,
         train_n_batches=1,
-        batch_size=8,
+        batch_size=5,
```
zhixiangteoh, line 46: 8 here?

tests/test_api.py
```diff
@@ -30,139 +32,80 @@ def bliss_client(cwd, cfg):
...
+    def test_load_survey(self, bliss_client, monkeypatch):
+        monkeypatch.setattr(requests, "get", mock_get)
+        monkeypatch.setattr(requests, "post", mock_post)
+        monkeypatch.setattr(gzip, "decompress", lambda x: x)
+        monkeypatch.setattr(bz2, "decompress", lambda x: x)
+
+        download_dir = "data"
+        bliss_client.load_survey("sdss", 94, 1, 12, download_dir)
+        assert os.path.exists(f"{bliss_client.cwd}/{download_dir}/94/1/photoField-000094-1.fits")
+        assert os.path.exists(
```
zhixiangteoh, line 97: Nit: prefer

```python
from pathlib import Path

Path(...).exists()
```

tests/test_api.py
```diff
@@ -30,139 +32,80 @@ def bliss_client(cwd, cfg):
...
         train_n_batches=1,
-        batch_size=8,
+        batch_size=5,
         val_split_file_idxs=[1],
-        pretrained_weights_filename=pretrained_weights_filename,
-        training={"n_epochs": 1, "trainer": {"check_val_every_n_epoch": 1, "log_every_n_steps": 1}},
+        training={
+            "n_epochs": 1,
+            "trainer": {"check_val_every_n_epoch": 1, "log_every_n_steps": 1},
+            "pretrained_weights": cfg.predict.weight_save_path,
```
zhixiangteoh, line 51: Why is this using `predict.weight_save_path`?

tests/test_api.py
```diff
@@ -30,139 +32,80 @@ def bliss_client(cwd, cfg):
...
+            "pretrained weights "
+            + f"{bliss_client.cwd}/data/pretrained_models/{filename} "
+            + "not found"
+        )
+        assert Path(
+            bliss_client.cwd + f"/data/pretrained_models/{filename}"
+        ).exists(), not_found_err_msg
+
+    def test_train(self, bliss_client, monkeypatch):
+        monkeypatch.setattr(api, "_train", mock_train)
```
zhixiangteoh, line 85: Instead of overriding `bliss.api._train`, could we override `bliss.train` instead? And so we can import as `from bliss.api import BlissClient` too

zhixiangteoh (changes requested):

tests/test_api.py
aakashdp6548: 5 is the minimum value we can use since batch norm requires `batch_size > 4`. Is there anything 8 would give us that 5 doesn't? Tests are taking longer again with multi band so any time saved is useful. I can change it to 3x5 instead if we want to keep testing the max per file behavior.

tests/test_api.py
```diff
@@ -30,139 +32,80 @@ def bliss_client(cwd, cfg):
...
-            )
-        bliss_client.plot_predictions_in_notebook()
-
-
-def download_decals_base_from_git(download_dir: str):
-    assert os.environ["GITHUB_TOKEN"] is not None, "GITHUB_TOKEN environment variable not found"
+    def test_load_pretrained_weights(self, bliss_client, monkeypatch):
+        monkeypatch.setattr(requests, "get", mock_get)
+        monkeypatch.setattr(requests, "post", mock_post)
+        monkeypatch.setenv("GITHUB_TOKEN", "test")
```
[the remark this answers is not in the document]
aakashdp6548: Good point, not sure why I did this. I'll remove it.

tests/test_api.py
aakashdp6548: Temporary workaround until #838 is fixed. Training from uninitialized weights causes it to only make one prediction, which causes the issues with `squeeze`.

tests/test_api.py
zhixiangteoh: Yea 5 works, I just thought 8 is nicer since it’s even (and the lowest power-of-2 that is >4).

tests/test_api.py
```diff
@@ -30,139 +32,80 @@ def bliss_client(cwd, cfg):
...
+
+    def test_load_survey(self, bliss_client, monkeypatch):
+        monkeypatch.setattr(requests, "get", mock_get)
+        monkeypatch.setattr(requests, "post", mock_post)
+        monkeypatch.setattr(gzip, "decompress", lambda x: x)
+        monkeypatch.setattr(bz2, "decompress", lambda x: x)
+
+        download_dir = "data"
+        bliss_client.load_survey("sdss", 94, 1, 12, download_dir)
+        assert os.path.exists(f"{bliss_client.cwd}/{download_dir}/94/1/photoField-000094-1.fits")
```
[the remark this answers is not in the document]
aakashdp6548: `load_survey` "downloads" the mocked data and saves it to the files. So they exist, they just contain garbage, but since we don't need the data in this test case that doesn't matter.

tests/test_api.py
zhixiangteoh: And this is good becos we don’t re-test bliss.train that already gets tested in tests like end_to_end

tests/test_api.py
zhixiangteoh: Should this be blocked by #838? Or will should fix to #838 handle this?

tests/test_api.py
aakashdp6548: It should be the same either way, since `bliss.api._train` is just an alias for `bliss.train.train` right? I just added a separate import for BlissClient so now it has both api and BlissClient in the global namespace.

tests/test_api.py
aakashdp6548: I think it should handle it, but I'm not sure if there might be other downstream issues. Probably better to leave it in for this PR, and remove it in the PR for #838 if it does indeed fix it.

tests/test_api.py
aakashdp6548: Changed.

tests/test_api.py
aakashdp6548: Keeping this at 5 as per the other comment.

tests/test_api.py
aakashdp6548: Changed

tests/test_api.py
zhixiangteoh: I don't think we should use an underscore import `_train`; like other underscore names it should be treated as private; just slightly cleaner to use 

```python
from bliss import train

monkeypatch.setattr(train, "train", mock_train)
```

tests/test_api.py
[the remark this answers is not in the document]
zhixiangteoh: Ah got it, makes sense. Thanks.

tests/test_api.py
```diff
@@ -64,13 +64,12 @@ def test_get_dataset_file(self, bliss_client, cached_data_path_api):
...
         monkeypatch.setattr(requests, "get", mock_get)
         monkeypatch.setattr(requests, "post", mock_post)
-        monkeypatch.setenv("GITHUB_TOKEN", "test")
 
         filename = "sdss_pretrained.pt"
         bliss_client.load_pretrained_weights_for_survey(
             survey="sdss",
             filename=filename,
-            request_headers={"Authorization": f"Bearer {os.environ['GITHUB_TOKEN']}"},
+            request_headers={"Authorization": "test"},
```
zhixiangteoh, line 72: Can remove?

zhixiangteoh (approved):
Thanks!

aakashdp6548:
@zhixiangteoh  I made a few more changes, including mocking `generate` and `predict_sdss`. Can you take another look and make sure everything still looks good?

`test_api` now runs in 35s on both CPU and GPU :grinning:

tests/test_api.py
```diff
@@ -29,38 +24,18 @@ def bliss_client(cwd, cfg):
...
-        bliss_client.cached_data_path = cached_data_path_api
+    def test_get_dataset_file(self, cfg, bliss_client):
+        bliss_client.cached_data_path = cfg.paths.data + "/tests/multiband_data"
         dataset0 = bliss_client.get_dataset_file(filename="dataset_0.pt")
         assert isinstance(dataset0, list), "dataset0 must be a list"
 
+    def test_generate(self, bliss_client, monkeypatch):
+        monkeypatch.setattr(api, "_generate", mock_generate)
+        bliss_client.generate(n_batches=2, batch_size=1, max_images_per_file=2)
+        assert Path(bliss_client.cached_data_path).exists()
```
zhixiangteoh, line 37: I think we want to at least generate one dataset file and check that it exists in this `bliss_client.cached_data_path`? So maybe we shouldn't mock `_generate`? Or at least return a test file from `mock_generate`?

tests/test_api.py
```diff
@@ -104,7 +83,8 @@ def test_load_survey(self, bliss_client, monkeypatch):
             f"{bliss_client.cwd}/{download_dir}/94/1/12/psField-000094-1-0012.fits"
         ).exists()
 
-    def test_predict_sdss_default_rcf(self, bliss_client, weight_save_path, cfg):
-        paths = {"sdss": cfg.paths.sdss, "decals": cfg.paths.decals}
-        bliss_client.predict_sdss(weight_save_path=weight_save_path, paths=paths)
+    def test_predict_sdss_default_rcf(self, cfg, bliss_client, monkeypatch):
+        monkeypatch.setattr(api, "_predict_sdss", mock_predict_sdss)
+        # cached predict data stored at cfg.paths.data, copied to temp dir in mock_predict_sdss
+        bliss_client.predict_sdss("test_path", paths={"data": cfg.paths.data})
```
zhixiangteoh, line 89: I am conflicted about overriding `paths.data` because it is not something an end user would typically do.

tests/test_api.py
```diff
@@ -84,6 +59,10 @@ def test_train(self, bliss_client, monkeypatch):
         monkeypatch.setattr(api, "_train", mock_train)
         bliss_client.train("test", training={"n_epochs": 1})
 
+    def test_train_on_cached_data(self, cfg, bliss_client, monkeypatch):
+        monkeypatch.setattr(api, "_train", mock_train)
+        bliss_client.train_on_cached_data("test.pt", 4, 4, None, None, "test_pretrained.pt")
```
zhixiangteoh, line 64: Why `"test.pt"` / these arguments? In general, the goal of test_api is to mimic how an end user would call our `BlissClient` functions, so I would want to use arguments like

`("output/test.pt", 4, 4, [1])`

(Did you include one more argument here than `train_on_cached_data` expects?)

zhixiangteoh (changes requested):
In general, we want to mimic how an end user would call our api functions—i.e., such that the test functions themselves in `test_api.py`, or at least the calls to `BlissClient` methods, should be as if an end user made them (oblivious to the session `cfg` object, oblivious to anything available in our local Git checkout).

tests/test_api.py
```diff
@@ -27,145 +24,67 @@ def bliss_client(cwd, cfg):
...
+
+    def test_load_survey(self, bliss_client, monkeypatch):
+        monkeypatch.setattr(requests, "get", mock_get)
+        monkeypatch.setattr(requests, "post", mock_post)
+        monkeypatch.setattr(gzip, "decompress", lambda x: x)
+        monkeypatch.setattr(bz2, "decompress", lambda x: x)
+
+        download_dir = "data"
```
zhixiangteoh, line 72: Should this be "data/sdss" instead?

tests/test_api.py
aakashdp6548: I'm only overriding it so the correct path gets passed in the config to `mock_predict_sdss`, which uses it to load the cached predictions and return them. The alternative would be to hardcode it in `mock_predict_sdss` - not sure if that would be better

tests/test_api.py
aakashdp6548: Whoops, I thought it took both val and test indices but I guess not. None of these arguments are actually going to be used since `_train` is mocked, so I just filled in the arguments arbitrarily

tests/test_api.py
aakashdp6548: Changed to mock the simulator instead, since we already test simulate. So `test_api.test_generate` will run through all of the generate code and write (empty) files.

tests/test_api.py
aakashdp6548: Just hardcoded it for simplicity

tests/test_api.py
aakashdp6548: Ah, the extra argument was added in Sawan's multi-band PR yesterday, so it does take test indices now. I also updated the arguments and how `test_train` and `test_train_on_cached_data_work` altogether - see comment below.

tests/test_api.py
aakashdp6548: Doesn't matter what we call it, it's just a temporary directory we download to

aakashdp6548:
Lot of new changes! I restructured many of the API tests to actually test functional behavior instead of just sanity checking that the code runs. Tests now run in <2 minutes on both the CPU and GPU, with better code coverage and better test behavior! :grinning:
## Major changes:
### test_api.py
- `test_api.test_generate` now mocks `SimulatedDataset` instead of `generate`. So it now actually runs all of generation, including creating and writing files appropriately, and checks that the correct files were written.
- `test_api.test_train` mocks the trainer itself to no-op `trainer.fit` and `trainer.test`, and to load a fake checkpoint for purposes of saving the best model.
- `test_api.test_train_on_cached_data` actually does a full training run, no mocking. I removed `test_train` from `test_end_to_end` so this is the only place we actually do training now.
## Minor changes
### test_end_to_end.py
- Deleted `test_end_to_end` since train and generate are being tested elsewhere in more depth. I moved predict to its own file, `test_predict`.
### test_cached_dataset.py
- Renamed `test_cached_dataset` to `test_generate`, and removed the training test (since that's covered by `test_api.test_train_on_cached_data`) but changed the fixture to a test case since it's actually checking some good stuff.
### test_sdss.py, test_simulate.py
- Small changes in `test_sdss` and `test_simulate` to mock downloads, and use cached data in `data/tests` when possible
### test_catalogs.py
- Minor restructuring to group similar tests into classes for readability

tests/test_api.py
zhixiangteoh: I think we prefer “data/sdss” to stay consistent with whatever our repo is set up with

zhixiangteoh: Yup; nit: but I think we should prefer “data/sdss” to stay consistent with whatever our repo is set up with? But probably could just be done in a later PR that expands this test’s functionality—e.g., when API changes due to generalizing the predict function.

tests/test_sdss.py
```diff
@@ -22,6 +30,7 @@ def test_sdss(self, cfg):
         assert an_obj["gain"][3] == pytest.approx(4.76)
         assert isinstance(an_obj["wcs"], list)
 
+        # fake download 745 and just check that the size is right
```
zhixiangteoh, line 33: as per discussion, should we just make 745 available in repo / Git LFS? And remove the monkeypatch overriding code above? I think it'll just be cleaner, and we can use the 2-field dataset in future.

tests/test_api.py
```diff
@@ -1,22 +1,18 @@
...
-def cwd(tmpdir_factory):
-    return tmpdir_factory.mktemp("cwd")
+from bliss.catalog import FullCatalog
 
 
 @pytest.fixture(scope="class")
-def bliss_client(cwd, cfg):
-    client = BlissClient(str(cwd))
+def bliss_client(cfg, tmpdir_factory):
+    client = BlissClient(str(tmpdir_factory.mktemp("cwd")))
```
zhixiangteoh, line 15: Can we add a top-line comment documenting specifically, briefly, that this test does not fully check functionality (of `train`, `generate`, `predict`), but focuses on API-code-specific things, like directory structure from `cwd`?

zhixiangteoh (approved):
thanks sm! looks good (mostly for `mock_tests.py` and `test_api.py`), and will assume the other tests were just minor changes / moving files, so didn't look too closely at them.

tests/test_api.py
aakashdp6548: Done.

tests/test_sdss.py
aakashdp6548: I'm just going to remove that second test for multiple fields for now. Since the downloader is already kind of sketchy, we should update this to have more in-depth tests that also test downloader functionality when we fix that.

PR #841, opened 2023-06-26
Allow specifying config file again in generate script

aakashdp6548:
Re-added the option to specify `config-path` and `config-name` (Hydra args) for the generate script. Having this option is useful when you need to change many different parameters.

The usage is now `generate_data_in_parallel.sh -n NUM_PROCESSES [-cp CONFIG_PATH] [-cn CONFIG_NAME] [CONFIG_OVERRIDES]`. The `num-processes` flag is required, but `config-path` and `config-name` are optional - they will use the default in `api.main` if not specified. Individual overrides can still be specified positionally.

### DIFF
--- a/scripts/generate_data_in_parallel.sh
+++ b/scripts/generate_data_in_parallel.sh
@@ -13,19 +13,44 @@ cleanup() {
 # gracefully exit on these conditions
 trap "cleanup" SIGINT SIGTERM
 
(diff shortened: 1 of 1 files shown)


zhixiangteoh (approved):
use of `shift` is super clean! nice! thanks!

PR #848, opened 2023-06-28
Fix bug in table conversion from torch.squeeze

### LINKED ISSUE #838: api.fullcat_to_astropy_table can crash when there is only one source in a catalog

zhixiangteoh:
Took way too long to fix this than I should've (lol)

Test:`pytest tests/test_api.py -k "predict or table"`

Fixes #838.

### DIFF
--- a/bliss/api.py
+++ b/bliss/api.py
@@ -216,19 +216,29 @@ class BlissClient:
 
 
 def fullcat_to_astropy_table(est_cat: FullCatalog):
-    assert list(est_cat.keys()) == [
+    required_keys = [
         "star_log_fluxes",
         "star_fluxes",
         "source_type",
         "galaxy_params",
     ]
+    assert all(k in est_cat.keys() for k in required_keys), "`est_cat` missing required keys"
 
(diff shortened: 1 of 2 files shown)


jeff-regier (approved):
Nice

Issue #821, opened 2023-06-22
Return fluxes in nanomaggies instead of photon counts


aakashdp6548:
BLISS currently measures photon flux, whereas surveys like SDSS and DECaLS use nanomaggies instead. The conversion between the units depends on the catalog and the specific image, making it hard to accurately compare the predicted values to the actual catalog. As part of our pipeline that makes predictions on a specific SDSS or DECaLS frame, we should determine the right conversion and return the fluxes in the appropriate units.

For SDSS at least, we actually already have this information in `SloanDigitalSkySurvey[field]["nelec_per_nmgy_list"]`. We will need to figure out where to look in the DECaLS catalog to get the gain/calibration parameters.

Issue #772, opened 2023-05-22
infer catalogs based on DECaLS images


jeff-regier:
Available via a tap interface (supported by astroquery) from https://datalab.noirlab.edu/ls/dataAccess.php

See also https://www.legacysurvey.org/rawdata/

I believe we'd want to fetch the calibrated images.

PR #851, opened 2023-06-29
Convert fluxes to nmgy after prediction

zhixiangteoh:
Addresses #821, but only for SDSS.

### DIFF
--- a/bliss/api.py
+++ b/bliss/api.py
@@ -6,7 +6,7 @@ from typing import Dict, Literal, Optional, Tuple, TypeAlias
 import hydra
 import torch
 from astropy import units as u  # noqa: WPS347
-from astropy.table import Table
+from astropy.table import Table, hstack
 from einops import rearrange
 from omegaconf import OmegaConf
 from torch import Tensor
@@ -140,7 +140,7 @@ class BlissClient:
         self,
         weight_save_path: str,
         **kwargs,
-    ) -> Tuple[FullCatalog, Table, Table, Table]:
+    ) -> Tuple[FullCatalog, Table, Table]:
         """Predict on SDSS images.
 
         Args:
@@ -149,8 +149,9 @@ class BlissClient:
             **kwargs: Keyword arguments to override default configuration values.
 
         Returns:
-            Tuple[FullCatalog, Table, Table]: Tuple of predicted catalog, astropy.table
-            of predicted catalog, and astropy.table of predicted galaxy_params.
+            Tuple[FullCatalog, Table, Table]: Tuple of estimated catalog, estimated
+                catalog as an astropy table, and probabilistic predictions catalog as
+                an astropy table
         """
         cfg = OmegaConf.create(self.base_cfg)
         # apply overrides
@@ -160,9 +161,9 @@ class BlissClient:
 
         est_cat, _, _, _, pred = _predict_sdss(cfg)
         full_cat = est_cat.to_full_params()
-        est_cat_table, galaxy_params_table = fullcat_to_astropy_table(full_cat)
+        est_cat_table = fullcat_to_astropy_table(full_cat)
         pred_table = pred_to_astropy_table(pred)
-        return full_cat, est_cat_table, galaxy_params_table, pred_table
+        return full_cat, est_cat_table, pred_table
 
     def plot_predictions_in_notebook(self):
         """Plot predictions in notebook."""
@@ -275,9 +276,9 @@ def fullcat_to_astropy_table(est_cat: FullCatalog):
         galaxy_params_dic["galaxy_a_b"].unit = u.arcsec
         galaxy_params_list.append(galaxy_params_dic)
     galaxy_params_table = Table(galaxy_params_list)
-    est_cat_table["galaxy_params"] = galaxy_params_table
+    est_cat_table.remove_column("galaxy_params")
 
-    return est_cat_table, galaxy_params_table
+    return hstack([est_cat_table, galaxy_params_table])
 
 
 def pred_to_astropy_table(pred: Dict[str, Tensor]) -> Table:
--- a/bliss/api.py
+++ b/bliss/api.py
@@ -15,7 +15,7 @@ from bliss.catalog import FullCatalog
 from bliss.conf.igs import base_config
 from bliss.generate import generate as _generate
 from bliss.predict import predict_sdss as _predict_sdss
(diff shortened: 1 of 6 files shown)


bliss/api.py
```diff
@@ -234,20 +234,33 @@ def fullcat_to_astropy_table(est_cat: FullCatalog):
...
-    log_nmgy = (u.LogUnit(u.nmgy),)
-    est_cat_table["star_log_fluxes"] = est_cat_table["star_log_fluxes"] * log_nmgy
-    est_cat_table["star_fluxes"] = est_cat_table["star_fluxes"] * (u.nmgy,)
+    log_nmgy = u.LogUnit(u.nmgy)
+    nmgy = u.nmgy
+    for bl in "ugriz":
+        est_cat_table[f"star_log_fluxes_{bl}"].unit = log_nmgy
+        est_cat_table[f"star_fluxes_{bl}"].unit = nmgy
 
     # Create inner table for galaxy_params
```
aakashdp6548, line 256: Why do we create a separate table for galaxy parameters? Can't these just be columns in the previous table?

bliss/api.py
```diff
@@ -317,17 +330,17 @@ def pred_to_astropy_table(pred: Dict[str, Tensor]) -> Table:
...
-    pred_table["galsim_beta_radians_mean"] = pred_table["galsim_beta_radians_mean"] * u.radian
-    pred_table["galsim_beta_radians_std"] = pred_table["galsim_beta_radians_std"] * u.radian
-    pred_table["galsim_a_d_mean"] = pred_table["galsim_a_d_mean"] * u.arcsec
-    pred_table["galsim_a_d_std"] = pred_table["galsim_a_d_std"] * u.arcsec
-    pred_table["galsim_a_b_mean"] = pred_table["galsim_a_b_mean"] * u.arcsec
-    pred_table["galsim_a_b_std"] = pred_table["galsim_a_b_std"] * u.arcsec
+        pred_table[f"star_log_flux {bnd}_mean"].unit = log_nmgy
+        pred_table[f"star_log_flux {bnd}_std"].unit = log_nmgy
+        pred_table[f"galsim_flux_{bnd}_mean"].unit = u.nmgy
+        pred_table[f"galsim_flux_{bnd}_std"].unit = u.nmgy
```
aakashdp6548, line 336: Doesn't need to be done in this PR, but we should probably rename this to `galaxy_flux_*` since the galaxy flux isn't just a galsim parameter

bliss/predict.py
```diff
@@ -60,34 +60,54 @@ def align(img, sdss):
...
         ptc = cfg.encoder.tiles_to_crop * cfg.encoder.tile_slen
         cropped_image = images[0, 0, ptc:-ptc, ptc:-ptc]
         cropped_background = background[0, 0, ptc:-ptc, ptc:-ptc]
         plot_predict(cfg, cropped_image, cropped_background, sdss_plocs, est_full)
 
-    return est_cat, images[0], background[0], sdss_plocs, pred
+    return est_full, images[0], background[0], sdss_plocs, pred
+
+
+def nelec_to_nmgy(est_cat, nelec_per_nmgy_per_band):
```
aakashdp6548, line 96: Can we name this something more descriptive, like `convert_catalog_from_nelec_to_nmgy`?

bliss/predict.py
```diff
@@ -22,7 +22,7 @@ def prepare_image(x, device):
     return x  # noqa: WPS331
 
 
-def align(img, sdss):
+def align(img, ref_wcs):
```
aakashdp6548, line 25: Nice, this is better than passing in the SDSS object

aakashdp6548 (approved):
Some minor comments, but looks good overall!

bliss/api.py
zhixiangteoh: it's actually because for some reason, i cannot have an astropy column also be an (inner) astropy table.

```sh
>>> from astropy.table import Table
>>> inner_table = Table()
>>> outer_table = Table()
>>> outer_table.add_column(inner_table)
>>> outer_table
<Table length=0>
  col0 
float64
-------
>>> outer_table["col0"]
<Column name='col0' dtype='float64' length=0>

>>> outer_table.show_in_notebook()
<IPython.core.display.HTML object>
>>> outer_table["col0"].show_in_notebook()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Column' object has no attribute 'show_in_notebook'
>>> inner_table.show_in_notebook()
<IPython.core.display.HTML object>
```

zhixiangteoh: So we could either keep this two-table format, or just split `galaxy_params` into 11 individual columns and combine all into one table.

bliss/api.py
zhixiangteoh: Yea, think better addressed in a separate PR, to keep this PR's functionality to mainly handle returning fluxes in nmgy.

bliss/predict.py
zhixiangteoh: Done.

bliss/api.py
aakashdp6548: I think separate columns in one table is better. We're moving the galaxy fluxes out of `galaxy_params`, so we would need to move those to the main table anyway, and then we might as well just move the others to keep it as one clean table.

jeff-regier (approved):
Nice

jeff-regier:
I lnked this PR to #821 because it seems to address it as written. (though good to get it working for decals too, perhaps as part of #772)

bliss/encoder.py
```diff
@@ -91,8 +91,8 @@ def __init__(
     @property
     def dist_param_groups(self):
         # create a unique parameter for each per-band flux
-        star_fluxes = [f"star_log_flux {bnd}" for bnd in self.sdss_bands]
-        gal_fluxes = [f"galsim_flux_{bnd}" for bnd in self.sdss_bands]
+        star_fluxes = [f"star_log_flux {bnd}" for bnd in SloanDigitalSkySurvey.BANDS]
+        gal_fluxes = [f"galsim_flux_{bnd}" for bnd in SloanDigitalSkySurvey.BANDS]
```
aakashdp6548, line 95: Using a class constant for the bands is a nice change btw :) We should definitely start using this in other places once this gets merged in.

bliss/api.py
zhixiangteoh: Done

aakashdp6548:
Looks good!

Issue #826, opened 2023-06-23
verify colors used for multiband generation are ratios of n_electrons, not nanomaggies


jeff-regier:
The SDSS catalog contains fluxes in units of nanomaggies per band. But within BLISS, fluxes are stored in units of counts (number of electrons) per band. It may not seem like the distinction would matter since we're taking ratios of fluxes in our data generation procedure. But here's the potential issue: the number of electrons per nanomaggy is not constant across bands.

jeff-regier:
This issue is related to #821

PR #852, opened 2023-06-29
Compute SDSS light source flux ratios in units of (adjusted) electron counts

sawanp813:
This PR addresses #826 by ensuring that all light source flux values are actually kept in units of electron counts in `prior.py` before being passed to `simulator.py` for image generation. We previously kept the flux in terms of nanomaggies within BLISS, which did not account for the fact that the number of electrons per nanomaggy varies across bands. Even though we are computing ratios for sampling SDSS-like fluxes for a given source type, this error could not be ignored.

This is now accounted for within `prior.py` by reading in relevant SDSS frames and using the `n_elec_per_nmgy_list` entry within each frame to adjust our ratios to actually reflect ratios of electron counts and not only nanomaggies.


bliss/simulator/prior.py
```diff
@@ -206,10 +206,31 @@ def _load_color_distribution(self, sdss_fields):
...
+            camcol_path = sdss_path.joinpath(str(run), str(camcol))
+            pf_file = f"photoField-{run:06d}-{camcol:d}.fits"
+            pf_path = camcol_path.joinpath(pf_file)
+            if not Path(pf_path).exists():
+                self.downloader.download_pf()
+            msg = (
+                f"{pf_path} does not exist. "
+                + "Make sure data files are available for specified fields."
+            )
+            assert Path(pf_path).exists(), msg
```
aakashdp6548, line 222: Instead of manually loading the files here, it would be better to just instantiate an SDSS object (or pass one as a parameter).

bliss/simulator/prior.py
```diff
@@ -249,6 +272,45 @@ def _draw_truncated_pareto(alpha, min_x, max_x, n_samples) -> Tensor:
         uniform_samples = torch.rand(n_samples) * u_max
         return min_x / (1.0 - uniform_samples) ** (1 / alpha)
 
+    def _compute_flux_ratio(self, obj_fluxes, run, camcol, field, field_dir, gains):
```
aakashdp6548, line 275: Maybe this is fine for this PR since everything is still in terms of SDSS, but we should eventually move this to the SDSS object itself

bliss/simulator/prior.py
```diff
@@ -228,12 +249,14 @@ def _load_color_distribution(self, sdss_fields):
 
                 for obj, _ in enumerate(objc_type):
                     if objc_type[obj] == 6 and thing_id[obj] != -1 and torch.all(fluxes[obj] > 0):
-                        # flux relative to r-band
-                        ratios = fluxes[obj] / fluxes[obj][2]  # noqa: WPS220
+                        ratios = self._compute_flux_ratio(  # noqa: WPS220
+                            fluxes[obj], run, camcol, field, field_dir, gains
+                        )
```
aakashdp6548, line 254: This can be moved out of the conditional since it's the same in each branch

bliss/simulator/prior.py
```diff
@@ -249,6 +272,45 @@ def _draw_truncated_pareto(alpha, min_x, max_x, n_samples) -> Tensor:
...
+            cal_mean = calibration.mean()
+
+            # gain varies across bands
+            bnd_gain = gains[i]
+            r_gain = gains[2]  # hard-code r-gain
+
+            # distribution of nelec_per_nmgy very clustered around mean
+            nelec_per_nmgy_bnd = bnd_gain / cal_mean
+            nelec_per_nmgy_r = r_gain / cal_mean
+            nelec_per_nmgy_rat = nelec_per_nmgy_bnd / nelec_per_nmgy_r
```
aakashdp6548, line 306: Should this be the `r_gain` divided by the mean calibration of the r band, instead of `cal_mean`? Otherwise the division in line 306 simplifies to `bnd_gain / r_gain`.

bliss/simulator/prior.py
```diff
@@ -249,6 +272,45 @@ def _draw_truncated_pareto(alpha, min_x, max_x, n_samples) -> Tensor:
...
+
+            # (nelec ratio relative to r-band) * (nmgy/nelec ratio relative to r band)
```
aakashdp6548, line 308: I think the code is right, but this comment seems backward. We have (nmgy ratio relative to r), so we multiply by (nelec/nmgy ratio) right? Instead of the other way around

bliss/simulator/prior.py
```diff
@@ -228,12 +249,14 @@ def _load_color_distribution(self, sdss_fields):
```
aakashdp6548, line 249: I didn't notice this before, but are we just using the "psfflux" column to sample the ratios? IIRC "psfflux" is the modeled star fluxes - we need to look at "cmodelflux" for galaxies, and then pick depending on the object type.

aakashdp6548 (changes requested):

bliss/simulator/prior.py
sawanp813: I figured I would just load the frames individually since the instantiation of an SDSS object results in some additional overhead (mostly in `read_frame_for_band`) that I thought could be avoided since we probably want to keep everything during generation as cut and dry as possible. It definitely would have been a lot easier to just instantiate an SDSS object, but I thought it was worth the effort.

bliss/simulator/prior.py
sawanp813: Yeah "psfflux" is used for the nanomaggy count for stars and galaxies. If that's not actually the case for galaxies then that should definitely be changed.

bliss/simulator/prior.py
sawanp813: This was tricky because I suppose we don't want to compute the ratios for objects that aren't 'things' in the catalog. It's not clean at all really but I think we can just combine the logic in both conditions [i.e. if (the object is a star OR a galaxy) AND `(thing_id[obj] != -1)` AND `torch.all(fluxes[obj] > 0)`].

bliss/simulator/prior.py
sawanp813: Yes it should, thanks for the pointer!

bliss/simulator/prior.py
sawanp813: Yeah I actually had my units wrong the first time around in the first commit for this branch and had to fix it to be in units of electron counts. I forgot to revise the comment, though.

jeff-regier (approved):
Looks good. Please address Aakash's comments if possible and mark them resolved.

bliss/simulator/prior.py
aakashdp6548: You're reading the frames anyway though in `_compute_flux_ratio`. Unless there is a frame with no valid objects, I think you're doing the same amount of work either way, and imo just using the SDSS object is much cleaner than duplicating all of its functionality here (and also makes maintaining it much easier).

bliss/simulator/prior.py
aakashdp6548: Yeah, that's definitely a problem then. See how it's done in PhotoFullCatalog - we should do the same thing here.
https://github.com/prob-ml/bliss/blob/f1b7fcf23873c0e9985453b939dcd05a10b96814/bliss/surveys/sdss.py#L276-L282

bliss/simulator/prior.py
aakashdp6548: Oh, I see. I guess that makes sense. Though another problem is that `_compute_flux_ratio` is called for each source in the for loop, and it's loads the fits files for each band each time - that's really inefficient. I'm more convinced that just passing an SDSS object is the way to go, ensuring everything is only loaded once. And that will make it easier to integrate with the changes in #853, which ideally will make it so this object gets passed down from the top.

bliss/simulator/prior.py
aakashdp6548: I double checked, it does vary by band:
```
>>> sdss = SDSS("data/sdss", 94, 1, (12,))
>>> sdss[0]["calibration"].shape
(5, 2048)
```
So you would need to calculate the mean of the r-band calibration before the loop, and divide `r_gain` by that instead of `cal_mean`. Or if you use the SDSS object, you can do it in just two lines with `nelec_per_nmgy_list`: 
```
>>> mean_nelec_per_nmgy = sdss[0]["nelec_per_nmgy_list"].mean(axis=1)
>>> mean_nelec_per_nmgy
array([ 219.55634, 1039.936  ,  987.3069 ,  747.56915,  139.61816],
      dtype=float32)
>>> ratios = mean_nelec_per_nmgy / mean_nelec_per_nmgy[2]
>>> ratios
array([0.22237901, 1.0533057 , 1.        , 0.75718015, 0.14141314],
      dtype=float32)
```

bliss/simulator/prior.py
sawanp813: Yeah it would make sense if we're planning on relying on the SDSS object going forward and for maintenance. Just updated to do this in a new commit.

bliss/simulator/prior.py
sawanp813: I just did it the latter way with the SDSS object, should be all set.

bliss/simulator/prior.py
sawanp813: Added!

aakashdp6548 (approved):

Issue #850, opened 2023-06-29
filter_tile_catalog_by_flux doesn't work with multiband data


aakashdp6548:
`TileCatalog._get_fluxes_of_on_sources` now returns the flux in each band, but `filter_tile_catalog_by_flux` assumed the last dimension  would be 1.

Issue #849, opened 2023-06-29
Combine per-band flux metric state into single 5d tensor


aakashdp6548:
Instead of having one list to store tensors of flux per band, we use a single list of 5d tensors, making metrics a bit cleaner.

PR #854, opened 2023-06-30
Refactor galaxy fluxes

### LINKED ISSUE #847: Split galaxy_params into galaxy_fluxes and remaining params
Currently, we store galaxy params as a `n x nth x ntw x d` tensor (`n = batch_size`, `nth,ntw = size_in_tiles`). The first `n_bands` channels of the last dimension correspond to the galaxy fluxes in each band, and the remaining ones are the galsim parameters in a fixed order. This is tricky to work with, since the number of bands can change, changing the indices of the galsim parameters, and also it's just hard to read and understand.

We should split the galaxy fluxes into a separate item in `TileCatalog`, and leave `galaxy_params` for just the galsim parameters. This will help overall readability and also aid in generalizing to surveys with different bands. So `tile_cat["galaxy_fluxes"]` will be a `n x nth x ntw x n_bands` tensor, and `tile_cat["galaxy_params"]` will be a `n x nth x ntw x 6` tensor (6 galsim parameters).

Code changes needed:
- `prior.sample_galaxy_prior` should return separate tensors for flux and params
- `prior.sample_prior` should return galaxy_fluxes as separate key
- `encoder.variational_mode` should keep galaxy fluxes as a separate key, instead of concatenating onto galaxy params
- `encoder._get_loss` needs to compute loss on galaxy fluxes separately
- `metrics` needs to update indices for getting fluxes and params

Other changes:
- This will require regenerating all of the current data. This is probably the best time for it though, since we'll have to regenerate anyway once #826 is fixed.

aakashdp6548:
Fixes #847, #849, and #850. **This PR breaks compatibility with old data. See the data conversion section below for details on converting existing data.**

### Major changes:
- Galaxy fluxes are now stored in `catalog["galaxy_fluxes"]`, instead of as part of `galaxy_params`. This refactoring doesn't really change any functional behavior, just how we store the data and how we compute loss/metrics. `generate` will automatically generate data in this format.
- The galaxy fluxes are now in the main table returned by `BlissClient.full_cat_to_astropy_table`. The galaxy params table only contains the 6 galaxy params (though I think this will be folded into the main table too in #851?)
- `TileCatalog.get_brightest_source_per_tile` and `TileCatalog.get_brightest_source_per_tile` now need a band argument to select the band to use for comparison, since "brightest" flux is ambiguous with multiple bands.
- Fixed compatibility between multiband and psf inputs. Now `Encoder.get_input_tensor` indexes deconvolution and psf parameters by bands correctly.
- `Encoder._get_loss` now only computes flux loss over bands that were specified. If you're deliberately using a single band instead of five, there's probably a good reason and there's not really a point to penalizing the model for mispredicting things you chose not to show it.

### Minor changes:
- Metrics keeps a single list each for galaxy and star fluxes, instead of one per band per type. This is just code cleanup, it doesn't affect the computation or the values returned by metrics.
- Some minor renaming, like `galsim_flux_*` -> `galaxy_flux_*` in astropy tables, and moving the flux and galsim parameter names to be class constants in `Encoder`.
- Updated tests for filtering catalogs to use multiband data.

### Data conversion
To avoid having to regenerate old data, I added a script to convert old data to the new format. It's under `/data/scratch/aakash/convert_data.py`; I didn't add it to the repo since once all of the old data is converted we shouldn't need it ever again. The usage is `python convert_data.py -d <data_directory>`.


jeff-regier (approved):
Looks good

PR #857, opened 2023-07-03
Updated PSF case study

aakashdp6548:
Updated with new models to support multi-band encoder changes. Also added galaxy param metrics to notebook.


jeff-regier (approved):
LGTM

aakashdp6548:
@zhixiangteoh Could you take a look at the most recent commit ([4fbd99c](https://github.com/prob-ml/bliss/pull/857/commits/4fbd99c21c447c145457c3567235ba87e610ccee))? I reverted the compatibility change in `encoder` since the old models still don't work, and replaced `single_field.pt` in `test_predict` with the new single-band model trained with the current architecture.

PR #858, opened 2023-07-03
Miscellaneous fixes

aakashdp6548:
A couple of minor fixes that have been bugging me for a while :grin: 
- Cleaned up the main `train` function (refactored setting up paths and saving best model into separate function, fixed bug where we try to save best model even when no checkpoint exists, made `config.training.test` a boolean value instead of a useless path, fixed profiler file name)
- Fixed bug with `bulge_frac` metrics computation. In theory we don't need this at all since the math works out so that `bulge_frac_mae = disk_frac_mae`, but that's not really obvious so we might as well keep it.
- Updated `test_metrics` to use cached tile catalog instead of generating on the fly


zhixiangteoh (approved):
nice clean up to make train function more readable.

PR #865, opened 2023-07-05
Training with multiple sources per tile

### LINKED ISSUE #823: generate multiple sources per tile
In our synthetic training data, currently each tile has at most one light source. But in real data, the number of light sources per tile is better modeled as following a Poisson distribution. To make our simulated data more realistic, lets set `simulator.prior.max_sources` to 2, decrease `simulator.prior.star_flux_min` and `simulator.prior.galaxy_flux_min` so that undetectable light sources are occasionally generated, and then use the new catalog filtering routines to create a `target_catalog` for training that has at most one light source per tile and only contains detectable light sources (i.e., flux greater than some value).

aakashdp6548:
Closes #823. This PR enables training on data with multiple sources per tile. We filter by the brightest source in each tile, followed by thresholding fluxes above some minimum detectable level.

Minor changes:
- minor fix to squeezing/unsqueezing in `TileCatalog.filter_catalog_by_flux` to handle case where `batch_size=1`
- Added `validate_args=False` to `UnconstrainedLogNormal` to allow evaluating log prob at 0 (for tiles that were filtered out)

image


jeff-regier (approved):
Looks good!

Issue #866, opened 2023-07-06
Model star fluxes in the variational distribution as LogNormal


aakashdp6548:
We currently model `star_log_fluxes` as a `Normal` distribution, and `galaxy_fluxes` as a `LogNormal`. We should make these consistent by changing to model `star_fluxes` as a `LogNormal` as well.

Issue #864, opened 2023-07-05
FullCatalog.plot_plocs belongs in plotting.py


aakashdp6548:
It's only ever called in `plotting.py` anyway and we should keep plotting code together for overall code cleanliness.

PR #869, opened 2023-07-07
Miscellaneous fixes

### LINKED ISSUE #863: Add plocs to fullcat_to_astropy_table
`TileCatalog` and `FullCatalog` store `locs/plocs` and `n_sources` as attributes instead of keys in the `UserDict`. This is confusing and can lead to them being missed, like in the table returned by `fullcat_to_astropy_table`. We should change this to be consistent with the other keys. This will likely need special logic to handle `n_sources` in places that mask with `TileCatalog.is_on_mask` or `FullCatalog.get_is_on_mask()`, but this still seems cleaner than the current method.

aakashdp6548:
Fixes #863, #864, #866, and #868.

Main changes:
- We now model the star fluxes in the variational distribution as a `LogNormal`, instead of modeling the log fluxes as a `Normal` distribution. This is consistent with galaxy fluxes, which are modeled as `LogNormal`.
- `plot_plocs` has been moved to `plotting.py`, and now accepts either a string or a list of indices to determine which sources to plot
- Added `plocs` to the astropy table returned in `api.fullcat_to_astropy_table`. This was missed initially since `plocs` isn't in the dictionary, so isn't found in `catalog.items()`.

Minor changes:
- Added tests for using the deconvolution/psf params to encode the psf, and for encoding a batch with multiple sources per tile.
- Modified some tests to use `encoder` and `decoder` fixtures

### DIFF
--- a/bliss/api.py
+++ b/bliss/api.py
@@ -218,7 +218,6 @@ class BlissClient:
 
 def fullcat_to_astropy_table(est_cat: FullCatalog):
     required_keys = [
-        "star_log_fluxes",
         "star_fluxes",
         "source_type",
         "galaxy_params",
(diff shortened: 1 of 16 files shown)


jeff-regier (approved):
Nice!

zhixiangteoh (approved):

PR #870, opened 2023-07-07
Updated catalog comparison case study to better visualize predictions

aakashdp6548:
Added a new function `metrics.three_way_matching` to better compare three catalogs, and added new visualization of BLISS vs SDSS vs DECaLS with three-way metrics
image


jeff-regier (approved):
LGTM

Issue #859, opened 2023-07-03
predict_decals doesn't work with plotting


aakashdp6548:
`predict_decals` fails when `config.predict.plot.show_plot = True`. It looks like there's two things going on:
1) `trainer.predict` returns a tile catalog, but  `plot_predict` expects a full catalog. Should just be converted in between like in `predict_sdss`.
2) `decals_plocs_from_decals` seems to return an empty tensor, which is causing problems in `predict.crop_plocs`.

Our tests right now are run without plotting enabled in the config (line 21 was added manually below) - maybe we should enable it to catch this kind of thing in the future?

![image](https://github.com/prob-ml/bliss/assets/26699876/415c49bc-8e41-4307-b835-b70bca4b3ceb)

zhixiangteoh:
Yes, I intentionally set to false in the test I wrote in `test_predict.py` now because I’m punting this to after #853.

Issue #871, opened 2023-07-10
DECaLS end-to-end - generate, train, predict


zhixiangteoh:
Predict (mostly) implemented in #772.

Generate:
- PSFs (using https://galsim-developers.github.io/GalSim/_build/html/des.html#des-psf-models)
- Simulate co-added images (#732)

zhixiangteoh:
generate - #732  
training - #860 
predict - #770, #859

PR #874, opened 2023-07-12
Make predictions on multiple image units

### LINKED ISSUE #770: bulk predict
predict should allow inferring catalogs for a large set of images (from one survey)

zhixiangteoh:
Allows making predictions on multiple `image_id`s (e.g., RCF, brick)

WIP:
- Tests (`test_sdss_reconstruct.py`)
- Predictions plots per image unit

Will fix #770, #859.

### DIFF
--- a/bliss/api.py
+++ b/bliss/api.py
@@ -1,7 +1,7 @@
 import logging
 from os import environ, getenv
 from pathlib import Path
-from typing import Dict, Literal, Optional, Tuple, TypeAlias
+from typing import Any, Dict, Literal, Optional, Tuple, TypeAlias
 
 import hydra
 import torch
@@ -26,8 +26,8 @@ class BlissClient:
     """Client for interacting with the BLISS API."""
 
     def __init__(self, cwd: str):
-        self._cwd = cwd
         self.base_cfg = base_config()
+        self._cwd = cwd
         self.base_cfg.paths.root = self.cwd
 
     def generate(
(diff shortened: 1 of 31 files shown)


bliss/api.py
```diff
@@ -133,14 +133,14 @@ def train_on_cached_data(
...
-        SDSSDownloader(run, camcol, field, self.cwd + f"/{download_dir}").download_all()
+        SDSSDownloader([(run, camcol, field)], self.cwd + f"/{download_dir}").download_all()
         # assert files downloaded at download_dir
 
     def predict_sdss(
         self,
         weight_save_path: str,
         **kwargs,
-    ) -> Tuple[FullCatalog, Table, Table]:
+    ) -> Tuple[FullCatalog, Table, Dict[Any, Table]]:
```
jeff-regier, line 143: Is `image_id` always an integer? If so, can we get rid of `Any` here?

jeff-regier (commented):
Looks good so far!

bliss/api.py
zhixiangteoh: Not always an integer. For example, for SDSS it is `Tuple[int, int, int]` for `(run, camcol, field)`; for DECaLS it is `str` for `brickname`.

zhixiangteoh:
@XinyueLi1012 ptal and lmk if anything seems off, especially regarding the changes to predictions that touch code you've written

bliss/api.py
```diff
@@ -149,19 +149,52 @@ def predict_sdss(
...
+        pred_tables = {}  # indexed by image_id
+        for image_id, pred in pred_for_image_id.items():
+            pred_tables[image_id] = pred_to_astropy_table(pred)
+        return est_cat, est_cat_table, pred_tables
+
+    def predict_decals(
+        self,
+        weight_save_path: str,
+        **kwargs,
+    ) -> Tuple[FullCatalog, Table, Dict[Any, Table]]:
```
jeff-regier, line 172: In api.py, can we merge `predict_sdss` and `predict_decals` into a single `predict` function? They look pretty similar.

jeff-regier: I suppose it's kind of beyond the scope of this PR though. Sometime to work on as part of #871 ? It'd be good to merge all this before we get merge conflicts.

jeff-regier (approved):

Issue #862, opened 2023-07-05
GMM color prior


jeff-regier:
- [x] fit GMM prior for SDSS log-nmgy-ratios
- [x] specify SDSS reference band prior in units of nanomaggies
- [x] `ImageDecoder` constructor to take calibration information as an argument (along with PSF, which it already takes)
- [ ] remove `SDSSPrior` class (prior shouldn't be unique to any one dataset)
- [x] rename `ImagePrior` to `CatalogPrior`

PR #875, opened 2023-07-12
GMM Color Prior (SDSS support only)

sawanp813:
This PR addresses #862 by adding a pair of Gaussian Mixture Models trained on star flux and galaxy flux data collected across several runs in SDSS. Both models were trained offline, as described in `bliss/case_studies/color_models/load_ratios.ipynb` and are used during generation to sample novel flux ratios in the `SDSSPrior` class. These samples are roughly of the same distribution as the true SDSS flux ratios are, judged visually and empirically following a cross-validation fitting procedure.

Additions:
- GMMs fit to nmgy ratios in log-space
- Changes to r-band pareto distribution parameters, now in nmgy units
- Conversion from nmgy units to nelec counts in `decoder.py` via `nmgy_to_nelec_dict`
- Refactoring to pass `survey` into `prior`


bliss/simulator/prior.py
```diff
@@ -222,7 +205,7 @@ def multiflux(self, flux_ratios, flux_base) -> Tensor:
 
         return total_flux
 
-    def _flux_ratios_against_b(self, survey_data_dir, image_ids, items, b) -> Tuple[dict, dict]:
+    def _flux_ratios_against_b(self, survey_data_dir) -> Tuple[dict, dict]:
         """Sample and compute all star, galaxy fluxes relative to `b`-band based on real image data.
 
         Instead of pareto-sampling fluxes for each band, we pareto-sample `b`-band flux values,
```
jeff-regier, line 211: This comment may be out-of-date as there's no `b` in scope

jeff-regier: A new function name maybe helpful too now that `b` isn't an argument

bliss/surveys/sdss.py
```diff
@@ -507,63 +508,44 @@ def __init__(
...
     ):
         super().__init__(bands, **prior_config)
-        self.stars_fluxes, self.gals_fluxes = self._flux_ratios_against_b(
-            survey_data_dir, image_ids, image_items, b=ref_band
-        )
-
-    def _flux_ratios_against_b(self, survey_data_dir, image_ids, items, b) -> Tuple[dict, dict]:
-        stars_fluxes = {}
-        gals_fluxes = {}
+        self.stars_fluxes, self.gals_fluxes = self._flux_ratios_against_b(survey_data_dir, ref_band)
```
jeff-regier, line 511: Rather than drawing samples in the constructor, could the prior distribution have a function called `sample(...)`, taking `num_samples` as an argument?

jeff-regier: Like, it makes sense for the constructor to load the `pkl` files. But you wouldn't want to redo it every time you generate a batch. You'd want to do it once, keep the `SDSSPrior` object in memory, and reuse it every time you generate a batch. Is that what's happening currently?

bliss/simulator/prior.py
```diff
@@ -193,11 +176,11 @@ def _draw_truncated_pareto(self, alpha, min_x, max_x, n_samples) -> Tensor:
         return min_x / (1.0 - uniform_samples) ** (1 / alpha)
 
     def _sample_star_fluxes(self, star_ratios):
-        latent_dims = (self.batch_size, self.n_tiles_h, self.n_tiles_w, self.max_sources)
+        latent_dims = (self.batch_size, self.n_tiles_h, self.n_tiles_w, self.max_sources, 1)
         b_flux = self._draw_truncated_pareto(
             self.star_flux_alpha, self.star_flux_min, self.star_flux_max, latent_dims
         )
-        total_flux = self.multiflux(star_ratios, b_flux)
```
aakashdp6548, line 200: Can we remove the `multiflux` function completely? It looks like it isn't being used anywhere anymore.

bliss/simulator/prior.py
sawanp813: I actually have b as an argument in `SDSSPrior` so I just added it back. Think all of the '{survey}Prior' will need that argument.

bliss/simulator/prior.py
```diff
@@ -134,25 +133,9 @@ def sample_prior(self, batch_image_ids) -> TileCatalog:
...
-                self.stars_fluxes[image_id]
-                if image_id in self.stars_fluxes
-                else random.choice(list(self.stars_fluxes.values()))
-            )
-
-        galaxy_fluxes, galaxy_params = self._sample_galaxy_prior(select_gal_flux_ratios)
-        star_fluxes = self._sample_star_fluxes(select_star_flux_ratios)
+
+        galaxy_fluxes, galaxy_params = self._sample_galaxy_prior(self.gals_fluxes)
+        star_fluxes = self._sample_star_fluxes(self.stars_fluxes)
```
aakashdp6548, line 138: I don't really like this because `self.gals_fluxes` and `self.stars_fluxes` aren't attributes in `ImagePrior`, they're only in the subclasses. That can easily break if a subclass decides to implement/store ratios in a different way.

Imo it would be cleaner to have an abstract `get_star_ratios` function with a `n_samples` parameter (same for galaxies). Then `_sample_galaxy_prior` and `_sample_star_fluxes` can just call those respective functions and get the ratios themselves instead of having to pass them in.

bliss/surveys/sdss.py
aakashdp6548: ^ I agree with that overall approach. But I think the current code doesn't even redo it each batch - `ImagePrior` only gets instantiated once, and `_flux_ratios_against_b` only samples `batch_size` ratios. So I think each batch is going to use the same set of ratios instead of resampling each time.

bliss/simulator/prior.py
aakashdp6548: The general use of this function is similar to what we want the `subclass.sample` functions to be. But I think it would be better to just get rid of this and add abstract `_sample_star_ratios` and `_sample_galaxy_ratios` as mentioned in my other comment.

aakashdp6548 (changes requested):
A couple of comments regarding the general flow of flux sampling.

Also, it looks like the .pkl files haven't been committed - was that intentional? And when you do add them, can you check how large the files are? Pickle doesn't compress files by default, so if it feels large maybe we should compress them manually with gzip or something.

bliss/simulator/prior.py
sawanp813: Actually taking this back ^, `b` shouldn't be an argument for this method since it's survey-specific.

bliss/surveys/sdss.py
sawanp813: It does not redo the ratios every batch - fixed!

sawanp813:
This PR now includes the addition of two color model GMMs (stars, galaxies) that are fit to nmgy ratios relative to the r-band in log-space. We load these models once in the constructor for `ImagePrior` and sample from them during the creation of the tile catalog for each batch. 

Since these models are fit to nmgy ratios, we convert the nmgy fluxes contained in the tile catalog to electron count fluxes within `decoder.py` using an `image_id`-indexed dictionary `nmgy_to_elect_dict`. This dictionary is passed down from the inheritor of the `survey` class (e.g. `SloanDigitalSkySurvey`). I was thinking of making this an abstract method within `survey.py` to be implemented across all new surveys we choose to add but held off on it for now. 

Additionally, we now sample the r-band flux in `prior.py` in units of nmgys to be compatible with the above changes. This resulted in changes to `base_config.yaml`. 

This branch fails `test_simulate.py` because this test case loads a cached tile catalog and renders an image based solely on this catalog. This tile catalog has fluxes in units of electron counts already so the conversion that is done within `decoder.py` results in meaningless fluxes. The cached tile catalog that is loaded in this test case has to be regenerated so that its fluxes are in units of nmgys prior to rendering. All other test cases pass!

bliss/surveys/sdss.py
```diff
@@ -179,6 +178,14 @@ def get_from_disk(self, idx):
         ret.update({"field": field})
         return ret
 
+    def nmgy_to_nelec(self):
```
sawanp813, line 181: Think this should be an abstract method for all survey objects.

bliss/simulator/prior.py
```diff
@@ -186,47 +193,52 @@ def _sample_star_fluxes(self, star_ratios):
...
+        samples = (self.batch_size, self.n_tiles_h, self.n_tiles_w, self.max_sources)
+        star_ratios_flat = self.gmm_star.sample(np.prod(samples))[0]
+        gal_ratios_flat = self.gmm_gal.sample(np.prod(samples))[0]
+
+        # Reshape drawn values into appropriate form
+        samples = samples + (self.n_bands - 1,)
+        star_ratios_nmgy = np.exp(np.reshape(star_ratios_flat, samples))
+        gal_ratios_nmgy = np.exp(np.reshape(gal_ratios_flat, samples))
+
+        hardcoded_ratios = np.array([219, 1039, 747, 139])
```
aakashdp6548, line 230: What is this here for?

bliss/conf/base_config.yaml
```diff
@@ -161,16 +161,18 @@ surveys:
             mean_sources: 0.2
             min_sources: 0
             prob_galaxy: 0.72
-            star_flux_min: 622
-            star_flux_max: 1e6
+            star_flux_min: 0.63094948
+            star_flux_max: 1014
```
aakashdp6548, line 165: Instead of making these values based on the conversion from 94/1/12, I think it would be better to just use some general values like 0.6 - 1500 or something

aakashdp6548 (approved):
Looks good, just one small comment and a question

aakashdp6548:
For the `test_simulate` cached data, you should be able to convert it by just dividing the flux tensors by 987 (since I think they were all generated from 94/1/12). 
```
data = torch.load("data/tests/test_image/dataset_0.pt")
data["star_fluxes"]  = data["star_fluxes"] / 987
data["galaxy_fluxes"]  = data["galaxy_fluxes"] / 987
torch.save(data, "data/tests/test_image/dataset_0.pt")
```

bliss/conf/base_config.yaml
sawanp813: These are actually from 94, 125, 752 and 3900. I used the same values that we had been using (in units of electron counts) and just divided by the average r_band ratio across all of the above runs.

bliss/simulator/prior.py
sawanp813: forgot to delete that - it was a quick fix when I didn't have the `nmgy_to_nelec_dict` available.

sawanp813:
> For the `test_simulate` cached data, you should be able to convert it by just dividing the flux tensors by 987 (since I think they were all generated from 94/1/12).
> 
> ```
> data = torch.load("data/tests/test_image/dataset_0.pt")
> data["star_fluxes"]  = data["star_fluxes"] / 987
> data["galaxy_fluxes"]  = data["galaxy_fluxes"] / 987
> torch.save(data, "data/tests/test_image/dataset_0.pt")
> ```
> 
> I think it would be good to get it fixed as part of this PR.

Think I would have to divide by nelec to nmgy ratio for all five bands right? Which isn't just 987? Think it would be [219, 1040, 987, 748, 140].

bliss/conf/base_config.yaml
jeff-regier: Either way seems OK to me. Maybe better to keep them exact for now so that if our results change following the PR, we'll know it wasn't because of rounding these values.

bliss/surveys/sdss.py
```diff
@@ -503,67 +510,10 @@ def _psf_fun(self, r, sigma1, sigma2, sigmap, beta, b, p0):
...
-                        gal_fluxes_ratios.append(ratios)  # noqa: WPS220
-
-            if star_fluxes_ratios:
-                stars_fluxes[(run, camcol, field)] = star_fluxes_ratios
-            if gal_fluxes_ratios:
-                gals_fluxes[(run, camcol, field)] = gal_fluxes_ratios
-
-        return stars_fluxes, gals_fluxes
+        self.survey_data_dir = survey_data_dir
+        self.reference_band = reference_band
```
jeff-regier, line 516: Can we delete (what little is left of) this class?

bliss/simulator/prior.py
```diff
@@ -121,11 +126,12 @@ def __init__(
...
 
-    def sample_prior(self, batch_image_ids) -> TileCatalog:
-        """Samples latent variables from the prior of an astronomical image.
+        self.survey_data_dir = survey_data_dir
+        self.b_band = reference_band
+        self.gmm_star, self.gmm_gal = self._load_color_models()
 
-        Args:
-            batch_image_ids: image_ids needed for sampling.
+    def sample_prior(self) -> TileCatalog:
```
jeff-regier, line 133: Could we call this function `sample` rather than `sample_prior`, for consistency with the PyTorch distributions?

sawanp813:
There are more underlying issues with `test_simulate.py` and the cached tile catalog that gets loaded from disk that I haven't figured out yet by e.o.d.. I think I'm just going to regenerate the tile catalog entirely.

sawanp813:
> There are more underlying issues with `test_simulate.py` and the cached tile catalog that gets loaded from disk that I haven't figured out yet by e.o.d.. I think I'm just going to regenerate the tile catalog entirely.

Think the issues here are because the pretrained multiband model isn't trained on the GMM-generated fluxes so the flux predictions are worse than before. Once this model is updated with a newer version, we can revert back to more reasonable numbers in the `test_simulate.py` case.

bliss/conf/base_config.yaml
```diff
@@ -142,35 +165,16 @@ predict:
...
-            star_flux_max: 1e6
-            star_flux_alpha: 0.43
-            galaxy_flux_min: 622.0
-            galaxy_flux_max: 1e6
-            galaxy_alpha: 0.47
-            galaxy_a_concentration: 0.39330758068481686
-            galaxy_a_loc: 0.8371888967872619
-            galaxy_a_scale: 4.432725319432478
-            galaxy_a_bd_ratio: 2.0
+        max_bands: 5
```
jeff-regier, line 177: Will you change `max_bands` to `n_bands` (because it's just a constant property of the survey) and move `reference_band` to the prior (because it's not really a property of the survey--it's our model)?

bliss/conf/base_config.yaml
```diff
@@ -27,6 +27,29 @@ simulator:
     num_workers: 32
     valid_n_batches: 10  # 256
     fix_validation_set: true
+    prior:
+        _target_: bliss.simulator.prior.CatalogPrior
+        reference_band: ${simulator.survey.reference_band}
+        max_bands: ${simulator.survey.max_bands}
+        color_model_dir: ${simulator.survey.dir_path}/color_models
```
jeff-regier, line 34: as discussed, rather than a `color_model_dir`, we'll specify file paths to a star color model and a galaxy color model

jeff-regier (approved):
Great 👍

Issue #829, opened 2023-06-23
image transformations for improved encoding


jeff-regier:
Evaluate whether/which transformations of images improve encoder performance. Some transformations to consider:

- [x] log (with various offsets and rescalings)
- [x] sigmoid (with various offsets and rescalings)
- [x] stacking/coaddition (across bands)
- [x] z-scoring (per pixel, at perhaps other ways too)
- [ ] compositions of the above transformations

For all these, [Hydra multi-run](https://hydra.cc/docs/tutorials/basic/running_your_app/multi-run/) should be helpful. We'd like to keep a record or what did/didn't work well.

Issue #811, opened 2023-06-20
Incorrect flux predictions for very bright sources (multi-band)


sawanp813:
The multi-band BLISS model, currently requiring a z-score transform of the input images to the encoder, achieves desirable results for detecting light sources. However, there is a degree of degraded performance when it comes to estimating the flux parameter of  a light source, particularly those which are much brighter than the majority of the distribution of light sources. For instance, in `test_sdss_reconstruct`, we relax the assertion for residual deviation of pixel intensities in a reconstructed image to 7 deviations, as opposed to 5 deviations in the single-band model `(assert ((res_img.abs() / recon_img.sqrt()) > 7).sum() == 0).` From an initial inspection, this is presumably due to the presence of a small subset of very bright light sources atypical of the distribution in SDSS images. 

1. Investigate this issue further, perhaps with synthetic images, and verify that the issue is indeed present when encoding images containing very bright light sources. 
2. Consider and test additional transformations that may circumvent this issue to allow correct encoding of extremely bright sources (e.g. sigmoid transform of only extremely bright sources located in image as an extra channel).

PR #878, opened 2023-07-18
#829 input transforms

sawanp813:
This PR tracks #829 and #811 to find a combination of input transformations that optimizes BLISS performance across any form of semi-synthetic SDSS image, particularly with a focus on images containing bright light sources.


tests/test_sdss_reconstruct.py
```diff
@@ -32,5 +32,5 @@ def test_sdss_reconstruct(self, cfg, decoder):
         flux_diff = res_bright.abs().sum()
```
sawanp813, line 32: Have to update this file to get the correct output from `mock_predict` - it should fail for this commit.

bliss/conf/base_config.yaml
```diff
@@ -81,8 +81,10 @@ encoder:
     input_transform_params:
         use_deconv_channel: false
         concat_psf_params: false
-        log_transform: true
+        log_transform: false
+        rolling_z_score: true
         z_score: false
+        clahe: false
```
jeff-regier, line 87: Can we remove clahe and z_score, if rolling z score is all we need? (I'd like to keep our software simple to use by minimizing the number of configuration options, even at the expense of some flexibility.)

bliss/transforms.py
```diff
@@ -21,3 +67,8 @@ def pixelwise_norm_source(img):
 def log_transform(img):
     """Perform pixel-wise log transformation of full image."""
     return torch.log(img)
+
+
+def tanh_mod(img, scale, shift):
```
jeff-regier, line 72: Did this help? If not, probably better to remove it. Better not to keep it around "just in case" because it's extra code to maintain.

sawanp813: Nope, we should remove that.

tests/test_sdss_reconstruct.py
sawanp813: (This is all set now)

bliss/conf/base_config.yaml
sawanp813: Would it be worth keeping the `z_score` config option around for the multiband paper? I'm assuming we'll want to talk about the global z-scoring procedure a bit.

jeff-regier: Either way. I wouldn't really call the z_scoring procedure global though -- it just operates on 80x80 the subimages that we use for training.

sawanp813: The way that `z_scoring` procedure works though within the `transforms.py` code is that it takes the mean and std over the entire image, not just 80x80 patches. On real data it computes those values over the whole image.

jeff-regier: That's more a defect though than a feature. It makes it so the training data the prediction data are from a different in distribution.

jeff-regier: Thus my lack of enthusiasm for the class z_scoring procedure.

sawanp813: So if we have to re-implement that z_scoring procedure, we should probably just get rid of it here then.

bliss/transforms.py
sawanp813: By 'this', what functions are you referring to exactly? Just the hyperbolic tangent?

jeff-regier: Yep

jeff-regier (approved):

PR #882, opened 2023-07-18
#836 3D Convolution

sawanp813:
This PR tracks #836 which will allow the grouping of all band-specific channels together, preventing mismatching channels (which is currently the case in BLISS).


bliss/base_detect/base_common.py
```diff
@@ -0,0 +1,182 @@
...
+def autopad(k, p=None, d=1):  # kernel, padding, dilation
+    # Pad to 'same' shape outputs
+    if d > 1:
+        k = (
+            d * (k - 1) + 1 if isinstance(k, int) else [d * (x - 1) + 1 for x in k]
+        )  # actual kernel-size
+    if p is None:
+        p = k // 2 if isinstance(k, int) else [x // 2 for x in k]  # auto-pad
+    return p
+
```
jeff-regier, line 15: Seems like there must be a better way than including all this yolo code in our repo. Is it possible to mock/monkeypatch the yolo module, to somehow make it use our `Conv3d` class?

jeff-regier: Or can we get away with only including a customized version of `parse_model` in our repo, rather than all this yolo code?

sawanp813: I think overwriting the yolo `parse_model` could be a promising idea - I was planning on rewriting that function a decent amount anyways since it's a little difficult to inject a dimension `squeeze` layer given how it is currently written (i.e. going from the 5-D 'head' back to 4-D). I don't think we'd have to change anything else besides that and including our own `Conv3d` class. 

Cleaning/handling all of the yolo code ourselves is definitely a lot (as shown by the several `#noqa` and `pylint: disable` statements).

sawanp813: I think it's a good idea to get the 3D conv working end-to-end with the extra YOLO code first, seeing what we'd specifically have to keep and what we can get away with, and then I can cut down on the copied code once it's running. But I'm pretty sure that only the two aforementioned things are really necessary for our purposes as you mentioned.

bliss/base_detect/base_detect.py
```diff
@@ -0,0 +1,415 @@
...
+        # Build strides, anchors
+        m = self.model[-1]  # Detect()
+        if isinstance(m, (Detect)):
+            s = 256  # 2x min stride
+            m.inplace = self.inplace
+            forward = lambda x: self.forward(x)  # pylint: disable=C3001, W0108  # noqa: E731
+
+            # NOTE: 5-D input
+            d = 2  # set d = 2 in 5-band case
+            m.stride = torch.tensor([s / x.shape[-2] for x in forward(torch.zeros(1, ch, d, s, s))])
```
sawanp813, line 237: This model stride `(m.stride)` had to be overwritten within the constructor of `DetectionModel` - it was originally passing a 4-D zero-tensor of size `(1 x ch x s x s)` to set the stride. Other than this, all other changes that would enable us to use a 5-D input are within `parse_model` and within our `base_config.yaml`. 

If we import `DetectionModel` from yolo, it would be passing in this 4-D zero-tensor input to our 3-D convolution head, which would throw an error during construction. Is there some clever way to get around this without having to add the constructor to our codebase?

bliss/backbone.py
```diff
@@ -0,0 +1,135 @@
...
+    no = na * (nc + 5)  # number of outputs = anchors * (classes + 5)
+
+    layers, save, c2 = [], [], ch[-1]  # layers, savelist, ch out
+    for i, (f, n, m, args) in enumerate(d["backbone"] + d["head"]):  # from, number, module, args
+        m = eval(m) if isinstance(m, str) else m  # eval strings  # noqa: S307, WPS421
+        for j, a in enumerate(args):
+            with contextlib.suppress(NameError):
+                args[j] = eval(a) if isinstance(a, str) else a  # eval strings  # noqa: S307, WPS421
+
+        n = (  # noqa: WPS429, WPS120, F841 # pylint: disable=W0612
```
jeff-regier, line 93: Would you disable pylint and flake8 for the whole file rather than going line-by-line, since there are so many violoations? And perhaps add a comment to the top of the file explaining that this isn't our code, it's yolo5v code, but we're including it so we can use conv3d.

bliss/base_detect/base_common.py
```diff
@@ -0,0 +1,182 @@
...
+        return self.act(self.conv(x))
+
+
+class Conv3d_down(Conv3d):  # noqa: N801
+    def forward(self, x):
+        z = self.act(self.bn(self.conv(x)))
+        return torch.squeeze(z, 1)
+
+
+class Bottleneck(torch.nn.Module):
```
jeff-regier, line 98: Why do we need to include `Bottleneck` in our repo, rather than using the version in yolov5?

jeff-regier: Same for some of the other classes below.

sawanp813: Everything in `base_detect` is outdated. We can delete that entire directory now

jeff-regier (changes requested):
Did some changes not get included here?

bliss/encoder.py
```diff
@@ -177,8 +183,10 @@ def encode_batch(self, batch):
         # give us output of the right dimension
         self.model.model[-1].training = True
 
-        assert inputs.size(2) % 16 == 0, "image dims must be multiples of 16"
         assert inputs.size(3) % 16 == 0, "image dims must be multiples of 16"
+        assert inputs.size(4) % 16 == 0, "image dims must be multiples of 16"
+        bn_warn = "batchnorm training requires a larger batch. did you mean to use eval mode?"
+        assert (not self.training) or batch["images"].size(0) > 4, bn_warn
```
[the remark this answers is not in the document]
jeff-regier: @XinyueLi1012 deleted the assertion on line 189 for the encoder to work with dc2 a few days ago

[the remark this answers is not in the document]
sawanp813: I think the image dimensions should be fine with not being multiples of 16. The only thing that can't change within the architecture is that the number of output channels have to be a multiple of 8. I can remove these lines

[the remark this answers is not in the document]
jeff-regier: I was just referring to line 189, which is about the batch size (not the image dimensions)

bliss/backbone.py
```diff
@@ -0,0 +1,135 @@
...
+        for mi, s in zip(m.m, m.stride):  # from
+            b = mi.bias.view(m.na, -1)  # conv.bias(255) to (3,85)
+            b.data[:, 4] += math.log(8 / (640 / s) ** 2)  # obj (8 objects per 640 image)
+            b.data[:, 5 : 5 + m.nc] += (
+                math.log(0.6 / (m.nc - 0.99999)) if cf is None else torch.log(cf / cf.sum())
+            )  # cls
+            mi.bias = torch.nn.Parameter(b.view(-1), requires_grad=True)
+
+
+def parse_model(d, ch, n_imgs):  # model_dict, input_channels(3)
```
jeff-regier, line 69: Please add a comment explaining that this isn't our code in this file: it's a modified version of code from yolov5. (We want to give them credit.)

jeff-regier: It also would be helpful to future developers to explain why we're including yolov5 code here

bliss/conf/base_config.yaml
```diff
@@ -102,26 +105,26 @@ encoder:
...
-    max_images_per_file: 4096
+    max_images_per_file: 2048
     cached_data_path: ${paths.data}/cached_dataset
     file_prefix: dataset
 
 training:
-    name: null
-    version: null
+    name: multi-band
+    version: testing
```
jeff-regier, line 127: better to leave these two lines unmodified?

bliss/conf/base_config.yaml
```diff
@@ -102,26 +105,26 @@ encoder:
...
-            [[17], 1, Detect, [nc, anchors]],
+            [[-1], 1, Detect, [nc, anchors]],
         ]
 
 generate:
     n_workers_per_process: 0
     n_batches: ${simulator.n_batches}
     batch_size: ${simulator.prior.batch_size}
-    max_images_per_file: 4096
+    max_images_per_file: 2048
```
jeff-regier, line 121: better to leave it as 4096, unless there's a reason related to this PR to change it

bliss/encoder.py
```diff
@@ -114,15 +115,14 @@ def dist_param_groups(self):
             d[gal_flux] = UnconstrainedLogNormal()
         return d
 
-    def _get_num_input_channels(self):
+    def _get_num_images(self):
         """Determine number of input channels for model based on desired input transforms."""
-        num_channels = 2  # image + background
+        num_imgs = 2  # img + bg
```
jeff-regier, line 120: `num_channels` makes more sense to me; is there something wrong with it?

jeff-regier: (I wouldn't necessarily thing of the background as a distinct image.)

bliss/backbone.py
sawanp813: Will do!

bliss/encoder.py
jeff-regier: Oh I see, this is the length of the third axis below, whereas channels is the second axis. Is there another name for it we can use? (`num_imgs` seems like it could refer to any axis.)

sawanp813: I think in the typical sense, we have `num_channels` set to the number of bands (i.e. in line 90). `num_imgs` definitely isn't the correct term but it is whatever we want to call the third dimension in the 5-D input containing all band-specific things (e.g. image, background, deconvolved image, PSF parameters, etc.)

Definitely open to a better name - just think that `num_channels` should be reserved for the number of bands since the actual number of 'input channels' we pass into the encoder corresponds with the number of bands.

sawanp813: I'm struggling to think of an ideal name - would `num_features` be more appropriate? Still don't think it's great..

bliss/modules.py
```diff
@@ -0,0 +1,45 @@
...
+        self.conv = torch.nn.Conv3d(
+            c1, c2, k, s, autopad(k, p, d), groups=g, dilation=d, bias=False
+        )
+        self.bn = torch.nn.BatchNorm3d(c2)
+        self.act = (
+            self.default_act
+            if act is True
+            else act  # noqa: WPS509
+            if isinstance(act, torch.nn.Module)
+            else torch.nn.Identity()
```
jeff-regier, line 32: This if-else-if-else block is hard to reach, although it's nice that the variable your assigning to (`self.act`) only has to appear once. I wonder if there's a clearer way to write it.

jeff-regier: Can you replace `if act is True` with `if act`?

sawanp813: This is actually also modified yolo code - I'm sure there's a cleaner way to write it without the coverage issue

jeff-regier: I see, it's fine as if it's just copied code -- but please put a comment next to it saying where the code came from

bliss/encoder.py
jeff-regier: `num_features_per_band` seems OK ... is that too long?

jeff-regier (approved):

bliss/encoder.py
sawanp813: It's better than `num_imgs` for sure - think that definitely is descriptive enough. I like it

Issue #806, opened 2023-06-19
probabilistic non-maximal suppression


jeff-regier:

aakashdp6548:
The overall objective of this issue is we want to probabilistically decide which tile a source close to the boundary belongs to. This [link](https://www.overleaf.com/read/pzgfmgymqkkb) explains the details of the approach and some basic implementation ideas. 

This will be divided across separate, mostly independent PRs:
- [ ] Adding a new `RegionCatalog` class + relevant infrastructure for non-constant width regions (#879)
- [ ] Update generation to satisfy constraints on which regions sources can be in
- [ ] Encoder changes - loss computation, `variational_mode`

PR #884, opened 2023-07-18
truncate location distribution in the unit box

### LINKED ISSUE #692: Truncate the location distribution given that a source is in a tile
Currently we just used untruncated normal distributions. Should probably truncate distributions so they lie in their tile.

jeff-regier:
This PR fixes #692 and may help with #806. Perhaps surprisingly, the former pretrained weights still seem to work with this change.

### DIFF
--- a/bliss/unconstrained_dists.py
+++ b/bliss/unconstrained_dists.py
@@ -7,6 +7,7 @@ from torch.distributions import (
     Normal,
     SigmoidTransform,
     TransformedDistribution,
+    Distribution
 )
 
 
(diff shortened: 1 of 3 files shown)


bliss/unconstrained_dists.py
```diff
@@ -32,15 +33,49 @@ def get_dist(self, params):
...
+        prob_in_unit_box_hw = multiple_normals.cdf(torch.ones_like(mu))
+        prob_in_unit_box_hw -= multiple_normals.cdf(torch.zeros_like(mu))
+        self.log_prob_in_unit_box = prob_in_unit_box_hw.log().sum(dim=-1)
+
+        self.base_dist = Independent(multiple_normals, 1)
+
+    def sample(self, **args):
+        # some ideas for how to sample it here, if we need to:
+        # https://cran.r-project.org/web/packages/truncnorm/
+        raise NotImplementedError("sampling a truncated normal isn't straightforward")
```
aakashdp6548, line 52: A question related to #806 - would we ever realistically need to sample from the variational distribution (since we've been talking about it)? Because this would prohibit that from working. If we just want to show how to sample without actually needing to implement it then this should be fine.

aakashdp6548 (approved):
Looks good, just one question.

Issue #726, opened 2023-04-19
DC2 benchmarking


jeff-regier:
A description of DC2: https://arxiv.org/abs/2010.05926
Data access: https://data.lsstdesc.org/doc/dc2_sim_sky_survey

Steps:

- [x] Create a class called `Dc2FullCatalog` in `surveys/lsst.py` that inherits from `FullCatalog` and loads the true DC2 catalog and provides access to it. It should be analogous to `PhotoFullCatalog` and `DecalsFullCatalog`, and be exercised by a pytest test.
- [x] Create a class called `DC2` in `surveys/lsst.py` that inherits from [`pl.LightningDataModule`](https://lightning.ai/docs/pytorch/stable/data/datamodule.html), and implements methods such as `train_dataloader`, `val_dataloader()`, and `test_dataloader()`. These data loaders should provide both images and the true catalog corresponding to each image to the encoder.
- [x] rebase on master and change the `DC2` class to inherit from `Survey` (like `DarkEnergyCameraLegacySurvey` does now)
- [x] train encoder with DC2 data and verify that training loss decreases consistently across epochs
- [x] verify the validation loss decreases with training, that validation metrics (especially f1 score) improve with training, and the validation metrics reach a nontrivial level of performance (i.e., detecting most bright light sources)

PR #885, opened 2023-07-19
add dc2 dataset

XinyueLi1012:
This PR tracks https://github.com/prob-ml/bliss/issues/726 and adds LSST DC2 benchmarking. 

1. Create a class `DC2` in `surveys/dc2.py` that inherit from `Survey` to load dc2 data batchs
2. Create a class called `Dc2FullCatalog` in `surveys/lsst.py` that inherits from `FullCatalog` to load dc2 catalogs
3. Include `merged_catalog_*.pkl` file which contains merged truth catalog and cosmo dc2 galaxy catalog for track 3828 and track 3829 (merging is based on galaxy id)

- For each dc2 `calibrated exposures` (4k * 4k) split it into 25 small image (800 * 800) to fit in GPU
- For each image, run `Dc2FullCatalog.from_file` to load corresponding full catalog
- add `data_source` for training in base_config to avoid binary flag (use_cached_simulator/use_dc2)


bliss/conf/base_config.yaml
```diff
@@ -184,3 +186,10 @@ surveys:
               dec: -0.9316385797930247
         bands: ${encoder.bands}
         reference_band: 1 # DECaLS r-band, DECaLS-indexed (via header)
+    dc2:
+        _target_: bliss.surveys.lsst.DC2
+        data_dir: ${paths.dc2}/calibrated_exposures/r/
+        cat_path: ${paths.dc2}/merged_catalog_bright_all_
+        band: r
+        batch_size: 1
```
jeff-regier, line 194: Batch size isn't really a property of the dataset; it's more like a property of how we're doing training and predictions.

bliss/surveys/lsst.py
```diff
@@ -0,0 +1,262 @@
+import glob
```
jeff-regier, line 1: We probably should rename this file from `lsst.py` to `dc2.py`.

jeff-regier:
We should merge this soon. Please the fix issues above, get the checks passing, and then request a review from @zhixiangteoh .

bliss/encoder.py
```diff
@@ -295,11 +293,13 @@ def _get_loss(self, pred: Dict[str, Distribution], true_tile_cat: TileCatalog):
         gal_bands = [self.GAL_FLUX_NAMES[band] for band in self.bands]
         for i, (star_name, gal_name) in enumerate(zip(star_bands, gal_bands)):
             # star flux loss
+            star_fluxes[star_fluxes == 0] = 1e-18
```
jeff-regier, line 296: Probably better to do `star_fluxes = star_fluxes.clamp(min=1e-18)`, so we don't keep flux values in (0,1e-18).

jeff-regier: Also, is there a way to do this substitution in the DC2 data loader, rather than here? It's only need for DC2 presumably.

bliss/encoder.py
```diff
@@ -179,8 +179,6 @@ def encode_batch(self, batch):
 
         assert inputs.size(2) % 16 == 0, "image dims must be multiples of 16"
         assert inputs.size(3) % 16 == 0, "image dims must be multiples of 16"
-        bn_warn = "batchnorm training requires a larger batch. did you mean to use eval mode?"
-        assert (not self.training) or batch["images"].size(0) > 4, bn_warn
```
zhixiangteoh, line 183: Do we not need these anymore?

tests/test_dc2.py
```diff
@@ -0,0 +1,26 @@
+import torch
+from hydra.utils import instantiate
+
+
+class TestDC2:
+    def test_dc2(self, cfg):
+        dataset = instantiate(cfg.surveys.dc2)
+        dataset.prepare_data()
```
zhixiangteoh, line 8: Better to add `prepare_data()` as a line in `DC2.__init__`?

bliss/train.py
```diff
@@ -23,6 +23,7 @@ def train(cfg: DictConfig):  # pylint: disable=too-many-branches, too-many-state
 
     # setup dataset
     simulator_cfg = cfg.cached_simulator if cfg.training.use_cached_simulator else cfg.simulator
+    simulator_cfg = cfg.surveys.dc2 if cfg.training.use_dc2 else simulator_cfg
```
zhixiangteoh, line 26: Are we training the encoder on real DC2 data and not simulated DC2 data?

bliss/surveys/dc2.py
```diff
@@ -0,0 +1,268 @@
...
+class Dc2FullCatalog(FullCatalog):
+    """Class for the LSST PHOTO Catalog.
+
+    Some resources:
+    - https://data.lsstdesc.org/doc/dc2_sim_sky_survey
+    - https://arxiv.org/abs/2010.05926
+    """
+
+    @classmethod
+    def load_dc2_catalog(cls, cat_path, band, wcs, plocs_lim):
```
zhixiangteoh, line 172: Is it possible to follow `FullCatalog::from_file` pattern here? See `SDSS::PhotoFullCatalog::from_file` and `DecalsFullCatalog::from_file`

zhixiangteoh (changes requested):
Generally looks good. Could you add a description to this PR of the general idea of what it is doing and the proposed changes? Thanks!

bliss/train.py
zhixiangteoh: I see. Since DC2 is simulated data, perhaps it would be good for `DC2` class to also inherit from `CachedSimulatedDataset`, and somehow set `CachedSimulatedDataset.cached_data_path` to be DC2's `survey_data_dir`.

bliss/encoder.py
jeff-regier: No longer. I probably shouldn't have added these in the first place.

.gitattributes
```diff
@@ -1,10 +1,5 @@
...
-case_studies/sdss_galaxies/notebooks/*.ipynb diff=ipynb
-case_studies/sdss_galaxies/notebooks/scratch/*.ipynb filter=nbstripout
-case_studies/sdss_galaxies/notebooks/scratch/*.ipynb diff=ipynb
 data/*.npz filter=lfs diff=lfs merge=lfs -text
-data/*.pt filter=lfs diff=lfs merge=lfs -text
-data/*.pkl filter=lfs diff=lfs merge=lfs -text
 data/*.npy filter=lfs diff=lfs merge=lfs -text
 *.pt filter=lfs diff=lfs merge=lfs -text
 *.fits filter=lfs diff=lfs merge=lfs -text
+data/*.pkl filter=lfs diff=lfs merge=lfs -text
```
jeff-regier, line 5: Did you both deleted this line (from line 7) and add it (to line 5)?

bliss/surveys/dc2.py
```diff
@@ -0,0 +1,271 @@
...
+        self.data_dir = data_dir
+        self.cat_path = cat_path
+        self.band = self.BANDS
+        self.dc2_data = []
+        self.data = []
+        self.valid = []
+        self.test = []
+        self.prepare_data()
+
+    def __len__(self):
```
jeff-regier, line 32: We probably need a basic test of the DC2 class.

bliss/train.py
jeff-regier: That's not a code that `DC2` can reuse from `CacheSimulatedDataset`.
But `DC2` should inherit from the parent classes of `CacheSimulatedDataset`? i.e., `pl.LightningDataModule` and `Dataset`?

jeff-regier: Oh never mind, I see `Survey` already inherits from these two classes.

bliss/conf/base_config.yaml
```diff
@@ -142,6 +143,7 @@ training:
     seed: 42
     weight_save_path: null
     use_cached_simulator: false
+    use_dc2: false
```
jeff-regier, line 146: I think perhaps we should delete both `use_cached_simulator` and `use_dc2`.
Instead, we'd have a new field called `data_source`, which would be set to either a Simulator, CacheSimulator, or DC2.
e.g., `data_source: $simulator`

jeff-regier: or, for example, `data_source: ${surveys.dc2}`

jeff-regier: @zhixiangteoh @XinyueLi1012 Would that make sense? (It seems like we shouldn't have to add another binary flag here anytime we add a data source.)

bliss/train.py
jeff-regier: See my comments above. If we replace `use_cached_simulator` and `use_dc2` with a new field called `data_source`, I think these two lines of can go away. Instead, we'd instantiate `cfg.training.data_source`, whatever it's set to.

bliss/conf/base_config.yaml
zhixiangteoh: I still think we should reuse `cfg.cached_simulator` and make `DC2` is-a `CachedSimulatedDataset`. Because DC2 dataset is essentially a disk dataset which is what `CachedSimulatedDataset` is—despite the slightly misleading name it actually does not do any simulation itself, since the simulation is done at generate step (which we skip for DC2). So we just set the `cfg.cached_simulator.cached_data_path` appropriately and reuse this dataset class as input to encoder, obviating the need for an additional `data_source` conditional parameter, just having `cfg.cached_simulator.cached_data_path` be the only thing that varies.

bliss/train.py
zhixiangteoh: this could work; but see https://github.com/prob-ml/bliss/pull/885#discussion_r1268872096

zhixiangteoh: @XinyueLi1012 reading the file as something other than .pt can possibly be made more general; I just think DC2 should be a `CachedSimulatedDataset` to stay as close as possible to BLISS' code pattern

zhixiangteoh: Is the transformation (to .pt format that BLISS uses) very cumbersome? Also, are there other difficulties with inheriting from `CachedSimulatedDataset`?

bliss/conf/base_config.yaml
zhixiangteoh: On another note, `use_cached_simulator` is a boolean that allows us to still simulate data on the fly, if we want to. However if we are moving away completely from that, then we can remove the parameter?

bliss/train.py
jeff-regier: For SDSS and DECaLS, we don't convert the survey data to `.pt` files. DC2 isn't exactly like these other surveys, in that it has a ground truth that we can use for training, but it is a lot like them; therefore, it's not obvious to me why DC2 should be converted to `.pt` before processing, while we don't require the same of the other surveys.

Not that it would be a problem to do so necessarily. The main drawback I see is that we'd need an additional 5 TB of disk space to store the converted version of DC2. Also, it would add another step, so a bit more complicated from a user's perspective.

bliss/conf/base_config.yaml
jeff-regier: We can still simulate on the fly `data_source: ${simulator}`. Not that we'd necessarily want to at this point because of how slow the simulator is. But I'm optimistic we can make it faster in the future (e.g., with jax-galsim or something like it we develop ourselves)

jeff-regier (changes requested):
Very close. We should include a test in the PR that covers almost all the new code. Currently functions like `DC2.predict_batch` aren't tested.

bliss/train.py
zhixiangteoh: I think for ease of implementation, we can use this `data_source` pattern. But eventually I think it'd be best to generalize the data loading to also handle non-`.pt` disk dataset files.

jeff-regier:
This new test provides good coverage, but as you predicted, it does run sort of slowly on the cpu. Is there anyway to make it faster by testing with less image data? (e.g., a smaller image, or just part of a full image).
@zhixiangteoh may have ideas because of his experience with developing tests for decals.

jeff-regier (approved):
Looks good!

PR #887, opened 2023-07-20
Simulate DECaLS co-added images

### LINKED ISSUE #871: DECaLS end-to-end - generate, train, predict

zhixiangteoh:
Plan:
- Simulate **unaligned** DES images (#893)
  - use original CCD/EXPNUM DES-image's RA/DEC and WCS
  - simulate DES-image PSF via `galsim.DES_PSFEx` (`des.py` > `DES_PSF._get_psf_via_despsfex`)
  - simulate DES-image background splinesky level via `scipy.interpolate.RectBivariateSpline` (`des.py` > `DarkEnergySurvey.splinesky_level_for_band`)
- Make co-add via [`make_coadd`](https://github.com/legacysurvey/legacypipe/blob/ba1ffd4969c1f920566e780118c542d103cbd9a5/py/legacypipe/coadds.py#L218-L651) 
  - [co-add PSF](https://github.com/legacysurvey/legacypipe/blob/ba1ffd4969c1f920566e780118c542d103cbd9a5/py/legacypipe/coadds.py#L486-L519)
  - simulate background via `COSKY_{band}` FITS header param
  
Uses #893 as building blocks. Fixes #871.


zhixiangteoh:
Still some cleaning up to be done, but otherwise ready for review for the general co-add E2E flow

bliss/conf/base_config.yaml
```diff
@@ -52,6 +52,7 @@ simulator:
     survey: ${surveys.sdss}
     prior: ${prior}
     n_batches: 128
+    coadd_depth: 1
```
jeff-regier, line 55: If `coadd_depth` is 1, is that the same as it being a raw exposure? If so, that may not be obvious to a user; a user may think that there's still some extra realignment or PSF homogenization for any coadd. So perhaps we should also have a explicit `use_coaddition` binary variable here too.

bliss/conf/base_config.yaml
```diff
@@ -188,7 +189,11 @@ surveys:
             # brick '3366m010' corresponds to SDSS RCF 94-1-12
             - ra: 336.6643042496718
               dec: -0.9316385797930247
-        bands: ${encoder.bands}
+        bands: [0, 1, 3] # grz
```
jeff-regier, line 192: Why is there a `3` here rather than a `2`?

jeff-regier: i.e., why wouldn't the bands be consecutive?

zhixiangteoh: Just because there is no i-band image for this brick

bliss/simulator/decoder.py
```diff
@@ -151,25 +154,26 @@ def render_images(self, tile_cat: TileCatalog, image_ids):
...
+                for band in range(self.n_bands):
+                    gs_img = galsim.Image(array=images[b, d, band], scale=self.pixel_scale)
+                    for s in range(n_sources):
+                        source_params = full_cat.one_source(b, s)
+                        source_type = source_params["source_type"].item()
+                        renderer = self.source_renderers[source_type]
+                        galsim_obj = renderer(psf, band, source_params)
+                        plocs0, plocs1 = source_params["plocs"]
+                        offset = np.array([plocs1 - (slen_w / 2), plocs0 - (slen_h / 2)])
+                        offset += shift[band]
```
jeff-regier, line 167: Don't you want to apply a different shift here to each raw image `d`? Currently it looks like a particular band (say, the `r` band) will have the same shift in all the raw images.

bliss/conf/base_config.yaml
jeff-regier: Maybe we could remove `bands` from the config files, replacing it with a `use_multiband` binary setting. If true, we use all available bands. If false, we process only the reference band. Would that work?

bliss/simulator/decoder.py
zhixiangteoh: Ah yes, the shifts will probably need to be indexed by `[band, d(epth)]`

bliss/simulator/simulated_dataset.py
```diff
@@ -98,19 +98,31 @@ def simulate_image(
...
+                        ref_band=self.catalog_prior.b_band,
+                    )
+                )
+        images = torch.squeeze(images, dim=1)  # remove coadd depth dimension if 1
+        # (2) coadd via inverse-variance weighting
+        if self.coadd_depth > 1:
+            coadded_images = np.zeros((batch_size, *images.shape[-3:]))  # 4D
+            for b in range(batch_size):
+                coadded_images[b] = self.survey.coadd_images(images[b])
+            images = torch.from_numpy(coadded_images).float()
```
jeff-regier, line 122: It's hard to follow what's going on here. Would you add some comments and/or break this up into smaller functions? It might be better to move "_apply_noise" and this coaddition code to `decoder.py`.

bliss/surveys/survey.py
```diff
@@ -53,6 +59,14 @@ def predict_batch(self):
     def predict_batch(self):
         """Set a batch of data for prediction."""
 
+    def nmgy_to_nelec(self):
```
jeff-regier, line 62: A lot of surveys (e.g., DC2) don't use `nanomaggies` as their units of flux. Is there a more generic way to encode the calibration parameters?

jeff-regier: calibration = conversion from counts to physical units (but not necessarily nanomaggies)

bliss/surveys/survey.py
```diff
@@ -19,6 +20,11 @@ def __init__(self):
 
         self.catalog_cls = None  # TODO: better way than `survey.catalog_cls`?
 
+    @staticmethod
+    def coadd_images(constituent_images):
```
jeff-regier, line 24: Ah, good idea making this a survey-specific routine.

bliss/surveys/survey.py
zhixiangteoh: Yes certainly; a more generic function name and/or arguments should suffice.

jeff-regier (changes requested):
Looks good. Please see changes requested above.

bliss/conf/base_config.yaml
jeff-regier: For debugging and benchmarking, it's still helpful be able to process single-band data, so I don't think we can switch entirely to multiband. But we'd never need to process single-band data that isn't the reference band. And we'd never need to process, say, just the odd numbered bands :-)

bliss/simulator/decoder.py
zhixiangteoh: > Isn't the point of `for d in range(coadd_depth)` loop to generate raw images that differ in some way? (What am I missing here?)

For ease of implementation, this manual dithering via WCS offset + variation due to prior sampling might suffice; I didn't think making the image generation even less synthetic, via something like inferring variations from different exposure numbers, to be viable / worth it. But I might be misunderstanding this whole process completely.

bliss/conf/base_config.yaml
zhixiangteoh: Done.

bliss/simulator/simulated_dataset.py
zhixiangteoh: Broke it up into smaller functions—leaving `_apply_noise` and coaddition in simulator for now.

bliss/surveys/survey.py
zhixiangteoh: Done.

bliss/conf/base_config.yaml
zhixiangteoh: This makes sense. Since this would be a refactoring change, I think it can be tracked by a new PR.

bliss/simulator/decoder.py
```diff
@@ -11,14 +11,19 @@
 
 class ImageDecoder(nn.Module):
     def __init__(
-        self, psf, bands: Tuple[int, ...], nmgy_to_nelec_dict: dict, pixel_shift: int, ref_band: int
+        self,
+        psf,
+        bands: Tuple[int, ...],
+        physical_to_nelec_dict: dict,
```
jeff-regier, line 17: People may not understand the name `physical_to_nelec_dict`. Maybe `flux_calibration` would be more widely understood.

bliss/simulator/decoder.py
```diff
@@ -124,54 +161,55 @@ def render_images(self, tile_cat: TileCatalog, image_ids):
...
+        images = np.zeros((batch_size, coadd_depth, *image_shape), dtype=np.float32)
 
         # use the PSF from specified image_id
         psfs = [self.psf_galsim[image_ids[b]] for b in range(batch_size)]
         param_list = [self.psf_params[image_ids[b]] for b in range(batch_size)]
         psf_params = torch.stack(param_list, dim=0)
 
         # use the specified nmgy_to_nelec ratios indexed by image_id
-        nmgy_to_nelec_rats = [self.nmgy_to_nelec_dict[image_ids[b]] for b in range(batch_size)]
+        nmgy_to_nelec_rats = [self.physical_to_nelec_dict[image_ids[b]] for b in range(batch_size)]
```
jeff-regier, line 179: better not to mention nanomaggies here

jeff-regier: because some surveys don't use these units

bliss/simulator/decoder.py
```diff
@@ -124,54 +161,55 @@ def render_images(self, tile_cat: TileCatalog, image_ids):
...
+                        image_dims=(slen_h, slen_w),
+                        band_shift=band_shifts[band],
+                        depth_shift=depth_shifts[d],
                     )
 
         # clamping here helps with an strange issue caused by galsim rendering
         images = torch.from_numpy(images).clamp(1e-8)
 
-        return images, psfs, psf_params, wcs_batch
+        return images, psfs, psf_params, wcs_batch, wcs_for_depth
```
jeff-regier, line 215: Would you add some more detailed documentation of the values returned by this function? People may not understand the distinction between `wcs_batch` and `wcs_for_depth`

bliss/simulator/simulated_dataset.py
```diff
@@ -44,22 +46,26 @@ def __init__(
...
+        assert survey.pixel_shift is not None, "Survey pixel_shift cannot be None."
+        assert (
+            survey.physical_to_nelec_dict is not None
+        ), "Survey physical_to_nelec_dict cannot be None."
         self.image_decoder = ImageDecoder(
             psf=survey.psf,
             bands=survey.BANDS,
             pixel_shift=survey.pixel_shift,
-            nmgy_to_nelec_dict=survey.nmgy_to_nelec_dict,
+            physical_to_nelec_dict=survey.physical_to_nelec_dict,
```
jeff-regier, line 57: I'd call this `flux_calibration` too. (I don't think people will understand that "physical" refers to "physical units of flux")

bliss/simulator/decoder.py
```diff
@@ -124,54 +161,55 @@ def render_images(self, tile_cat: TileCatalog, image_ids):
...
+                    self.draw_sources_on_band_image(
+                        band_img,
+                        n_sources,
+                        full_cat,
+                        b,
+                        psf,
+                        band,
+                        image_dims=(slen_h, slen_w),
+                        band_shift=band_shifts[band],
+                        depth_shift=depth_shifts[d],
```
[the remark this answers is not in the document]
zhixiangteoh: Done.

bliss/simulator/simulated_dataset.py
zhixiangteoh: Done.

bliss/simulator/decoder.py
zhixiangteoh: No longer needed with the new changes to unalignment

bliss/simulator/decoder.py
zhixiangteoh: Done.

bliss/simulator/decoder.py
zhixiangteoh: Done.

bliss/predict.py
```diff
@@ -9,8 +9,10 @@
 from torch import Tensor
 
 from bliss.catalog import FullCatalog
-from bliss.surveys.decals import DarkEnergyCameraLegacySurvey as DECaLS
-from bliss.surveys.decals import DecalsFullCatalog
+from bliss.surveys.decals import (
+    DarkEnergyCameraLegacySurvey as DECaLS,  # pylint: disable=cyclic-import
+)
+from bliss.surveys.decals import DecalsFullCatalog  # pylint: disable=cyclic-import
```
jeff-regier, line 15: Any way to avoid a cyclic import? These can be quite confusing.

jeff-regier (approved):
Looks great. Please remove the cyclic import if possible and merge when ready.

bliss/predict.py
zhixiangteoh: Yes, believe these are some ways:
1. move/refactor to third file
2. import within function, outside top-level
3. duplicate the code

I think 1 is cleanest and should be done in a future PR if possible. 2 is the simplest and can be done for this PR

zhixiangteoh: I actually ended up implementing a mix of 2 and 3, and an "option 4" that is removing the DECaLS -> DES dependency so that now DES can be thought of to be entirely "encapsulated" by DECaLS.

PR #890, opened 2023-07-21
Decoder bug during flux nmgy to count conversion

sawanp813:
This PR fixes a bug within `decoder.py` that results in background-dominated image generation. This is due to the conversion of fluxes from nmgy to electron count units occuring **after** the `tile_cat` has already been converted to a full catalog within `decoder.py`. Once the background is added back in `simulated_dataset.py`, the result is a set of background-dominated images.


aakashdp6548 (approved):
Looks good

PR #892, opened 2023-07-21
Use correct band for loss computation when not using all bands

aakashdp6548:
We used to use the wrong band for loss computation when using a single band due to `enumerate` not giving us the right indices. Instead, we now iterate over `self.bands` instead to ensure we're using the right index into the flux tensors.

Also, I added small constants in the `log_prob` for `LogNormal` distributions to ensure that the value is in the support (>0). The old code worked because our generated data has values in each tile, even if it contains no sources, but this is more robust anyway. The values where this is needed will get masked out anyway, we just need to make sure it doesn't produce a `nan` since that won't get masked properly.


sawanp813 (approved):
Looks good

PR #895, opened 2023-07-25
Region-based encoder (loss + mode)

aakashdp6548:
This PR adds functionality to compute loss over regions instead over tiles. This is done systematically by first evaluating the probability of the interiors, followed by the boundaries, and finally the corner regions. More details can be found in [this](https://www.overleaf.com/project/64a838b3c59252adc4b5e2c2) document.

The code has the same general structure as before. The main nuance is that we want to evaluate boundary and corner regions as if they were generated from certain tiles, but the number of boundary/corner regions doesn't match the number of tiles, so we need to do a lot of masking and tensor manipulation to evaluate the probabilities correctly. We also need to convert locations from region coordinates to coordinates in the tile we are evaluating in.

### TODO:
- [x] Implement loss in corner regions
- [x] Check for potential bug with interior locs loss
- [ ] Make horizontal boundary regions conditioned on vertical (?)
- [x] Create new subclass of `Encoder` and override `_get_loss`, and move all related code to a case study
- [ ] implement `RegionEncoder.variational_mode` and add metrics and plotting back in


aakashdp6548:
Marking this as ready for review. I'm still working on getting training to work end-to-end, which will probably uncover some bugs in the loss.

jeff-regier (commented):
Looks good so far

aakashdp6548:
Everything is now in `case_studies/prob_tile_suppression`, and can be used by setting `config.training.encoder_target = ${region_encoder}`. Still need to implement `RegionEncoder.variational_mode` to allow computing metrics and plotting detections during training.

aakashdp6548:
Merging to other branch for now, I'll implement mode in new branch/PR and add that to the `prob-tile-suppression` branch before that is ready to merge to master.

PR #896, opened 2023-07-25
Update galsim version

aakashdp6548:
pip doesn't support galsim 2.4.7 anymore, so we're upgrading to the latest version 2.4.10.


sawanp813 (approved):
Looks good.

jeff-regier (approved):

PR #897, opened 2023-07-25
Uploading temporary baseline model supporting 5D input, 3D conv

sawanp813:
This PR uploads a temporary baseline model supporting 5D input with 3D convolutions. The only transform it takes is a `log_transform` of the background-subtracted image.

### DIFF
--- a/data/pretrained_models/3d_conv_log.pt
+++ b/data/pretrained_models/3d_conv_log.pt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:2d5d75819d54d0a9c27e805cd4ef70400f591c56a82abff3181547b49f902586
+size 123516127


jeff-regier (approved):
Why isn't this part of #882?

sawanp813:
> Why isn't this part of #882?

If it wasn't separate then we wouldn't be able to have the API tests passing - same issue that we had with the end-to-end encoder PR a while back.

jeff-regier:
I thought the mocking/monkeypatch stuff took care of that? Which test is it that's problematic?

sawanp813:
> 

I can verify that with @zhixiangteoh and open an issue if that's still the case.

jeff-regier:
I see, in that case, please try to do it all in #882. If it's a problem, we should open an issue about the dependence between branches.

PR #900, opened 2023-07-27
Test DES predict (without plotting SDSS/DECaLS "true" catalogs)

zhixiangteoh:
Finishing up DES E2E, less support for `plot_predict` due to mismatch in total number of available band-pass filters.


jeff-regier (approved):
Looks good

PR #901, opened 2023-07-28
Custom log transforms

sawanp813:
This PR allows for users to specify custom log transformations within the `base_config`. It also sets the number of epochs between validation runs to 1 as a default.


bliss/conf/base_config.yaml
```diff
@@ -81,7 +81,9 @@ encoder:
     input_transform_params:
         use_deconv_channel: false
         concat_psf_params: false
-        log_transform: true
+        log_transform:
+            use_log_transform: true
```
jeff-regier, line 85: Can we omit the `use_log_transform` boolean here? Instead, always use the log transform if it's provided, and don't if it's not.

jeff-regier (changes requested):
Seems like it would be simpler if we removed the `use_log_transform` boolean

bliss/conf/base_config.yaml
sawanp813: Had to make a few other changes so that the tests would all pass with the current baseline model - a few might still fail  (i.e. `test_sdss_reconstruct`, `test_simulate`) because I did the log transform a bit differently for the baseline model than how it is currently done in the encoder. But everything should still run fine.

I have to upload a new model anyways prior to the version release - is this fine?

sawanp813:
> Seems like it would be simpler if we removed the `use_log_transform` boolean

Fixed + encoder still compatible with current baseline model.

bliss/conf/base_config.yaml
jeff-regier: Sounds good

bliss/conf/base_config.yaml
```diff
@@ -81,7 +81,8 @@ encoder:
     input_transform_params:
         use_deconv_channel: false
         concat_psf_params: false
-        log_transform: true
+        log_transform:
+            scales: [10]  # add new log transforms
```
jeff-regier, line 85: @XinyueLi1012 I think the ability to preprocess the images with multiple log transformations simply by modifying the config file will be useful for cataloging DC2.

jeff-regier (approved):

sawanp813:
@jeff-regier The failed tests are either failing on assertions due to model performance (makes sense because the log-transform is being done differently than in the baseline model) or due to a size mismatch (`test_predict_decals_bricks`). 

Should we still go ahead and merge?

jeff-regier:
Not without them passing. Do we need to retrain the model?

sawanp813:
Yeah, I'd have to train two models - one with five bands and another with one band.

PR #902, opened 2023-07-28
Realign imgs

sawanp813:
This PR tracks #891 by perturbing our simulated images during rendering and subsequently realigning them within `simulated_dataset.py` via a customized WCS specified to the random perturbation. 

Adds:
- Unaligned images as the output of `decoder.render_images()` in addition to a list of WCS objects that can align these images.
- Alignment of the unaligned images within `simulated_dateset.simulate_image()` as we do in our real SDSS image processing pipeline.
- Compatible with all surveys by passing in `catalog_prior.b_band` parameter to `decoder` object.


bliss/simulator/decoder.py
```diff
@@ -91,6 +92,22 @@ def source_renderers(self):
...
 
+    def pixel_shift(self):
+        """Generate random pixel shift and corresponding WCS list to undo shifts."""
+        shift = np.random.uniform(-2, 2, (self.n_bands, 2))
+        wcs_base = WCS()
+        base = np.array([5.0, 5.0])
+        wcs_base.wcs.crpix = base
+        wcs = []
+        for i in range(self.n_bands):
+            if i == 2:
```
jeff-regier, line 103: I suppose we should move the reference band to the config file rather than hardcoding it. I'm not sure where in the config to put it though. There's already a reference band for color in the config for the prior, but it's awkward to use that setting here.

bliss/simulator/decoder.py
```diff
@@ -91,6 +92,22 @@ def source_renderers(self):
...
+                wcs.append(wcs_base.low_level_wcs)
+            else:
+                bnd_wcs = WCS()
+                bnd_wcs.wcs.crpix = base + shift[i]
```
jeff-regier, line 107: @zhixiangteoh This code might be useful for your work with coadds. (Just using WCS from DES may not work if you are simulating 80x80 images rather than large ones.) Sawan is dithering the images by a small amount here.

bliss/simulator/decoder.py
jeff-regier: `-2` and `2` probably are survey-specific too. Should we put them in the config file too, so we can vary them for different surveys?

jeff-regier (changes requested):
Looks really good. See changes requested above.

bliss/simulator/decoder.py
sawanp813: I was thinking that about the reference band but I also just realized that it doesn't really matter which band is the reference band in this scenario - it can be arbitrary. For example, we could treat the band indexed with '0' to be the reference band for simplicity.

-2 and 2 as the amount of jitter could probably be placed in the config as well - we'd have to place that within the surveys config.

jeff-regier: I agree, as long the reference band used to generate the simulated data is the same one we use as the reference band at prediction time

sawanp813: That's not the case currently - I think the fix is as simple as changing the`ref_band`within `predict()` (inside of `bliss.predict.py` when we call `align()` to be 0. (Referring to L116 and L121).

zhixiangteoh (approved):
This looks great! Helpful for co-adding images.

bliss/simulator/decoder.py
sawanp813: We just used the `cfg.prior.reference_band` for simplicity and to avoid having multiple things called 'reference band.' But I think it does make sense to override a `cfg.predict.reference_band` instead.

bliss/simulator/simulated_dataset.py
```diff
@@ -92,10 +96,21 @@ def simulate_image(
...
         images += background
+        for i in range(images.shape[0]):
+            images[i] = torch.from_numpy(
+                align(
+                    images[i].numpy(),
+                    wcs_batch[i],
+                    0,
+                )
+            )
+        images = torch.clamp(images, min=1e-6)
```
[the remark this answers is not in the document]
sawanp813: `images` is a tensor so I have to include the `torch.from_numpy()` call since `align()` outputs a NumPy array.

jeff-regier (approved):

PR #906, opened 2023-07-31
Check validation every 1 epoch

sawanp813:
This PR sets the default validation check during training to be for every 1 epoch rather than 10.


jeff-regier (approved):
Thanks!

PR #920, opened 2023-08-09
Fix flux calibrations for DES/DECaLS

zhixiangteoh:
Final fixes for v0.3. Also improves coadd depth-wise alignment but not completely—will address post-v0.3.


jeff-regier (approved):
Looks good

PR #924, opened 2023-08-11
adding relevant plotting code from Mendoza et al. (2023)

sawanp813:
Adding a few functions used to generate the Detection PR and Classification PR figures (9, 10) from Mendoza et al. (2023). 
`create_detection_figure()` is called to reproduce Figure 9 and `create_classification_figure()` is used to reproduce Figure 10. Both take a `data` dictionary as input indexed with values from a catalog necessary to reproduce the figures.


zhixiangteoh (approved):
Thanks for adding these! Will be very helpful for our (+ @XinyueLi1012) detection metrics plotting.

Think the code either:
- doesn't need to get added to `plotting.py` yet; or
- if we want to add it to `plotting.py`, do
  - add `NOTE: ...` code comment(s) in `plotting.py` calling attention to the fact that these are verbatim from Ismael and not compatible with the current status of the code
  - mark the code region as not testable (e.g., `# pragma: no cover`)

jeff-regier:
@sawanp813 Please add tests of the plotting code before merging. It may be necessary to modify the code a bit to be compatible with the rest of our repo.

sawanp813:
> @sawanp813 Please add tests of the plotting code before merging. It may be necessary to modify the code a bit to be compatible with the rest of our repo.

I was talking with @zhixiangteoh and @XinyueLi1012 about this and we figured that it might not be worth adding the actual plotting code back to the repo since all of us need to modify it in different ways for our own purposes (e.g. we can't exactly recover SNR information as easily as is required in the plotting code). So I just wanted to add a branch including my `Figures.ipynb` notebook that has a modified version of the code from Mendoza et al. (2023) used to generate my own plots that we can each use as a skeleton. The intent wasn't to actually add the code back to the repo unless that's something you still think would be a good idea. 

We were thinking that since we each will have to make a different version of the plotting code, adding it in as an active branch/PR would make it easier to find. I can actually close this PR unless you'd like to add an updated version of the plotting code to `plotting.py`.

jeff-regier:
Fair enough. If there's a way to generalize this plotting code so that one version meets everyone's need, it'd be a nice addition to our shared codebase. If not, we can include one-off versions in each case study.

sawanp813:
> Fair enough. If there's a way to generalize this plotting code so that one version meets everyone's needs, it'd be a nice addition to our shared codebase. If not, we can include one-off versions in each case study.

Agreed - I think once we all have our individual figures done, it's definitely worth adding some generalized version of that code to the repo if possible. Otherwise, we can stick with individual case studies like you said.

PR #927, opened 2023-08-17
Don't load real images when simulating data

### LINKED ISSUE #907: Avoid loading real images when generating semi-synthetic images
Survey subclasses should have a binary variable that controls whether images or just metadata are loaded. We only need to load metadata when generating semi-synthetic images.

zhixiangteoh:
Added binary variable `load_image_data` that conditionally loads real survey image data only when it is needed (e.g., in predict).

NOTE: DES image data is currently loaded for DECaLS data simulation because the parameter `sig1` used for DECaLS coadd PSF simulation uses this image data for its computation. This dependence likely can be easily removed.

Fixes #907.

### DIFF
--- a/bliss/surveys/des.py
+++ b/bliss/surveys/des.py
@@ -80,10 +80,13 @@ class DarkEnergySurvey(Survey):
                 "z": "decam/CP/V4.8.2a/CP20170926/c4d_170927_025655_ooi_z_ls9",
             },
         ),
+        load_image_data: bool = False,
     ):
         super().__init__()
 
         self.des_path = Path(dir_path)
+
+        self.load_image_data = load_image_data
 
         self.image_id_list = self.process_image_ids(image_ids)
         self.bands = tuple(range(len(self.BANDS)))
@@ -98,10 +101,12 @@ class DarkEnergySurvey(Survey):
         self.flux_calibration_dict = self.get_flux_calibrations()
 
         self.catalog_cls = TractorFullCatalog
-        self._predict_batch = {"images": self[0]["image"], "background": self[0]["background"]}
+        if self.load_image_data:
+            self._predict_batch = {"images": self[0]["image"], "background": self[0]["background"]}
 
     def prepare_data(self):
-        self.downloader.download_images()
+        if self.load_image_data:
+            self.downloader.download_images()
         self.downloader.download_backgrounds()
         self.downloader.download_psfexs()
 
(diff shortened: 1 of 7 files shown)


bliss/surveys/des.py
```diff
@@ -157,31 +164,34 @@ def read_image_for_band(self, des_image_id, band):
...
+                brickname, ccdname, des_image_id[band], image_shape
             )
             * flux_calibration,
             "wcs": wcs,
             "flux_calibration_list": np.array([[[flux_calibration]]]),
-            "sig1": sig1,
         }
+        if self.load_image_data:
+            # compute sig1 (cf. https://github.com/dstndstn/tractor/blob/cdb82000422e85c9c97b134edadff31d68bced0c/projects/desi/decam.py#L313-L333) # noqa: E501 # pylint: disable=line-too-long
+            if self.load_image_data:
```
jeff-regier, line 185: the same if statement appears twice?

jeff-regier (commented):
Have you verified that semi-synthetic data generation still runs following these changes?

bliss/surveys/sdss.py
```diff
@@ -197,23 +202,25 @@ def read_frame_for_band(self, bl, field_dir, run, camcol, field, gain):
...
+        d = {
             "background": large_sky_nelec,
             "gain": np.array(gain),
             "flux_calibration_list": nelec_per_nmgy,
             "calibration": calibration,
             "wcs": wcs,
         }
+        if self.load_image_data:
+            d.update({"image": pixels_nelec})
+        return d
```
jeff-regier, line 223: Looks good

bliss/surveys/des.py
zhixiangteoh: Ah, sorry, forgot to delete it.

bliss/surveys/sdss.py
```diff
@@ -104,14 +108,15 @@ def prepare_data(self):
...
-        self.downloader.download_images()
-        for rcfgc in self.rcfgcs:
-            run, camcol, field, _ = rcfgc
-            field_path = self.sdss_path / f"{run}/{camcol}/{field}"
-            for bl in SloanDigitalSkySurvey.BANDS:
-                frame_name = f"frame-{bl}-{run:06d}-{camcol:d}-{field:04d}.fits"
-                frame_path = field_path / frame_name
-                assert Path(frame_path).exists(), f"{frame_path} does not exist."
+        if self.load_image_data:
+            self.downloader.download_images()
```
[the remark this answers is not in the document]
zhixiangteoh: Ah yes that's true. Overlooked this because our tests don't expose this (see reply comment below).

jeff-regier (approved):

PR #928, opened 2023-08-17
Generalize predict user API function

zhixiangteoh:
`predict_{sdss|decals}(...)` → `predict("sdss", ...)` or `predict("decals", ...)`


bliss/api.py
```diff
@@ -151,18 +153,9 @@ def predict_sdss(
...
-            OmegaConf.update(cfg, k, v)
-        est_cat, _, _, _, pred_for_image_id = _predict(cfg)
-        est_cat_table = fullcat_to_astropy_table(est_cat)
-        pred_tables = {}  # indexed by image_id
-        for image_id, pred in pred_for_image_id.items():
-            pred_tables[image_id] = pred_to_astropy_table(pred)
-        return est_cat, est_cat_table, pred_tables
+        return self.predict("sdss", weight_save_path, **kwargs)
 
+    @deprecated(deprecated_in_version="0.3.1", message="Use the general predict function instead")
```
jeff-regier, line 158: At this stage, better to just delete it; our codebase is still very much in flux and our code is easier to understand if there's less of it. Let's start worrying about backward compatibility once we release v1.0.

zhixiangteoh: Not a problem!

jeff-regier (approved):

PR #931, opened 2023-08-18
Prune large DECam CCDs PSF params file

### LINKED ISSUE #922: Remove (or significantly reduce size of) ccds-annotated-decam FITS file
Currently, `ccds-annotated-decam-dr9.fits`, from which DECaLS PSF params are extracted, is large (~3.8 GB). Either remove it completely for an alternative, or reduce its size in repo via e.g., using a subset, if possible.

zhixiangteoh:
Fixes #922.


jeff-regier (approved):
Nice!

Issue #904, opened 2023-07-29
data augmentation


jeff-regier:
Each training image (including DC2 images) can be rotated by 90 degrees, increasing the training set size by 4; flipped, doubling the training set size; and shifted up and to the right by between 0, 1, 2, or 3 pixels (tile size is 4), increasing the training set size by 16x. All these transformations together would give us 128 times as many unique examples, without requiring additional data generation/loading.

We'd want to perform these transformations on the fly, perhaps in a parent class of `SimulatedDataset`, `CachedSimulatedDataset`, and `DC2`, right before providing the batch to the encoder. Or else in the `Encoder` itself. An epoch would still consist of one pass over the original training set, but now we'd apply a random transformation each visit.

Some of the transformations here seem useful for manipulating the images:
https://pytorch.org/vision/main/transforms.html

We'd need to write out own methods to transform the shift/flip/rotate the true tile catalog.

Issue #898, opened 2023-07-26
demonstrate that BLISS outperforms the LSST pipeline at cataloging DC2


jeff-regier:
Some changes will likely be required to achieve this result:

- [x] train with randomly selected 80x80 DC2 subimages, not just one 800x800 subimage per batch, for better regularization via batchnorm
- [x] turn on multiband processing for DC2
- [x] turn on preprocessing of the images with multiple types of log transformations
- [x] hyperparameter tuning, especially for learning rate. We want to see a fairly smooth U-shaped validation loss curve
- [x] turn on training with variable PSFs, with both deconvolved images and PSF parameters appended to the encoder input
- [x] try the new rolling z-score image transformation with and without various log transformations
- [x] #904

PR #932, opened 2023-08-23
add data augmentation and dc2 psf/deconv

XinyueLi1012:
1. add data augmentation 
2. add psf/deconv for dc2
will resolve https://github.com/prob-ml/bliss/issues/904 and https://github.com/prob-ml/bliss/issues/898


bliss/conf/base_config.yaml
```diff
@@ -84,6 +84,9 @@ encoder:
         concat_psf_params: false
         log_transform: false
         rolling_z_score: true
+    data_augmentation:
+        do_data_augmentation: true
+        epoch_start: 1
```
jeff-regier, line 89: Do we need the epoch_start option?

bliss/data_augmentation.py
```diff
@@ -0,0 +1,117 @@
+import random
+
+import numpy as np
+import torchvision as TF
```
jeff-regier, line 4: Why TF?

bliss/data_augmentation.py
```diff
@@ -0,0 +1,117 @@
...
+    num_transform_list = [1, 2, 3]
+    num_transform = random.choices(num_transform_list, weights=[0.7, 0.2, 0.1], k=1)[0]
+    transform_list = ["vflip", "hflip", "rotate90", "rotate180", "rotate270", "shift"]
+
+    aug_method = np.random.choice(
+        transform_list, num_transform, p=(0.1, 0.1, 0.1, 0.1, 0.1, 0.5), replace=False
+    )
+
+    aug_image, aug_tile, aug_deconv = origin_full, image, deconv_image
+    for i in aug_method:
```
jeff-regier, line 22: `i` usually denotes an integer. Would you use a more descriptive variable name here?

bliss/data_augmentation.py
```diff
@@ -0,0 +1,117 @@
...
+        if i == "vflip":
+            aug_image, aug_tile, aug_deconv = aug_vflip(aug_image, aug_tile, aug_deconv)
+        if i == "hflip":
+            aug_image, aug_tile, aug_deconv = aug_hflip(aug_image, aug_tile, aug_deconv)
+        if i == "rotate90":
+            aug_image, aug_tile, aug_deconv = aug_rotate90(aug_image, aug_tile, aug_deconv)
```
jeff-regier, line 28: Can these lines be exercised by latest?

bliss/catalog.py
```diff
@@ -488,6 +488,10 @@ def to_tile_params(
         for ii in range(self.batch_size):
             n_sources = int(self.n_sources[ii].item())
             for idx, coords in enumerate(tile_coords[ii][:n_sources]):
+                if coords[0] >= tile_n_sources.shape[1] or coords[1] >= tile_n_sources.shape[2]:
+                    continue
+                # ignore sources outside of the image (usually caused by data augmentation - shift)
```
jeff-regier, line 493: It seems preferable to crop the sources after the shift rather than doing it here

bliss/encoder.py
```diff
@@ -330,6 +335,23 @@ def _get_loss(self, pred: Dict[str, Distribution], true_tile_cat: TileCatalog):
         return loss_with_components
 
     def _generic_step(self, batch, logging_name, log_metrics=False, plot_images=False):
+        do_data_augmentation = self.data_augmentation.get("do_data_augmentation")
+        data_augmentation_start = self.current_epoch >= self.data_augmentation.get("epoch_start")
+
+        if data_augmentation_start and (logging_name == "train") and do_data_augmentation:
+            deconv_image = None
+            if self.input_transform_params.get("use_deconv_channel"):
```
jeff-regier, line 343: Wouldn't we always want to rotate, shift, etc the deconvolution if one exists and we're rotating the main image?

jeff-regier (changes requested):

bliss/conf/base_config.yaml
jeff-regier: i.e., why not always do data augmentation beginning with epoch 0 if augmentation is enabled?

bliss/conf/base_config.yaml
```diff
@@ -227,7 +230,9 @@ surveys:
     dc2:
         _target_: bliss.surveys.dc2.DC2
         data_dir: /nfs/turbo/lsa-regier/lsstdesc-public/dc2/run2.2i-dr6-v4/coadd-t3828-t3829/deepCoadd-results/
-        cat_path: ${paths.dc2}/merged_catalog/merged_catalog_622.pkl
-        batch_size: 128
+        cat_path: ${paths.dc2}/merged_catalog/merged_catalog_psf_100.pkl
```
jeff-regier, line 233: Is the code for creating these pkl files somewhere in our repo, e.g. the scripts directory?

bliss/data_augmentation.py
```diff
@@ -0,0 +1,117 @@
...
+def augment_data(tile_catalog, image, deconv_image=None):
+    origin_tile = TileCatalog(4, tile_catalog)
+    origin_full = origin_tile.to_full_params()
+
+    num_transform_list = [1, 2, 3]
+    num_transform = random.choices(num_transform_list, weights=[0.7, 0.2, 0.1], k=1)[0]
+    transform_list = ["vflip", "hflip", "rotate90", "rotate180", "rotate270", "shift"]
+
+    aug_method = np.random.choice(
+        transform_list, num_transform, p=(0.1, 0.1, 0.1, 0.1, 0.1, 0.5), replace=False
```
jeff-regier, line 18: We'd each of the eight rotation-flip pairs to be equally likely? Does this achieve that? Is there a simpler way to write it?

jeff-regier: Don't we always want to call shift? With probability 1/16 it will shift by (0,0), effectively leaving the image unchanged, right?

bliss/conf/base_config.yaml
jeff-regier: It'd be best to remove `epoch_start` and to assume `epoch_start = 0`. It's easier on users to have few options to set, all else equal. Also, if we're using data augmentation, I think we'd want it from the beginning. Otherwise, we may overfit on some examples (the non-transformed ones) at first while underfitting on others.

bliss/encoder.py
jeff-regier: Can the data augmentation be applied automatically to all the channels, i.e., whichever ones are present, without explicitly referencing any one channel? For example, the background channel should be rotated too if the image is rotated. It'd be nice if that were implemented without us having a special case for the background channel. Same from the deconv channel -- it doesn't seem like a special case should be necessary for it.

bliss/conf/base_config.yaml
jeff-regier: If possible, please include it in a case study as part of this PR.

bliss/data_augmentation.py
jeff-regier: I don't think the code as is gives us uniform coverage of rotation-flip pairs, which is what we want. It also seems a bit hard to understand.

This may be simpler way to do it: First, select one of four rotations (including 0 degrees -- a "no op") and apply. Next, select whether to flip the image vertically or not. (No need to every flip horizontally--that's the same as a vertical flip and a 180 degree rotation.) Finally, apply a random shift.

bliss/data_augmentation.py
```diff
@@ -0,0 +1,117 @@
...
+    plocs = origin_full["plocs"].clone()
+    origin_full["plocs"][:, :, 1] = image_size - plocs[:, :, 0] - 1
+    origin_full["plocs"][:, :, 0] = plocs[:, :, 1]
+    aug_tile = origin_full.to_tile_params(4, 4).get_brightest_source_per_tile().to_dict()
+    return aug_image, aug_tile, aug_deconv
+
+
+def aug_shift(origin_full, image, deconv_image):
+    shift_x = random.randint(0, 3)
+    shift_y = random.randint(0, 3)
```
jeff-regier, line 102: Would it work to select a shift from `random.randint(-1, 2)` instead? I doubt it will make a big difference, but perhaps shifting by -1 rather than 3 seems less likely to truncate sources.

bliss/encoder.py
```diff
@@ -330,6 +335,23 @@ def _get_loss(self, pred: Dict[str, Distribution], true_tile_cat: TileCatalog):
         return loss_with_components
 
     def _generic_step(self, batch, logging_name, log_metrics=False, plot_images=False):
+        do_data_augmentation = self.data_augmentation.get("do_data_augmentation")
```
jeff-regier, line 338: We only want to do data augmentation during training, never during validation or testing. Therefore, could we move the data augmentation call to `training_step(...)` rather than here?

bliss/encoder.py
jeff-regier: In other words, how about concatenating all the images, background, deconv images, etc along the channel axis beforehand? The operations like `vflip` will operate on all of them automatically with just one call.

bliss/data_augmentation.py
jeff-regier: That is, `TF` seems like a potentially misleading name for `torchvision`.

How about instead doing `from torchvision.transforms import functional`?

bliss/catalog.py
jeff-regier: OK sounds good

bliss/data_augmentation.py
```diff
@@ -0,0 +1,98 @@
...
+    return aug_image, origin_full
+
+
+def aug_rotate90(origin_full, image):
+    num_channel = image.size(2)
+    aug_images = []
+    for i in range(num_channel):
+        rotated_slice = TF.rotate(image[:, :, i, :, :], 90)
+        aug_images.append(rotated_slice)
+    aug_image = torch.stack(aug_images, dim=2)
```
jeff-regier, line 42: The first 6+ lines of `aug_rotate90`, `aug_rotate180`, and `aug_rotate270` are duplicated code. Could you move them to a new helper function that takes an angle as an argument?

bliss/data_augmentation.py
```diff
@@ -0,0 +1,98 @@
...
+    plocs = origin_full["plocs"].clone()
+    origin_full["plocs"][:, :, 1] = plocs[:, :, 0]
+    origin_full["plocs"][:, :, 0] = image_size - plocs[:, :, 1] - 1
+    return aug_image, origin_full
+
+
+def aug_rotate180(origin_full, image):
+    num_channel = image.size(2)
+    aug_images = []
+    for i in range(num_channel):
```
jeff-regier, line 54: Is it necessary to loop over the channels? Can you just call `TF.rotate` once for all the channel? (It may be necessary to combine dimensions 1 and 2 first using `rearrange`.)

bliss/data_augmentation.py
```diff
@@ -0,0 +1,98 @@
...
+
+def augment_data(tile_catalog, image):
+    origin_tile = TileCatalog(4, tile_catalog)
+    origin_full = origin_tile.to_full_params()
+    aug_full, aug_image = origin_full, image
+
+    rotate_list = [None, aug_rotate90, aug_rotate180, aug_rotate270]
+    flip_list = [None, aug_vflip]
+    rotate_choice = random.choice(rotate_list)
+    flip_choice = random.choice(flip_list)
```
jeff-regier, line 17: Looks good!

bliss/data_augmentation.py
```diff
@@ -0,0 +1,98 @@
...
+
+def aug_shift(origin_full, image):
+    shift_x = random.randint(-1, 2)
+    shift_y = random.randint(-1, 2)
+    image_size = image.size(3)
+    image_lim = [2 - shift_x, 2 - shift_x + image_size, 2 - shift_y, 2 - shift_y + image_size]
+
+    num_channel = image.size(2)
+    aug_images = []
+    for i in range(num_channel):
```
jeff-regier, line 89: Same here, it'd be nice not to loop over the channels. We should be able to pad all the channels at once.

bliss/encoder.py
```diff
@@ -385,7 +409,8 @@ def _generic_step(self, batch, logging_name, log_metrics=False, plot_images=Fals
 
     def training_step(self, batch, batch_idx, optimizer_idx=0):
         """Training step (pytorch lightning)."""
-        return self._generic_step(batch, "train")
+        do_data_augmentation = self.do_data_augmentation
```
jeff-regier, line 412: Can you do the data augmentation here in `training_step` rather than in `_generic_step`? (We don't want to do data augmentation unless we're training.)

bliss/encoder.py
jeff-regier: Sound good

bliss/data_augmentation.py
jeff-regier: Nice job with these changes

jeff-regier (approved):
Nice

PR #933, opened 2023-08-24
Add sampling from variational distribution

aakashdp6548:
Added rejection sampling from truncated normal and sampling from the region-based variational distribution.

Also fixed a bug with the mean of the truncated normal - the parametrization is based on the untruncated mean and variance. I left the standard deviation alone for now; it's harder to compute and the error is small, and we aren't explicitly using it anywhere.


bliss/unconstrained_dists.py
```diff
@@ -40,26 +40,43 @@ def __init__(self, mu, sigma):
...
     def mean(self):
-        return self.base_dist.mean
+        mu = self.base_dist.mean
+        offset = self.base_dist.log_prob(self.a).exp() - self.base_dist.log_prob(self.b).exp()
+        offset /= self.log_prob_in_unit_box.exp()
+        return mu + offset.unsqueeze(-1)
 
     @property
     def stddev(self):
+        # TODO: add offset term https://en.wikipedia.org/wiki/Truncated_normal_distribution
```
jeff-regier, line 79: How about raising a `NotImplemented` error if this function is called?

bliss/unconstrained_dists.py
```diff
@@ -40,26 +40,43 @@ def __init__(self, mu, sigma):
...
-        # https://cran.r-project.org/web/packages/truncnorm/
-        raise NotImplementedError("sampling a truncated normal isn't straightforward")
+    def sample(self, sample_shape=(1,), **args):
+        """Generate sample using rejection sampling."""
+        q = Independent(Normal(self.base_dist.mean, self.base_dist.stddev), 1)
+        samples = q.sample(sample_shape)
+        valid = (samples.min(dim=-1)[0] >= 0) & (samples.max(dim=-1)[0] < 1)
+        while not valid.all():
+            new_samples = q.sample(sample_shape)
+            samples[~valid] = new_samples[~valid]
```
jeff-regier, line 58: Maybe add a comment explaining this is inefficient, but good enough for now, and that we could do better by filtering the invalid samples from `new_samples` *before* replacing old invalid one with new ones.

case_studies/adaptive_tiling/region_encoder.py
```diff
@@ -79,52 +79,18 @@ def convert_locs_padded_to_unpadded(self, locs):
...
-
-        galsim_dists = [pred[f"galsim_{name}"] for name in self.GALSIM_NAMES]
-        galsim_param_lst = []
-        for d in galsim_dists:
-            # use median for transformed distributions since mode isn't implemented
-            est = d.icdf(torch.tensor(0.5)) if isinstance(d, TransformedDistribution) else d.mode
-            galsim_param_lst.append(est)
-        galaxy_params = torch.stack(galsim_param_lst, dim=3)
-
+    def suppress_tiles(self, d):
```
jeff-regier, line 82: Would you add a docstring explaining what is means to suppress tiles?

case_studies/adaptive_tiling/region_encoder.py
```diff
@@ -79,52 +79,18 @@ def convert_locs_padded_to_unpadded(self, locs):
...
-        batch_size, nth, ntw = tile_is_on_array.shape
+        batch_size, nth, ntw = d["tile_is_on_array"].shape
         threshold = (
             self.overlap_slen / 2 / self.tile_slen,
             1 - (self.overlap_slen / 2 / self.tile_slen),
         )
 
+        locs = d["locs"]
         n_rows, n_cols = nth * 2 - 1, ntw * 2 - 1
         for b, i, j in itertools.product(range(batch_size), range(nth), range(ntw)):
```
jeff-regier, line 92: This loop looks really slow. Is that the case? Can it not be done without a loop, using tensor operations?

case_studies/adaptive_tiling/region_encoder.py
```diff
@@ -79,52 +79,18 @@ def convert_locs_padded_to_unpadded(self, locs):
...
-            if not tile_is_on_array[b, i, j]:
+            if not d["tile_is_on_array"][b, i, j]:
                 continue
 
             new_i, new_j = region_for_tile_source(locs[b, i, j], (i, j), n_rows, n_cols, threshold)
```
jeff-regier, line 96: Could you add a docstring to `region_for_tile_source` that explains the indexing system for regions?

case_studies/adaptive_tiling/region_encoder.py
```diff
@@ -151,24 +121,83 @@ def variational_mode(
                     (i_above, i_above, i_below, i_below),
                     (j_left, j_right, j_left, j_right),
                 )
-                if aux_var[b, i, j] == aux_var[c_idx].max():
-                    tile_is_on_array[c_idx] = 0  # turn off other tiles
-                    tile_is_on_array[b, i, j] = 1
+                if d["aux_var"][b, i, j] == d["aux_var"][c_idx].max():
```
jeff-regier, line 124: This call to `max` looks suspicious, but I don't have a good understanding of what is going on here. In a probabilistic catalog I think you'd only check `max` if you're trying to form a map estimate, not if you're trying to sample.

jeff-regier (changes requested):
Looks good, please add a couple of comments and confirm that the call to `max` isn't a bug.

case_studies/adaptive_tiling/region_encoder.py
aakashdp6548: This max is over the auxiliary variables in the four corner tiles, not over a distribution. So we're checking if the aux var in the current tile is the highest of the four, and if so turning the other ones off. I'll add some better documentation describing what's going on here.

bliss/unconstrained_dists.py
aakashdp6548: Sounds good, I had left it in case we need a rough estimate but we can always just get that from `base_dist.stddev` I guess.

case_studies/adaptive_tiling/region_encoder.py
jeff-regier: Isn't it suppose to be probabilistic? i.e. the tile with the maximum probability isn't always the one selected to explain the corner, it's just the most frequent.

bliss/unconstrained_dists.py
jeff-regier: Yeah I'm wary of returning a value that's close but incorrect...could be difficult to detect a problem

bliss/unconstrained_dists.py
aakashdp6548: Done.

case_studies/adaptive_tiling/region_encoder.py
aakashdp6548: `suppress_tiles()` is supposed to get a point estimate of the auxiliary variables (a sample if coming from `sample()`, or the mode if from `variational_mode()`). So it is probabilistic, it's just being sampled before this point.

That hasn't fully been implemented yet (there's a todo in `sample()`) - so right now it is deterministic. I'll update the documentation and add that in too.

case_studies/adaptive_tiling/region_encoder.py
aakashdp6548: Done!

case_studies/adaptive_tiling/region_encoder.py
aakashdp6548: Done

case_studies/adaptive_tiling/region_encoder.py
aakashdp6548: There probably is some clever way to check all cases (vertical boundary, horizontal boundary, corner) in a vectorized way, but I was having trouble getting it to work so I just used the naive loop version.

Functionally, it's not that bad - it takes 0.02 seconds on average for a 1x40x40 (tiles) image, so should generally be negligible.

jeff-regier (approved):

PR #934, opened 2023-09-05
Pytorch compile

### LINKED ISSUE #905: torch.compile
Looks like we'd need to upgrade to PyTorch 2.0 to use `torch.compile`:
https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html
https://lightning.ai/blog/training-compiled-pytorch-2.0-with-pytorch-lightning/

aakashdp6548:
Upgrade pytorch version to 2.0 and compile yolo model in encoder. Closes #905.

In theory we can get better speedup by compiling all of `encode_batch`. But this is causing a lot of torch dynamo errors/warnings:
```
AttributeError: 'UnspecializedNNModuleVariable' object has no attribute 'module_key'

from user code:
   File "/home/aakashdp/.cache/pypoetry/virtualenvs/bliss-deblender-av05Bskt-py3.10/lib/python3.10/site-packages/torch/nn/modules/container.py", line 120, in __getitem__
    return self._get_item_by_idx(self._modules.values(), idx)

Set torch._dynamo.config.verbose=True for more information


You can suppress this exception and fall back to eager by setting:
    torch._dynamo.config.suppress_errors = True
```

I'm not confident that suppressing them is the right thing to do, so I'm just compiling the yolo model for now (which has more modest speedup, but works without issues).


bliss/encoder.py
```diff
@@ -103,7 +103,8 @@ def __init__(
 
         num_channels = len(self.bands)
         num_features_per_band = self._get_num_features()
-        self.model = Backbone(cfg=arch_dict, ch=num_channels, n_imgs=num_features_per_band)
+        model = Backbone(cfg=arch_dict, ch=num_channels, n_imgs=num_features_per_band)
+        self.model = torch.compile(model)
         self.tiles_to_crop = tiles_to_crop
```
jeff-regier, line 108: How long does it take to compile? If it's more than a few seconds, it'd probably be good to turn compiling off by default (using our config files) to help with interactive runs.

aakashdp6548: Yeah that sounds like a good idea. Compilation takes ~50 seconds, so it'll be amortized in a full training run but not for simple tests.

jeff-regier (approved):
Thanks!

PR #938, opened 2023-09-15
Filter region catalog by flux per tile when converting from tile catalog

aakashdp6548:
Previously, when converting a `TileCatalog` to a `RegionCatalog` we would just discard it if the conversion would produce a region catalog with multiple sources in one tile. Now, we instead keep the highest flux source and turn off all others in that tile.

The high-level algorithm is as follows:
```
1. Create a unique integer index for each region
2. Unfold with a 3x3 kernel with padding=1 and stride=2 to get number of sources and integer indices in each 3x3 tile
3. Determine which tiles have more than one source and the integer index of those regions
4. For each problematic tile:
    - Get the flux of all "on" sources
    - Find the integer index of the region with the highest flux
    - Turn off all regions in the tile (for simplicity)
    - Turn the max flux region back on
```

### DIFF
--- a/case_studies/adaptive_tiling/region_catalog.py
+++ b/case_studies/adaptive_tiling/region_catalog.py
@@ -8,7 +8,7 @@ import torch
 from einops import rearrange, repeat
 from torch import Tensor
 
-from bliss.catalog import TileCatalog
+from bliss.catalog import SourceType, TileCatalog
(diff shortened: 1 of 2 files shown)


jeff-regier (approved):
Great. Thanks Aakash!

PR #939, opened 2023-09-16
To tile coords

XinyueLi1012:
avoid for loop over tiles
 -- use sort(), it would be slow when the image is big (eg. 4000* 4000)

### DIFF
--- a/bliss/catalog.py
+++ b/bliss/catalog.py
@@ -453,7 +453,11 @@ class FullCatalog(UserDict):
         return type(self)(self.height, self.width, d_new)
 
     def to_tile_params(
-        self, tile_slen: int, max_sources_per_tile: int, ignore_extra_sources=False
+        self,
+        tile_slen: int,
+        max_sources_per_tile: int,
+        ignore_extra_sources=False,
+        filter_oob=False,
     ) -> TileCatalog:
         """Returns the TileCatalog corresponding to this FullCatalog.
 
@@ -463,6 +467,9 @@ class FullCatalog(UserDict):
             ignore_extra_sources: If False (default), raises an error if the number of sources
                 in one tile exceeds the `max_sources_per_tile`. If True, only adds the tile
                 parameters of the first `max_sources_per_tile` sources to the new TileCatalog.
+            filter_oob: If filter_oob is true, filter out the sources outside the image. (e.g. In
+                case of data augmentation, there is a chance of some sources located outside the
+                image)
 
         Returns:
             TileCatalog correspond to the each source in the FullCatalog.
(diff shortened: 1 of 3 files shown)


jeff-regier:
Great work @XinyueLi1012! I timed both the old and new versions of `to_tile_coords`: your new one appear to be more than 10 times faster (0.11 seconds vs 1.6 seconds).

I ran both implementations with some test data. The output from both implementations usually matches, so I think the new implementation is at least nearly right. (Possibly it is perfect and the old implementation is slightly off.) But I'm reluctant to merge until they match exactly or until we're really confident that the new version is correct. Please see my trial below. The difference may have to do with how these two implementations handle negative locations. Perhaps we should be filtering sources with out-of-bounds locations (either too high or too low) as part of our data augmentation procedure?

```
> new_tp = aug_full.to_tile_params(4, 4)
> old_tp = aug_full.old_to_tile_params(4, 4)
> aug_full.n_sources.sum()
tensor(4681, device='cuda:0')
> new_tp.n_sources.sum()
tensor(4681, device='cuda:0')
> old_tp.n_sources.sum()
tensor(4681, device='cuda:0')
> (new_tp.locs < 0).sum()
tensor(0, device='cuda:0')
> (old_tp.locs < 0).sum()
tensor(190, device='cuda:0')
> new_tp.locs.isclose(old_tp.locs).sum()
tensor(204610, device='cuda:0')
> (~new_tp.locs.isclose(old_tp.locs)).sum()
tensor(190, device='cuda:0')
```

jeff-regier:
Also, would you add a test to verify that if we rotate a catalog by 90 degrees and next by 270 degree, then we get back the original catalog? (Or do we already have a test like that?)

jeff-regier:
Thanks @XinyueLi1012. Before we merge, can you also verify that our performance on DC2 stays the same with this PR?

tests/test_dc2.py
```diff
@@ -87,17 +87,21 @@ def test_dc2_augmentation(self, cfg):
...
             assert aug_image[0, :, 1, :, :].shape == dc2_obj["background"].shape
             assert aug_image[0, :, 2, :, :].shape == dc2_obj["deconvolution"].shape
             assert aug_full["n_sources"] <= origin_full.n_sources
+
+        # test rotatation
+        aug_image90, aug_full90 = aug_rotate90(origin_full, aug_input_images)
+        _, aug_full270 = aug_rotate270(origin_full, aug_input_images)
+
+        _, aug_full90180 = aug_rotate180(aug_full90, aug_image90)
+        _, aug_full90270 = aug_rotate270(aug_full90, aug_image90)
```
jeff-regier, line 104: This is a good test to have, but I realize it doesn't quite test what I thought it would, because there's no conversion btw tile and full catalogs happening.

Would you add an additional test that converts from a tile catalog to a full catalog and then back again to a tile catalog? (And that the first and last tile catalogs are equal?)

bliss/catalog.py
```diff
@@ -489,7 +489,13 @@ def to_tile_params(
 
         for ii in range(self.batch_size):
             n_sources = int(self.n_sources[ii].item())
-            source_indices = tile_coords[ii][:n_sources]
+            plocs_ii = self.plocs[ii][:n_sources]
+            x0_mask = (plocs_ii[:, 0] > 0) & (plocs_ii[:, 0] < self.height)
+            x1_mask = (plocs_ii[:, 1] > 0) & (plocs_ii[:, 1] < self.width)
+            x_mask = x0_mask * x1_mask
+            n_filter_sources = x_mask.sum()
```
jeff-regier, line 496: I think it'd be better to do the filtering only on demand: better to give users and error if they try to convert catalog formats without first removing out-of-bounds locations. Would you a new method called something like `filter_out_of_bounds`? And then call that method following data augmentation?

jeff-regier: Here's an alternative that may be less tricky: add a keyword argument to `to_tile_coords`  called `filter_oob` which is false by default. Only if it's true will we filter light sources that are out of bounds (either too negative or too large).

PR #941, opened 2023-10-26
revised color model

jeff-regier:
While working on a case study, I noticed that we're generating some crazy fluxes. I saw cases where the `z`-band flux was 10,000,000 times as large as the `r`-band flux. In real data, I don't think you'll find many (any?) objects where fluxes for consecutive bands vary by more than 1 or 2 orders of magnitude, so our color prior seems off here.

This PR partially fixes the problem. I retrained the color model, and revised it to be more canonical: I fit the GMM to log flux ratios for adjacent bands (rather than comparing everything to the r band). Then, in the prior, I clip flux ratios to avoid the most extreme cases. I'm not entirely satisfied with this solution but it seems like an improvement.

@sawanp813 Please check out my changes to `load_ratios.ipynb` notebook.


jeff-regier:
@sawanp813 Near the end of `load_ratios.ipynb` I may have caught/corrected a bug. It appears that the star color GMM was being saved to `gal_gmm_nmgy.pkl` while the galaxy color GMM was being saved to `star_gmm_nmgy.pkl`.

sawanp813:
Changes look good - I actually was aware of that issue (seeing crazy ratios being sampled in the other bands) but didn't get around to correcting/clipping when we populate each source's flux in `simulator`. Thanks for catching that!

Is there a particular reason why we take the ratios relative to adjacent bands? It definitely is a reasonable method - is it just to avoid outliers solely with respect to when we take the ratios against the r-band?

sawanp813:
> @sawanp813 Near the end of `load_ratios.ipynb` I may have caught/corrected a bug. It appears that the star color GMM was being saved to `gal_gmm_nmgy.pkl` while the galaxy color GMM was being saved to `star_gmm_nmgy.pkl`.

I'm guessing I had the order of the models switched when saving via pickle - thanks for noticing that.

jeff-regier:
I'm not sure whether taking flux ratios with respect to the r band was the problem. I did it this way instead because it's more common/idiomatic to model log ratios of fluxes in adjacent bands: these are what are called "colors". The new GMM seems to be working better though. With the old one, once I started clipping unreasonable samples, a lot were being clipped. With the new GMM, it's much less frequent.

sawanp813:
Right - it makes sense to consider adjacent bands since that's how we are defining colors. And if the new GMM doesn't have as many crazy samples, that's also a positive!

PR #948, opened 2023-12-01
Add scripts and notebook for psf experiments

aakashdp6548:


jeff-regier:
Actually, let's merge now so this branch doesn't fall behind

PR #956, opened 2024-02-08
add dc2 script

XinyueLi1012:
This PR adds `train_dc2.sh` in a new case study `dc2`, along with two notebooks illustrating the parameter extraction and some data processing on a raw dc2 coadded image.

- A notebook will be added soon to show some experiment results. 
- When training with more objects, `to_tile_params` crash due to the sorting algorithm we used. (Previously it worked but in #939 we use some indices and sorting algorithm to get rid of the multiple for loops)


case_studies/dc2/train_dc2.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+# run the galaxy parameters and psf parameters extration notebook
+jupyter nbconvert --execute DC2_galaxy_psf_params.ipynb --to html
```
jeff-regier, line 4: I'm getting this error on this line: `ModuleNotFoundError: No module named 'GCRCatalogs'`

Would you run `poetry add --group dev lsstdesc-gcr-catalogs`?

jeff-regier: Same for `pyarrow`

case_studies/dc2/train_dc2.sh
```diff
@@ -0,0 +1,43 @@
...
+
```
jeff-regier, line 5: With those two packages added, I'm getting this error
```
ValueError: `base_dir` ^/lsstdesc-public/dc2/run2.2i-dr6-v4/truth_match is not a valid directory
```

from the call to `jupyter nbconvert --execute DC2_galaxy_psf_params.ipynb --to html`

jeff-regier: It's from this line:
```
 truth_cat = GCRCatalogs.load_catalog('desc_dc2_run2.2i_dr6_truth') 
```

Maybe we need to specify the path to the DC2 data?

case_studies/dc2/train_dc2.sh
```diff
@@ -0,0 +1,43 @@
...
+    train.pretrained_weights = None
+
+# train with data augmentation and psf, deconvolution
+bliss -cp ~/bliss/case_studies/bliss_dc2 -cn config mode=train \
+    train.trainer.logger.name=DC2_MODELS \
+    train.trainer.logger.version=dc2_psf_deconv_da \
+    encoder.image_normalizer.use_deconv_channel=True \
+    encoder.image_normalizer.concat_psf_params=True \
+    encoder.do_data_augmentation = True \
+    train.pretrained_weights = None
```
jeff-regier, line 43: Is there a notebook to execute after all these models are trained that produces figures based on the trained models?

case_studies/dc2/train_dc2.sh
```diff
@@ -0,0 +1,43 @@
...
+jupyter nbconvert --execute DC2_deconv.ipynb --to html
+
+# train base model
+bliss -cp ~/bliss/case_studies/bliss_dc2 -cn config mode=train \
+    train.trainer.logger.name=DC2_MODELS \
+    train.trainer.logger.version=base_model \
+    encoder.image_normalizer.use_deconv_channel=False \
+    encoder.image_normalizer.concat_psf_params=False \
+    encoder.do_data_augmentation = False \
+    train.pretrained_weights = None
```
jeff-regier, line 16: When I try running this command, I get the following error:
```
Primary config directory not found.
Check that the config directory '/home/regier/bliss/case_studies/bliss_dc2' exists and readable
```
Perhaps it should be `dc2` instead of `bliss_dc2`?

case_studies/dc2/train_dc2.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# train both PSF params and deconvolution channel model
+bliss -cp ~/bliss/case_studies/bliss_dc2 -cn config mode=train \
+    train.trainer.logger.name=DC2_MODELS \
+    train.trainer.logger.version=dc2_psf_deconv \
+    encoder.image_normalizer.use_deconv_channel=True \
+    encoder.image_normalizer.concat_psf_params=True \
+    encoder.do_data_augmentation = False \
```
jeff-regier, line 24: the white space around the `=` sign seems to be a problem

case_studies/dc2/train_dc2.sh
```diff
@@ -0,0 +1,43 @@
...
+    train.pretrained_weights = None
```
jeff-regier, line 25: I believe it's `null` rather than `None`

case_studies/dc2/train_dc2.sh
jeff-regier: Any progress?

jeff-regier (approved):
Thanks @XinyueLi1012 !

PR #961, opened 2024-03-06
toy simulator that is 11,000 times faster than galsim

jacksonloper:
((Also some tiny bug fixes where it was assumed that the reference band should always be 2.))

### DIFF
--- a/bliss/encoder/encoder.py
+++ b/bliss/encoder/encoder.py
@@ -43,6 +43,7 @@ class Encoder(pl.LightningModule):
         compile_model: bool = False,
         double_detect: bool = False,
         use_checkerboard: bool = True,
+        reference_band: int = 2,
     ):
         """Initializes Encoder.
 
@@ -61,6 +62,7 @@ class Encoder(pl.LightningModule):
             compile_model: compile model for potential performance improvements
             double_detect: whether to make up to two detections per tile rather than one
             use_checkerboard: whether to use dependent tiling
+            reference_band: band to use for filtering sources
         """
         super().__init__()
 
@@ -78,6 +80,7 @@ class Encoder(pl.LightningModule):
         self.do_data_augmentation = do_data_augmentation
         self.double_detect = double_detect
         self.use_checkerboard = use_checkerboard
+        self.reference_band = reference_band
 
         ch_per_band = self.image_normalizer.num_channels_per_band()
         assert tile_slen in {2, 4}, "tile_slen must be 2 or 4"
@@ -203,7 +206,9 @@ class Encoder(pl.LightningModule):
         return (marginal_loss + white_loss + black_loss) / 2
 
     def _double_detection_nll(self, target_cat1, target_cat, pred):
-        target_cat2 = target_cat.get_brightest_sources_per_tile(band=2, exclude_num=1)
+        target_cat2 = target_cat.get_brightest_sources_per_tile(
+            band=self.reference_band, exclude_num=1
+        )
 
         nll_marginal_z1 = self._single_detection_nll(target_cat1, pred)
         nll_cond_z2 = pred["second"].compute_nll(target_cat2)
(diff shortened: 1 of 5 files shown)


bliss/encoder/encoder.py
```diff
@@ -230,10 +235,15 @@ def _compute_loss(self, batch, logging_name):
...
 
         # filter out undetectable sources
-        target_cat = target_cat.filter_tile_catalog_by_flux(min_flux=self.min_flux_threshold)
+        target_cat = target_cat.filter_tile_catalog_by_flux(
+            min_flux=self.min_flux_threshold,
+            band=self.reference_band,
+        )
 
         # make predictions/inferences
-        target_cat1 = target_cat.get_brightest_sources_per_tile(band=2, exclude_num=0)
```
jeff-regier, line 236: Nice, this is a helpful fix/generalization

bliss/simulator/toy_simulated_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+
+    # get unscaled draws
+    x = torch.exp(-torch.log1p(-u * ratio) / b)
+
+    # scale and shift
+    return x * scale + loc
+
+
+@dataclasses.dataclass
+class ToySimulator:
```
jeff-regier, line 89: Could we inherit from `pl.LightningDataModule` and `IterableDataset` here, for symmetry with `SimulatedDataset`? Ideally `ToySimulator` would be as close to a drop-in replacement for `SimulatedDataset` as possib.e

bliss/simulator/toy_simulated_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+                samp = torch.poisson(brightness + self.noise_level, generator=generator)
+                img += samp
+
+            band_images.append(img)
+
+        # merge
+        return torch.stack(band_images, dim=1)
+
+
+class ToySimulatedDataset(IterableDataset):
```
jeff-regier, line 328: Maybe clearer to merge this class with `ToySimulator`, for symmetry with `SimulatedDataset`.  (The reason we have two classes in `simulated_dataset.py` is that the second reads from disk.)

bliss/simulator/toy_simulated_dataset.py
jeff-regier: Oh maybe this is OK, I see now it's `ToySimulatedDataset` that's intended as the drop-in replacement for `SimulatedDataset`. It does seem a bit cleaner to me to merge `ToySimulator` and `ToySimulatedDataset`, and to inherit from the same classes that `SimulatedDataset` does.

bliss/simulator/toy_simulated_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+    range_tensor = torch.arange(max_count, device=counts.device).unsqueeze(0)
+
+    # Expand counts to match the shape for broadcasting
+    expanded_counts = counts.unsqueeze(1)
+
+    # Compute the mask
+    return range_tensor < expanded_counts
+
+
+def draw_trunc_pareto(b, c, loc, scale, size, torch_generator=None):
```
jeff-regier, line 47: We used to have code like this in `prior.py` but now we use ``scipy.stats.truncpareto`.
Would it be possible to have your simulator take a prior as argument, the way `SimulatedDataset` does, for greater symmetry? Presumably you'd just ignore some of the fields for now (e.g., source type). That would make this more of a drop-in replacement for `SimulatedDataset`, which I think is the goal, at least if we're including this code in our main `bliss/bliss` directory (as opposed to a case study).

bliss/simulator/toy_simulated_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+                # compute brightness [B,height,width]
+                # we briefly store [B,height,width,max_sources] result at this
+                # stage, reaching apex memory usage
+                brightness = (
+                    torch.exp(
+                        -0.5
+                        * (
+                            (plocs_x - xx) ** 2 / self.psf_radius_h**2
+                            + (plocs_y - yy) ** 2 / self.psf_radius_w**2
+                        )  # noqa: C815
```
jeff-regier, line 313: black formatter strikes again. it's cleaner to use intermediate variables.

bliss/simulator/toy_simulated_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+        y = torch.arange(self.height, device=full_catalog.device) + 0.5
+        xx, yy = torch.meshgrid(x, y, indexing="ij")
+
+        # uncollated results will have shape [B,height,width,max_sources]
+        xx = xx[None, :, :, None]
+        yy = yy[None, :, :, None]
+
+        # create B,height,width image for each band
+        band_images = []
+        for b in range(self.n_bands):
```
jeff-regier, line 285: Is it possible to do this without `for` loops?

jeff-regier (changes requested):
These changes may not all be worth making: it depends how long they would take and what are goals are for the toy simulator. It would be nice to replace `SimulatedDataset` with it some day.

PR #964, opened 2024-03-13
Js/redshift estimation

cczw123:
The prior should work for any csv file. The truth_table csv file seems to have all redshift equal 0. I used the photometric csv file for testing.
The main file are prior_truth.py, redshift_truth.yaml, simulated_dataset_truth.py


AcessingPhotometricMagnitudes&Redshifts.ipynb
declanmcnamara, line 1: @cczw123 In your pull requests, can you request the minimal set of changes needed? It will make it easier to review and merge in. So for example, because notebooks aren't critical we can just keep my versions and ideally your versions of Jupyter notebooks wouldn't show up in pull requests. See my overall comments below for suggestions on how to do this.

case_studies/get_redshift.py
declanmcnamara, line 1: This is good to have, I will merge in this new file for reference.

case_studies/get_redshift.sh
declanmcnamara, line 1: Let's omit this file as I think it's pretty straightforward how to use the `get_redshift.py` script.

case_studies/redshift_estimation/prior.py
declanmcnamara, line 1: Rebase on the current version of `dm/redshift_estimation` so that `prior.py` has the two different classes of prior, `RedshiftUniformPrior` and `RedshiftGammaPrior`. After doing that, this file should be identical to the current version on my branch. Afterthat, take your new prior class and put it in this same file (no need to have an additional `prior_truth.py` file).

case_studies/redshift_estimation/prior_truth.py
declanmcnamara, line 1: Can you take this class and it to the `prior.py` file? You can give it a new name, like `RedshiftCSVPrior` or similar. Then you can safely delete this file.

case_studies/redshift_estimation/redshift_truth.yaml
declanmcnamara, line 1: With my other suggestions, we can delete this file. We can use any prior we want by modifying line 8 of `prior.py` to specify the appropriate prior class that we want to use.

case_studies/redshift_estimation/simulated_dataset_truth.py
declanmcnamara, line 1: Does this file have any differences from `simulated_dataset.py` on the current branch? If not, let's keep that file as is without adding a new one. Hopefully, all the necessary overrides can be done in the appropriate new class e.g. `RedshiftCSVPrior` in `prior.py` while leaving this file untouched.

declanmcnamara (changes requested):
@cczw123 Things look good, my comments below just request you to clean up the pull request a bit. In particular, please remove all notebooks from the pull request, and any other files, so that the pull request is only for modifications/additions to two files: `prior.py` and `get_redshifts.py`.

There are a few ways to do this. I think the easiest way would be to:
- create a new branch off of your current one. 
- on that branch, revert all notebooks back to current versions from `origin/dm/redshfit_estimation`. You should be able to do this by running `git fetch` followed by the commands similar to those [here](https://stackoverflow.com/questions/215718/how-can-i-reset-or-revert-a-file-to-a-specific-revision).
- Delete any unnecessary files you don't want to show up in the pull request (e.g. new notebooks you made).
- Open a pull request from this new branch. Once I merge it, you can safely delete this branch.

And then you can continue working on your usual branch. B/c of these changes, I think it may be easier to close this pull request and do a new one, but it's up to you.

AcessingPhotometricMagnitudes&Redshifts.ipynb
cczw123: Yes I will delete this and make a new one

PR #965, opened 2024-03-20
Js/redshift merge

cczw123:
add the get_redshift.py and have a new class in prior getting data from csv file.

### DIFF
--- a/case_studies/redshift_estimation/prior.py
+++ b/case_studies/redshift_estimation/prior.py
@@ -1,5 +1,7 @@
 from torch.distributions import Gamma, Uniform
-
+import torch
+import pandas as pd
+import numpy as np
 from bliss.simulator.prior import CatalogPrior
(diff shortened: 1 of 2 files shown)


case_studies/redshift_estimation/prior.py
```diff
@@ -110,3 +112,69 @@ def sample(self) -> RedshiftTileCatalog:
...
+    """
+
+    def __init__(
+        self,
+        *args,
+        **kwargs,
+    ):
+        """Initializes CatalogPrior."""
+        super().__init__(*args, **kwargs)
+        self.array = pd.read_csv('/home/../data/scratch/songju/STAR.csv').to_numpy()[:,1:7]
```
declanmcnamara, line 130: In the future, we may want to make this the file path to the csv an argument, but for now it's fine to leave it hard-coded.

declanmcnamara (commented):
Looks good! Will merge this into my branch.

PR #966, opened 2024-03-27
adding NLL metrics commits

yolandazzz13:
Added changes needed for compiling redshift NLL. But yamls still need to be modified.


declanmcnamara (commented):
Looks good, I will be merging these in. After merging, I might make some edits to `redshift.yaml` again as stated.

PR #968, opened 2024-03-31
Update prior.py for galaxy prior

cczw123:
Add prior for galaxy reading from csv


declanmcnamara (commented):
Thanks @cczw123, I will merge in this new class!

PR #970, opened 2024-05-05
Removed deconvolution

aakashdp6548:
Removed deconvolved image functionality during generation and training. The `psf_variation` case study still has some references to old functionality but I need to update that anyway with other changes so I'm leaving it for the other PR.


jeff-regier (approved):
Thanks Aakash

PR #984, opened 2024-05-08
Weak lensing

timwhite0:
This PR adds inference capabilities for weak lensing shear and convergence.

It involves minimal changes to the BLISS infrastructure:

- Adds shear and convergence as allowable params in `TileCatalog`.
- Adds a bivariate normal variational dist for shear (which has two components).

The remaining additions are self-contained in `case_studies/weak_lensing`:

- Lensing subclasses of `ImageDecoder`, `CatalogPrior`, `VariationalDistSpec`, and `VariationalDist`.
- A new lensing metric (tile-level MSE for shear and convergence).
- Several tutorial notebooks to demonstrate end-to-end functionality.


bliss/encoder/unconstrained_dists.py
```diff
@@ -35,6 +35,19 @@ def get_dist(self, params):
         return Normal(mean, sd)
 
 
+class ShearUnconstrainedNormal:
```
jeff-regier, line 38: It be better if this distribution had a name that wasn't specific to shear, so that it can be reused for other applications

bliss/encoder/plotting.py
```diff
@@ -112,3 +112,111 @@ def plot_detections(images, true_tile_cat, est_tile_cat, margin_px, ticks=None,
 
     fig.tight_layout()
     return fig
+
+
+def plot_maps(images, true_tile_cat, est_tile_cat, figsize=None):
```
jeff-regier, line 117: Ideally case studies would be able to define new plotting functions like this one within a case study subdirectory, rather than in the bliss package. These plotting routines would be passed to the encoder when instantiating it, like how metrics are handled already.

timwhite0: This will be handled in a separate PR.

jeff-regier (approved):
Looks good!

PR #986, opened 2024-05-10
Fix DC2 experiment

YicunDuanUMich:
* Fixed the DC2_exp.ipynb
* Slightly changed the DC2 class to enable it to give different output (including wcs and full_catalog) during plotting
* However, we still can't reprocess the result now, because the training failed. I trained the model on 3828 and 3829 last night. It took me about 12h (about 3~4h data loading and 8h training). The model can't converge during training and it early stopped at 15 epoch.

image


bliss/surveys/dc2.py
```diff
@@ -18,7 +18,7 @@ class DC2(Survey):
     # why are these bands out of order? why does a test break if they are ordered correctly?
     BANDS = ("g", "i", "r", "u", "y", "z")
 
-    def __init__(self, data_dir, cat_path, batch_size, n_split, image_lim):
+    def __init__(self, data_dir, cat_path, batch_size, n_split, image_lim, for_plotting=False):
```
jeff-regier, line 21: It's cleaner if the DC2 survey class doesn't "know" about whether it's being used for plotting or not. Survey classes are just supposed to provide access to the underlying data. Better to move the logic elsewhere, presumably to where the DC2 dataset is being used.

case_studies/dc2/train_dc2.sh
```diff
@@ -33,3 +33,6 @@ bliss -cp ~/bliss/case_studies/dc2 -cn config mode=train \
     train.trainer.logger.version=dc2_psf_da \
     encoder.image_normalizer.concat_psf_params=True \
     encoder.do_data_augmentation=True
+
+# new
+nohup bliss -cp ~/bliss/case_studies/dc2 -cn full_train_config > xxx.out 2>&1 &
```
jeff-regier, line 38: Would you modify this script so it runs the notebooks after training? e.g.,
```
jupyter nbconvert --execute DC2_galaxy_psf_params.ipynb --to html
```
Also, is there a different config file we should be using within this script, since `config.yaml` is no more? Ideally this script would run without errors a recreate the figures in the paper.

jeff-regier:
@YicunDuanUMich Please rebase on master and then force push to fix the conflict with the poetry.lock file. Conflicts in the poetry.lock file are a bit different than conflicts in other files because poetry.lock is auto generated. Here are some detailed instruction from ChatGPT (which look reasonable, but may have problems) for how to do so. 

> 
> If you want to resolve a merge conflict involving the `poetry.lock` file by preferring the version of the file on `master` and then rebasing your current branch onto the updated `master`, here's a step-by-step guide:
> 
> 1. Make sure you've committed or stashed any other changes on your current branch, as the rebase process can interfere with them.
> 
> 2. Fetch the latest commits from `origin` (where `origin` is the remote name which refers to the original repository you cloned from). This will update your local copy of the branches but won't affect your current branch:
> 
> ```bash
> git fetch origin
> ```
> 
> 3. Switch to the `master` branch (or whatever is your main integration branch):
> 
> ```bash
> git checkout master
> ```
> 
> 4. Pull the latest changes into your local `master`:
> 
> ```bash
> git pull origin master
> ```
> 
> 5. Now switch back to the branch where you were resolving the conflict:
> 
> ```bash
> git checkout your-feature-branch
> ```
> 
> 6. If you want to resolve any conflict by choosing the `master` version of `poetry.lock`:
> 
> ```bash
> git checkout origin/master -- poetry.lock
> ```
> 
> 7. Now it's time to rebase your feature branch on top of the recently updated `master` (this process may result in more merge conflicts which you'll need to resolve manually):
> 
> ```bash
> git rebase master
> ```
> 
>    While rebasing, if you encounter any conflicts with files other than `poetry.lock`, you will need to resolve them manually and continue the rebase:
> 
> ```bash
> git add <file-name>
> git rebase --continue
> ```
> 
>    If at any point you want to abort the rebase process, you can do so with:
> 
> ```bash
> git rebase --abort
> ```
> 
> 8. After resolving conflicts and successfully rebasing, your `poetry.lock` file is now effectively the version from `master`. Now you can run:
> 
> ```bash
> poetry update
> ```
> 
>    This will update your dependencies according to your `pyproject.toml` file and regenerate the `poetry.lock` file.
> 
> 9. After doing so, you'll want to commit the changes:
> 
> ```bash
> git add poetry.lock
> git commit -m "Update poetry.lock"
> ```
> 
> 10. At this point, your branch is rebased on top of `master` with a new `poetry.lock` file. You can now push your changes to the remote repository. Since the history of your branch has changed due to the rebase, if you've previously pushed this branch to the remote, you'll need to force push:
> 
> ```bash
> git push origin your-feature-branch --force
> ```
> 
> Caution: Force pushing can rewrite history and potentially cause issues for anyone else working on the same branch. Make sure to coordinate with your team before using `--force`.
> 
> Always ensure you have a backup of your work before performing operations like `rebase` and `push --force`, as these can potentially lead to loss of data if not handled carefully.
>

YicunDuanUMich:
The solution given by ChatGPT may not work. The poetry will throw error "can't overwrite xxx line" during "poetry update". So I regenerate my local poetry env using the latest poetry.lock file in our repository. However, this will still lead to merge conflict (due to the hash code in poetry.lock file). I have to fix this manually by merging the master into my branch.

.gitignore
```diff
@@ -35,3 +35,4 @@ case_studies/*/data/
 venv
 nohup.out
 multirun
+case_studies/dc2/*.pkl
```
jeff-regier, line 38: Could you create a .gitignore file within case_studies/dc2 rather than modifying the top-level one?

bliss/catalog.py
```diff
@@ -5,11 +5,13 @@
 from typing import Dict, Tuple
 
 import torch
+import torch.nn.functional as F
 from astropy import units as u  # noqa: WPS347
 from astropy.table import Table, hstack
 from astropy.wcs import WCS
 from einops import rearrange, reduce, repeat
 from torch import Tensor
+from torch.nn.utils.rnn import pad_sequence
```
jeff-regier, line 14: Could we use the regular pad function instead, even if it's a bit more verbose? Future developers may not be familiar with the rnn submodule

bliss/catalog.py
```diff
@@ -61,6 +63,70 @@ def __init__(self, tile_slen: int, d: Dict[str, Tensor]):
         self.batch_size, self.n_tiles_h, self.n_tiles_w, self.max_sources = self.locs.shape[:-1]
         super().__init__(**d)
 
+    def _test_tensor_all_close(self, left, right):
+        return torch.allclose(left, right, atol=1e-4)
```
jeff-regier, line 67: Will the checks pass with a smaller atol? 1e-4 is a bit lax for some values

bliss/catalog.py
```diff
@@ -422,6 +488,63 @@ def __init__(self, height: int, width: int, d: Dict[str, Tensor]) -> None:
         assert self.n_sources.shape == (self.batch_size,)
         super().__init__(**d)
 
+    def _test_tensor_all_close(self, left, right):
+        return torch.allclose(left, right, atol=1e-4)
```
jeff-regier, line 492: It's better to define this method outside of both catalog classes so it doesn't have to be duplicated. If we even need it. Could we just call `torch.allclose` directly with the default `atol` instead of calling `_test_tensor_all_close`? The code would be more readable if we could avoid one level of indirection.

bliss/encoder/metrics.py
```diff
@@ -178,6 +179,192 @@ def plot(self):
         return fig, axes
 
 
+class DetectionPerformanceOnFlux(DetectionPerformance):
```
jeff-regier, line 182: This class appears to be duplicating much of the code in `DetectionPerformance` (twice in fact -- once for stars and once for galaxies). That's probably not what we want. If our goal is to break out detection performance by source type, can we just add an argument to the constructor `DetectionPerformance` that filters by source type if the argument is not `None`? Then, in the hydra config file, we'd instantiate `DetectionPerformance` three times: once for all light source, once for stars, and once for galaxies.

bliss/surveys/dc2.py
```diff
@@ -262,33 +382,43 @@ def from_file(cls, cat_path, wcs, height, width, band):
             "star_log_fluxes": star_log_fluxes.reshape(1, nobj, 6),
         }
 
-        return cls(height, width, d), torch.stack(psf_params)
+        if return_match_id:
```
jeff-regier, line 385: Can we always return match_id? A return type that varies in dimension is something of an anti-pattern.

bliss/surveys/dc2.py
```diff
@@ -262,33 +382,43 @@ def from_file(cls, cat_path, wcs, height, width, band):
...
 
         image_frame.close()
         bg_frame.close()
 
         image = torch.nan_to_num(torch.from_numpy(image_data)[: image_lim[0], : image_lim[1]])
         bg = torch.from_numpy(bg_data.astype(np.float32)).expand(image_lim[0], image_lim[1])
 
+        # to make sure the background are positive
+        bg = bg.clone()
+        bg[bg <= 1e-6] = 1e-5
```
jeff-regier, line 415: Probably better to leave these unchanged. If they are negative and we run code that is expecting positive values, it would be better to have the code raise an error.

bliss/surveys/dc2.py
```diff
@@ -262,33 +382,43 @@ def from_file(cls, cat_path, wcs, height, width, band):
...
+
         image += bg
         image_list.append(image)
         bg_list.append(bg)
 
-    return image_list, bg_list, wcs
+    return image_list, bg_list, wcs, wcs_header_str
```
jeff-regier, line 421: Can we avoid returning both `wcs` and `wcs_header_str`?

.gitignore
YicunDuanUMich: Fix it in commit 1987fcc

bliss/catalog.py
YicunDuanUMich: I failed to find a better equivalent of torch's `pad_sequence`. I search for such function using gpt, and it gives the following result:
```python
def pad_sequence(sequences, padding_value=0):
    max_len = max(len(seq) for seq in sequences)
    padded_seqs = [
        seq + [padding_value] * (max_len - len(seq))
        for seq in sequences
    ]
    return padded_seqs

# Example usage
sequences = [[1, 2, 3], [4, 5], [6]]
padded_seqs = pad_sequence(sequences)
print(padded_seqs)
```

Maybe we can write more comments for `pad_sequence` to explain it better.

bliss/catalog.py
YicunDuanUMich: I'll fix it in the next commit

YicunDuanUMich: When testing these tensors, the tensors of fluxes sometimes mismatch and `atol=1e-4` is the minimal tolerance that can suppress this warning.

bliss/surveys/dc2.py
YicunDuanUMich: I return `wcs_header_str` because i want to ensure we can rebuild the same wcs from header string. In wcs's document, we can find the warning w.r.t `to_header` and `to_header_string` that 

> This function does not write out FITS WCS [distortion paper](https://www.atnf.csiro.au/people/mcalabre/WCS/dcs_20040422.pdf) information, since that requires multiple FITS header data units. To get a full representation of everything in this object, use [to_fits](https://docs.astropy.org/en/stable/api/astropy.wcs.WCS.html#astropy.wcs.WCS.to_fits).

Therefore, if we only return wcs and use it to generate header string, we may lose some information. Maybe we can just return `wcs_header_str`.

YicunDuanUMich:
I'll fix the failed test in this afternoon

bliss/catalog.py
jeff-regier: I see. Is that because the fluxes really large? If so, maybe we could specify `rtol` too?

bliss/catalog.py
jeff-regier: OK I see, I'm onboard with using the rnn submodule.

bliss/surveys/dc2.py
```diff
@@ -1,55 +1,217 @@
+import collections.abc  # noqa: WPS301
```
jeff-regier, line 1: Would changing this to
```
from collections import abc
```
satisfy the style checker, without disabling it?

bliss/catalog.py
YicunDuanUMich: i'll try it

bliss/surveys/dc2.py
```diff
@@ -1,55 +1,217 @@
...
+    tile_dict = tile_cat.to_dict()
+
+    tile_dict["locs"] = rearrange(tile_cat["locs"], "1 h w nh nw -> h w nh nw")
+    tile_dict["n_sources"] = rearrange(tile_cat["n_sources"], "1 h w -> h w")
+    tile_dict["source_type"] = rearrange(tile_cat["source_type"], "1 h w nh nw -> h w nh nw")
+    tile_dict["galaxy_fluxes"] = rearrange(tile_cat["galaxy_fluxes"], "1 h w nh nw -> h w nh nw")
+    tile_dict["galaxy_params"] = rearrange(tile_cat["galaxy_params"], "1 h w nh nw -> h w nh nw")
+    tile_dict["star_fluxes"] = rearrange(tile_cat["star_fluxes"], "1 h w nh nw -> h w nh nw")
+    tile_dict["star_log_fluxes"] = rearrange(
+        tile_cat["star_log_fluxes"], "1 h w nh nw -> h w nh nw"
```
jeff-regier, line 60: A for loop here might make the code more concise. e.g.,
```
for k, v in tile_dict.items():
    if k != "n_sources":
        tile_dict[k] = rearrange(v, "1 h w nh nw -> h w nh nw")
```

case_studies/dc2/train_dc2.sh
```diff
@@ -1,35 +1,5 @@
...
-    encoder.do_data_augmentation=True
-
-# train with data augmentation and psf
-bliss -cp ~/bliss/case_studies/dc2 -cn config mode=train \
-    train.trainer.logger.name=DC2_MODELS \
-    train.trainer.logger.version=dc2_psf_da \
-    encoder.image_normalizer.concat_psf_params=True \
-    encoder.do_data_augmentation=True
+jupyter nbconvert --execute DC2_galaxy_psf_params.ipynb --to html
+cp merged_catalog_with_flux_over_100.pkl ../../data/dc2/merged_catalog/
```
jeff-regier, line 3: Would it be possible to move `../../data/dc2` to the within `case_studies/dc2`? Then this script would be more self contained.

case_studies/dc2/train_dc2.sh
```diff
@@ -1,35 +1,5 @@
...
-    train.trainer.logger.version=dc2_data_aug \
-    encoder.image_normalizer.concat_psf_params=False \
-    encoder.do_data_augmentation=True
-
-# train with data augmentation and psf
-bliss -cp ~/bliss/case_studies/dc2 -cn config mode=train \
-    train.trainer.logger.name=DC2_MODELS \
-    train.trainer.logger.version=dc2_psf_da \
-    encoder.image_normalizer.concat_psf_params=True \
-    encoder.do_data_augmentation=True
```
jeff-regier, line 35: Could you include a call to `bliss mode=train ... ` at the beginning of this script? The idea of the script is that anyone can run our complete workflow from start to finish, without getting errors, and reproduce the results in our manuscript.

bliss/surveys/dc2.py
jeff-regier: I wonder if the distortion paper information is important...perhaps not if it doesn't get stored to the WCS object. In any case, it could be fine to only return `wcs_header_str`

bliss/catalog.py
jeff-regier: Oh it looks like rtol is set by default to a pretty small value. It must be that the fluxes were very large.

bliss/surveys/dc2.py
jeff-regier: It'd be even better if we didn't need to squeeze these tensors...is that possible? (Perhaps by using them below in the unsqueezed format.)

bliss/surveys/dc2.py
```diff
@@ -1,55 +1,217 @@
...
+
+
+def map_nested_dicts(cur_dict, func):
+    if isinstance(cur_dict, collections.abc.Mapping):
+        return {k: map_nested_dicts(v, func) for k, v in cur_dict.items()}
+
+    return func(cur_dict)
+
+
+def generate_split_file(image_index, object_attributes):
```
jeff-regier, line 31: Is there a more descriptive variable name than `object_attributes` we could use here? (Or could we break up the argument into multiple arguments?)

case_studies/dc2/full_train_config.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+
+surveys:
+    dc2:
+        _target_: bliss.surveys.dc2.DC2
+        data_dir: /data/scratch/dc2local/run2.2i-dr6-v4/coadd-t3828-t3829/deepCoadd-results/
+        split_result_folder: split_results_05_16_2024
+        cat_path: ${paths.dc2}/merged_catalog/merged_catalog_with_flux_over_100.pkl
+        batch_size: 64
+        n_split: 50
```
jeff-regier, line 44: Could we make `n_split` much smaller (5 perhaps) and set `batch_size` to 1? Then there'd be fewer light sources missed because they are on the border between splits.

case_studies/galaxy_clustering/README.md
```diff
@@ -37,24 +37,23 @@ This project is being conducted under the Undergraduate Research Program in Stat
...
-So far, the only change within the cluster was to increase the rate of galaxies. That is, we use the Possion distribution to first sample the number of sources based on number of tiles and the average number of sources within tiles. A predetermined probability of a source being a galaxy determines how many galaxies will ultimately be present within the image. 
+So far, the only change within the cluster was to increase the rate of galaxies. That is, we use the Possion distribution to first sample the number of sources based on number of tiles and the average number of sources within tiles. A predetermined probability of a source being a galaxy determines how many galaxies will ultimately be present within the image.
 
-For implementation details, we generate location, fluxes, params, and types for the cluster first. Then we follow the normal procedure of generating sources. In the end, we stack two sets of vectors together to have the final values. 
+For implementation details, we generate location, fluxes, params, and types for the cluster first. Then we follow the normal procedure of generating sources. In the end, we stack two sets of vectors together to have the final values.
```
jeff-regier, line 53: Better to leave the galaxy_clustering case study unmodified in this PR. (Even changing whitespace can cause merge conflicts and changes the git history of the file.)

bliss/catalog.py
YicunDuanUMich: I move it to test_dc2.py in commit a521a09.

bliss/surveys/dc2.py
YicunDuanUMich: Delete it in commit a521a09.

bliss/surveys/dc2.py
YicunDuanUMich: We now only return `wcs_header_str` in commit a521a09.

bliss/surveys/dc2.py
YicunDuanUMich: fix it in commit a521a09, and we will remove the squeeze in next PR

case_studies/dc2/train_dc2.sh
YicunDuanUMich: Fix it in commit a521a09.

case_studies/dc2/train_dc2.sh
YicunDuanUMich: we move all the config to `full_train_config.yaml`

bliss/surveys/dc2.py
YicunDuanUMich: change it to `self_copy` in commit a521a09

case_studies/dc2/full_train_config.yaml
YicunDuanUMich: we will address this in next PR

YicunDuanUMich:
The test fails because it can't access the data file in our disk.

> FAILED tests/test_dc2.py::TestDC2::test_dc2_size_and_type - PermissionError: [Errno 13] Permission denied: '/data'
FAILED tests/test_dc2.py::TestDC2::test_dc2_augmentation - PermissionError: [Errno 13] Permission denied: '/data'

I'll fix it this afternoon.

jeff-regier (approved):
Looks good @YicunDuanUMich . Thanks for all your work. We would ideally have no drop in pytest coverage, but I believe @shreyasc30 is waiting on these changes, so it's fine to merge as is, or whenever you're ready.

PR #987, opened 2024-05-10
Qh/redshift estimation

georgeyfly:
Built a network that can predict redshift from ground truth color.


case_studies/redshift_estimation/catalog.py
```diff
@@ -0,0 +1,50 @@
+import torch
+from einops import rearrange, reduce, repeat
+
+from bliss.catalog import FullCatalog, TileCatalog
+
+
+class RedshiftTileCatalog(TileCatalog):
```
jeff-regier, line 7: We should be able to delete these redshift-specific catalogs by using TileCatalog and FullCatalog instead to store redshift.

bliss/encoder/encoder.py
```diff
@@ -10,13 +10,14 @@
...
 from torchmetrics import MetricCollection
 
-from bliss.catalog import TileCatalog
 from bliss.encoder.convnet import CatalogNet, ContextNet, FeaturesNet
 from bliss.encoder.data_augmentation import augment_batch
 from bliss.encoder.image_normalizer import ImageNormalizer
 from bliss.encoder.metrics import CatalogMatcher
 from bliss.encoder.plotting import plot_detections
 from bliss.encoder.variational_dist import VariationalDistSpec
+from case_studies.redshift_estimation.catalog import RedshiftTileCatalog
```
jeff-regier, line 19: This seems backwards. We want the case studies to import the bliss package, not the other way around.

case_studies/redshift_estimation/catalog.py
georgeyfly: I think this is undergraduate wrote and not needed to be merged to master branch. So I deleted whole redshift_estimation dir now.

jeff-regier:
@georgeyfly It looks like the various continuous integration tests for your PR are failing. (see [log here](https://github.com/prob-ml/bliss/actions/runs/9142506665/job/25138129607?pr=987)). Have you run `pre-commit install` within the bliss directory? It's easier to fix these issues locally, either as you make commits, or by running `pre-commit run --all`.

bliss/encoder/plotting.py
```diff
@@ -112,3 +112,79 @@ def plot_detections(images, true_tile_cat, est_tile_cat, margin_px, ticks=None,
 
     fig.tight_layout()
     return fig
+
+
+# Adapted from plot_detections in bliss.plotting
+def my_plot_detections(
```
jeff-regier, line 118: Did changes to this file get include accidentally?

case_studies/redshift/network_rs.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def training_step(self, batch, batch_idx):
+        x, y = batch
+        x = x.view(-1, x.shape[-1])
+        y = y.view(-1, y.shape[-1])
+        y_hat = self.model(x)
+        if self.model.out_dim != 1:
+            y = y.long()
+        loss = self.loss_fcn(y_hat, y)
+        self.log('train_loss', loss, prog_bar=True)
```
jeff-regier, line 72: There's quite a bit of code that is duplicated across `training_step`, `validation_step`, and `test_step`.  It'd be good to refactor, moving shared code to a new function called `generic_step` that is called by all three methods.

bliss/encoder/plotting.py
georgeyfly: Oh, this file is also modifed by undergraduates. I will delete all the changes made by them.

case_studies/redshift/network_rs.py
```diff
@@ -0,0 +1,99 @@
...
+            nn.ReLU(),
+            # nn.Dropout(dropout_rate),
+            nn.Linear(hidden_dim, out_dim),
+        )
+
+    def forward(self, x):
+        return self.net(x)
+
+
+class LitRegressor(pl.LightningModule):
```
jeff-regier, line 50: It'd be better to merge `Regression` and `LitRegression` into a single subclass of `pl.LightningModule`

case_studies/redshift/train_rs.py
```diff
@@ -0,0 +1,191 @@
...
+import re
+
+import click
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import pytorch_lightning as pl
+import torch
+import torch.nn as nn
+from network_rs import LitRegressor, Regressor
```
jeff-regier, line 15: Would you merge train_rs.py and network_rs.py into a single file, to make it easier for others to understand what code in this directory are related? And also name the file something like `photoz_from_fluxes.py` (network_rs is somewhat cryptic).

bliss/encoder/plotting.py
jeff-regier: Sounds good. Let's add their code in our next PR, which will address issue #976

jeff-regier:
Please ignore my earlier comment (now deleted) about merging two files. Instead, how about adding a simple shell script that calls all the programs in this directory, in order, to demonstrate the complete workflow? For example:
https://github.com/prob-ml/bliss/blob/master/case_studies/multiband/multiband_exps.sh

A script like that would help others to recreate your results.

PR #995, opened 2024-05-20
Plotting workflow changes & corresponding updates

shreyasc30:
new PlotCollection class and moved weak lensing specific plots to lensing case study

also added weak_lensing notebook (looking at cosmo + dc2 combination v.a.v. shear and convergence)


jeff-regier (changes requested):
This is nice and it's just as we discussed. Now that I see it though, I'm more convinced that `PlotCollection` is duplicating `MetricCollection`. I think we can do it all just by adding new metrics that make plots, and that it would be cleaner this latter way. With this approach, to make a plot based on the first batch of image data, we'd create a metric whose `update` method returns immediately if this isn't the first batch of the epoch (according to its internal state, which gets reset each epoch). If it is the first batch, then it saves the batch and toggles the first batch indicator. The actual plotting wouldn't have to happen until the end of the epoch, when the rest of the metrics also get reported.

shreyasc30:
> This is nice and it's just as we discussed. Now that I see it though, I'm more convinced that `PlotCollection` is duplicating `MetricCollection`. I think we can do it all just by adding new metrics that make plots, and that it would be cleaner this latter way. With this approach, to make a plot based on the first batch of image data, we'd create a metric whose `update` method returns immediately if this isn't the first batch of the epoch (according to its internal state, which gets reset each epoch). If it is the first batch, then it saves the batch and toggles the first batch indicator. The actual plotting wouldn't have to happen until the end of the epoch, when the rest of the metrics also get reported.

attempted to address in da56157882283fe59d00d1cd22fa7d571b41783a and ad260ac614daadbd509e38ba0f5124ab9f1e3ed5

Note: still not ready to merge, have to perform similar refactor on case_studies/weak_lensing_plots if this looks good

bliss/encoder/encoder.py
```diff
@@ -271,51 +272,46 @@ def update_metrics(self, batch):
...
-        target_cat_cropped = target_cat.symmetric_crop(self.tiles_to_crop)
-        est_cat = self.sample(batch, use_mode=True)
-        mp = self.tiles_to_crop * self.tile_slen
-        fig = plot_detections(batch["images"], target_cat_cropped, est_cat, margin_px=mp)
-        title = f"Epoch:{self.current_epoch}/{logging_name} images"
-        if self.logger:
-            self.logger.experiment.add_figure(title, fig)
-        plt.close(fig)
+        self.plots.update(
+            self.tile_slen,
```
jeff-regier, line 276: Better I think if `tile_slen` is passed to this plotting metric in the constructor (through a hyrda). Because it will be the same each time. And that's how we do it for other classes where the tile_slen is needed.

jeff-regier: Same for `min_flux_threshold` and `tiles_to_crop`, I think

jeff-regier: e.g. instantiate this metric class with `${encoder.tile_slen}` as an argument in the hydra config

bliss/conf/base_config.yaml
```diff
@@ -104,6 +104,13 @@ encoder:
           - _target_: bliss.encoder.metrics.FluxError
             survey_bands: ${encoder.survey_bands}
           - _target_: bliss.encoder.metrics.GalaxyShapeError
+    plots:
+        _target_: torchmetrics.MetricCollection
+        metrics:
+          - _target_: bliss.encoder.plots.PlotSampleImages
+            frequency: 1
+            restrict_batch: 0
```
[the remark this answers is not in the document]
shreyasc30: considered this, but figured we may want to keep them separate for readability as they do serve different functions - i.e. PlotSampleImages is decidedly not a metric. 
And the plots have different functions applied to them in the encoder as well v.a.v. update, etc. which would make it difficult to merge the lists together (self.plots.update, self.report with custom plots text, etc.)
Happy to merge them together and then include a flag in the class def. but then would have to filter by that flag in the encoder anyways to apply the relevant methods, so this may indeed be easier, but will do either way

bliss/encoder/encoder.py
```diff
@@ -271,51 +272,46 @@ def update_metrics(self, batch):
...
+            batch,
+            self.min_flux_threshold,
+            self.tiles_to_crop,
+            self.sample,
```
jeff-regier, line 280: We already have drawn a `sample_cat` in this function. Can we pass that instead of `self.sample`?

bliss/conf/base_config.yaml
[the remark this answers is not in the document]
jeff-regier: They may start to look more similar if things like `min_flux_threshold` and `tiles_to_crop` are passed to the constructor rather than to `update(...)`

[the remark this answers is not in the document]
jeff-regier: That seems better than calling one group of metrics "plots" and another "metrics". Because some of the latter do have plotting methods. And all of the former are literally metrics (in terms of their type).

bliss/encoder/encoder.py
shreyasc30: addressed in 6e25cbbaabeb67023d89586940d5abebb6d124fc

bliss/encoder/encoder.py
shreyasc30: addressed in 6e25cbbaabeb67023d89586940d5abebb6d124fc

bliss/conf/base_config.yaml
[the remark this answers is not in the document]
shreyasc30: perhaps a better name for the plots is in order; given that they do different things, I feel that they should be enumerated separately; combining the two would have, say, RMSE and PlotImages in the same list, which (other than sharing a base class) don't have much similarity in terms of true functionality... 

I will make the filtering change if you feel that's optimal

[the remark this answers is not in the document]
shreyasc30: addressed by 60375523cbc3da564a48304c723568d91f071b94

shreyasc30:
> > This is nice and it's just as we discussed. Now that I see it though, I'm more convinced that `PlotCollection` is duplicating `MetricCollection`. I think we can do it all just by adding new metrics that make plots, and that it would be cleaner this latter way. With this approach, to make a plot based on the first batch of image data, we'd create a metric whose `update` method returns immediately if this isn't the first batch of the epoch (according to its internal state, which gets reset each epoch). If it is the first batch, then it saves the batch and toggles the first batch indicator. The actual plotting wouldn't have to happen until the end of the epoch, when the rest of the metrics also get reported.
> 
> attempted to address in [da56157](https://github.com/prob-ml/bliss/commit/da56157882283fe59d00d1cd22fa7d571b41783a) and [ad260ac](https://github.com/prob-ml/bliss/commit/ad260ac614daadbd509e38ba0f5124ab9f1e3ed5)
> 
> Note (cc @timwhite0) : still not ready to merge, have to perform similar refactor on case_studies/weak_lensing/weak_lensing_plots if this looks good

weak_lensing_plots refactor addressed by 

> > This is nice and it's just as we discussed. Now that I see it though, I'm more convinced that `PlotCollection` is duplicating `MetricCollection`. I think we can do it all just by adding new metrics that make plots, and that it would be cleaner this latter way. With this approach, to make a plot based on the first batch of image data, we'd create a metric whose `update` method returns immediately if this isn't the first batch of the epoch (according to its internal state, which gets reset each epoch). If it is the first batch, then it saves the batch and toggles the first batch indicator. The actual plotting wouldn't have to happen until the end of the epoch, when the rest of the metrics also get reported.
> 
> attempted to address in [da56157](https://github.com/prob-ml/bliss/commit/da56157882283fe59d00d1cd22fa7d571b41783a) and [ad260ac](https://github.com/prob-ml/bliss/commit/ad260ac614daadbd509e38ba0f5124ab9f1e3ed5)
> 
> Note (cc @timwhite0) : still not ready to merge, have to perform similar refactor on case_studies/weak_lensing/weak_lensing_plots if this looks good

refactor weak_lensing_plots addressed by c5b8b83aa954b74cd326d1e337d67375baa94c03

case_studies/dc2/dc2_shear_conv.ipynb
```diff
@@ -0,0 +1,764 @@
+{
```
jeff-regier, line 1: We should keep this notebook in a separate case study because @YicunDuanUMich is working on this one. Perhaps the current DC2 case study could be renamed something like `dc2_cataloging`.

jeff-regier (approved):
Looks good to me. Please move `dc2_shear_conv.ipynb` to  different case study, perhaps `weak_lensing`, before merging.

PR #998, opened 2024-05-31
Detect galaxy cluster members in GalSim generated images

### LINKED ISSUE #992: Detect galaxy clusters in GalSim images using the BLISS encoder

kapnadak:
- Wrapped up data generation into a single bash script
- Wrote custom variational distribution
- Wrote metrics for cluster membership (accuracy, precision, recall, F1)
- Modified network to include successive downsampling (to allow for tile sizes larger than 4)
- Set-up a complete pipeline from data generation to training and testing
- Resolves #992 

@gapatron


kapnadak:
Changes made to main BLISS directory:

1. `image_normalizer.py`: added a check to ensure `self.log_transform_stddevs` is non-empty before iterating
2. `encoder.py`: remove assertion requiring tile length to be 2 or 4, and calculate number of downsampling layers required
3. `convnet.py`: add successive downsampling layers to support tile lengths of greater than 4
4. `catalog.py`: added a few allowed parameters

bliss/conf/base_config.yaml
```diff
@@ -109,8 +109,6 @@ encoder:
             flux_error:
                 _target_: bliss.encoder.metrics.FluxError
                 survey_bands: ${encoder.survey_bands}
-            galaxy_shape_error:
-                _target_: bliss.encoder.metrics.GalaxyShapeError
```
jeff-regier, line 113: Can you override this in your case-study-specific config file, rather than changing the base config file?

bliss/encoder/convnet.py
```diff
@@ -66,7 +66,8 @@ def __init__(self, n_bands, ch_per_band, num_features, double_downsample=True):
             nn.Sequential(*[ConvBlock(64, 64, kernel_size=5, padding=2) for _ in range(4)]),
             ConvBlock(64, 128, stride=2),
             nn.Sequential(*[ConvBlock(128, 128) for _ in range(5)]),
-            ConvBlock(128, num_features, stride=(2 if double_downsample else 1)),  # 4
+            nn.Sequential(*[ConvBlock(128, 128, stride=2) for _ in range(num_downsample)]),
+            ConvBlock(128, num_features, stride=1),  # 4
```
jeff-regier, line 70: A potential issue here is that the downsampling doesn't happen until several layers into the network, which sometimes is what you want, but other times isn't. For high-resolution images, you likely want to downsample right away (and to set the kernel size to 5 rather than 3), to reduce memory usage.

bliss/encoder/encoder.py
```diff
@@ -1,6 +1,7 @@
 from copy import copy
 from typing import Optional
 
+import numpy as np
```
jeff-regier, line 4: can we use torch instead to avoid an import?

bliss/conf/base_config.yaml
kapnadak: Resolved

bliss/encoder/encoder.py
```diff
@@ -80,14 +81,15 @@ def __init__(
         self.double_detect = double_detect
         self.use_checkerboard = use_checkerboard
 
+        num_downsample = int(np.log2(self.tile_slen)) - 1
```
jeff-regier, line 84: Would you assert that `tile_size` is a power of 2 here (or somewhere)?

Also, probably a good idea to round the log of the tile size to the nearest integer before truncating, in case of slight floating point imprecision

bliss/encoder/encoder.py
```diff
@@ -80,14 +81,15 @@ def __init__(
...
+
         ch_per_band = self.image_normalizer.num_channels_per_band()
-        assert tile_slen in {2, 4}, "tile_slen must be 2 or 4"
```
jeff-regier, line 84: Would you assert that tile_size is a power of 2 here?

bliss/encoder/encoder.py
kapnadak: Done

poetry.lock
```diff
@@ -592,6 +592,17 @@ files = [
     {file = "cached_property-1.5.2-py2.py3-none-any.whl", hash = "sha256:df4f613cf7ad9a588cc381aaf4a512d26265ecebd5eb9e1ba12f1319eb85a6a0"},
 ]
 
+[[package]]
```
jeff-regier, line 595: Did you mean to change the poetry.lock file? (Are these changes necessary?)

bliss/encoder/encoder.py
kapnadak: Done

bliss/encoder/convnet.py
kapnadak: Pushed all the downsampling layers to the front and increased kernel size to 5. Please let me know if this works!

jeff-regier: Unfortunately not, because this change will break other case studies. Maybe two binary flags is the way to go: one for down sampling here and the other for down sampling right away

kapnadak: Do we want to allow the possibility of downsampling at both places? or just one of the two depending on the flag?

jeff-regier: I'm not really sure, either way seems OK for now

kapnadak: I added a `downsample_at_front` flag. Seems like the network would have one common backbone, and the flag either puts the downsampling layers at the front or the back, followed by upsampling back to `num_features`. Default value is `false` based on what works for other case studies I think. Added the flag to hydra config as well.

kapnadak: Resolved for now, things to look into next:
- Upscale number of channels per downsampling layer
- Experiment with attention blocks
- Intersperse successive downsampling layers with C3 blocls

Issue #981, opened 2024-05-06
DC2 performance plot: difference of differences


jeff-regier:
Figures 2 & 3 of the DC2 manuscript have one line for BLISS and one line for LSST (the competitor). I had in mind creating a new plot (in addition to Figures 2 & 3) that shows the difference in performance between BLISS and LSST (for each flux bin). Then we'd add error bars to this plot. Ideally these error bars would illustrate that BLISS has lower than LSST (with high confidence) at every magnitude.

Issue #974, opened 2024-05-06
DC2 classification performance stratified by flux


jeff-regier:
Figure 7 in the DC2 manuscript shows BLISS performance for DC2. However, as is, this figure reports average performance over all the light sources. It would more more useful to report performance as a function of magnitude/flux (like how it's done in Figures 2 & 3).

To close this issue, we might create a new `Metric` class, like the ones here:
https://github.com/prob-ml/bliss/blob/master/bliss/encoder/metrics.py

jeff-regier:
Here's an example of a nice plot showing star/galaxy classification performance as a function of flux (from Ismael's paper):
<img width="441" alt="Screenshot 2024-05-06 at 4 29 39 PM" src="https://github.com/prob-ml/bliss/assets/8953355/26c9692c-f192-4535-bbeb-8de0c91bab2a">

Issue #973, opened 2024-05-06
better DC2 precision & recall for the dimmest light sources


jeff-regier:
Figures 2 & 3 in the DC2 manuscript show BLISS being outperformed by the LSST pipeline for the dimmest light sources. We should confirm that these numbers are correct by rerunning everything. If they are correct, we should improve BLISS's image normalization for these dim light sources.

YicunDuanUMich:
roger that

jeff-regier:
If we need to improve BLISS's image normalization, the arcsinh transform (`torch.arcsinh`) may give us better results than log, which is what we use here:
https://github.com/prob-ml/bliss/blob/3b79b7e057777bcc504331b1fa84fd33ac3c6d31/bliss/encoder/image_normalizer.py#L93

Fiddling with some of the thresholds may help too. Currently the thresholds we use for log transforms are set here:
https://github.com/prob-ml/bliss/blob/3b79b7e057777bcc504331b1fa84fd33ac3c6d31/bliss/conf/base_config.yaml#L86

YicunDuanUMich:
![image](https://github.com/prob-ml/bliss/assets/144750316/22f50d84-64be-463c-8659-3f5dae3dd67b)

YicunDuanUMich:
we can now outperform lsst

PR #999, opened 2024-05-31
Yd/dc2 new plots

YicunDuanUMich:
address issue #981, #974 and #973


bliss/encoder/image_normalizer.py
```diff
@@ -150,3 +173,32 @@ def clahe(cls, images, min_stdev, kernel_size=9, padding=4):
         # Output rolling z-score
         normalized_img = res_img / torch.clamp(stdev, min=min_stdev)
         return normalized_img.unsqueeze(2)
+
+    # just for debug
```
jeff-regier, line 177: Please remove debugging code

bliss/encoder/image_normalizer.py
```diff
@@ -106,17 +130,16 @@ def get_input_tensor(self, batch):
         if self.use_clahe:
             renormalized_img = self.clahe(raw_images, self.clahe_min_stdev)
             inputs.append(renormalized_img)
-            inputs[0] = self.clahe(backgrounds, self.clahe_min_stdev)
+            if not self.asinh_params:
```
jeff-regier, line 133: This is also surprising behavior (from the perspective of user). If the user sets `use_clahe` to true, they would expect bliss to indeed use clahe normalization.

bliss/encoder/image_normalizer.py
```diff
@@ -106,17 +130,16 @@ def get_input_tensor(self, batch):
...
 
         if self.asinh_params:
-            backgrounds = backgrounds.clamp(min=1e-6)
-            for threshold in self.asinh_params["thresholds"]:
-                inputs.append(
-                    torch.asinh(
-                        ((raw_images - backgrounds) / backgrounds.sqrt() - threshold)
-                        * self.asinh_params["scale"],
-                    )
+            inputs.append(
```
jeff-regier, line 137: Statements that span multiple lines using parentheses are hard to read. (Better not to rely on the `black` formatter to reformat long lines.) Can you define some intermediate variables instead?

bliss/encoder/image_normalizer.py
```diff
@@ -106,17 +130,16 @@ def get_input_tensor(self, batch):
...
+                torch.asinh(
+                    (raw_images - self.asinh_thresholds_tensor)
+                    * torch.exp(self.asinh_scales_tensor),
```
jeff-regier, line 140: Statements that span multiple lines using parentheses are hard to read. (Better not to rely on the `black` formatter to reformat long lines.) Can you define some intermediate variables instead?

bliss/encoder/image_normalizer.py
```diff
@@ -39,12 +43,31 @@ def __init__(
...
         self.asinh_params = asinh_params
 
+        if self.asinh_params:
+            thresholds_num = len(self.asinh_params["thresholds"])
+            thresholds = (
+                torch.tensor(self.asinh_params["thresholds"])
+                .view(1, -1)
+                .expand(len(self.bands), thresholds_num)
+                .view(1, len(self.bands), thresholds_num, 1, 1)
+            ).clone()
```
jeff-regier, line 53: Statements that span multiple lines using parentheses are hard to read. (Better not to rely on the `black` formatter to reformat long lines.) Can you define some intermediate variables instead?

bliss/encoder/image_normalizer.py
```diff
@@ -80,7 +104,7 @@ def get_input_tensor(self, batch):
 
         raw_images = batch["images"][:, self.bands].unsqueeze(2)
         backgrounds = batch["background"][:, self.bands].unsqueeze(2)
-        inputs = [backgrounds]
+        inputs = [] if self.asinh_params else [backgrounds]
```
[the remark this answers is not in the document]
YicunDuanUMich: Maybe not everyone wants to use our new asinh normalization. Qiaozhi told me that the flux error is still high even if we use asinh.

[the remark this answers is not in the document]
YicunDuanUMich: i'll add a binary flag for this

bliss/encoder/variational_dist.py
```diff
@@ -153,10 +154,10 @@ def compute_nll(self, true_tile_cat: TileCatalog):
...
-        galsim_true_vals = rearrange(true_tile_cat["galaxy_params"], "b ht wt 1 d -> b ht wt d")
-        for i, param_name in enumerate(self.GALSIM_NAMES):
-            galsim_pn = f"galsim_{param_name}"
-            loss_term = -q[galsim_pn].log_prob(galsim_true_vals[..., i] + 1e-9) * true_gal_bools
-            loss += loss_term
+        # galsim_true_vals = rearrange(true_tile_cat["galaxy_params"], "b ht wt 1 d -> b ht wt d")
+        # for i, param_name in enumerate(self.GALSIM_NAMES):
+        #     galsim_pn = f"galsim_{param_name}"
+        #     loss_term = -q[galsim_pn].log_prob(galsim_true_vals[..., i] + 1e-9) * true_gal_bools
+        #     loss += loss_term
```
jeff-regier, line 161: Our flake8 checks don't allow commented out code. If it's problematic to estimate these galaxy parameters for our DC2 benchmarking, we could pass a binary flag to the constructor of this class (set in our hydra configs) to control whether galaxy params are estimated

bliss/surveys/dc2.py
```diff
@@ -293,6 +186,160 @@ def test_dataloader(self):
...
+        ) = load_image_and_catalog(
+            image_idx,
+            self.image_files,
+            self.bg_files,
+            self.image_lim,
+            self.bands,
+            self.n_bands,
+            self.cat_path,
+            self.min_flux_threshold,
+        )
```
jeff-regier, line 211: So many arguments/return values/lines. Can this be refactored somehow?

bliss/surveys/dc2.py
```diff
@@ -293,6 +186,160 @@ def test_dataloader(self):
...
+    ) = load_image_and_catalog(
+        image_index,
+        self_copy["image_files"],
+        self_copy["bg_files"],
+        self_copy["image_lim"],
+        self_copy["bands"],
+        self_copy["n_bands"],
+        self_copy["cat_path"],
+        self_copy["min_flux_threshold"],
+    )
```
jeff-regier, line 287: So many arguments/return values/lines. Can this be refactored somehow?

bliss/surveys/dc2.py
```diff
@@ -253,6 +145,7 @@ def prepare_data(self):  # noqa: WPS324
                             "bands": self.bands,
                             "n_split": self.n_split,
                             "split_file_path": self.split_file_path,
+                            "min_flux_threshold": self.min_flux_threshold,
```
jeff-regier, line 148: This deeply nested code is somewhat hard to read. Can you define an intermediate variable instead?

bliss/encoder/variational_dist.py
```diff
@@ -1,3 +1,4 @@
+# flake8: noqa: E800
```
jeff-regier, line 1: Please try not to disable the flake8 checks, unless there's no good way around them

jeff-regier: If the flake8 checks are misguided, I'll usually disable them project-wide by adding them to `.flake8`. But E800 is really good one to have around

bliss/encoder/image_normalizer.py
```diff
@@ -1,6 +1,10 @@
+# flake8: noqa: WPS348
 import warnings
+from pathlib import Path
 from typing import Dict
 
+import matplotlib.pyplot as plt  # noqa: WPS301
```
jeff-regier, line 6:
```
from matplotlib import pyplot as plt
```

bliss/encoder/image_normalizer.py
```diff
@@ -1,6 +1,10 @@
+# flake8: noqa: WPS348
```
jeff-regier, line 1: WPS348 is a helpful convention. The black formatter has its moments, but for long lines, it's better to fix them manually.

bliss/surveys/dc2.py
```diff
@@ -293,6 +186,160 @@ def test_dataloader(self):
...
+        cat_path, wcs, height, width, bands, min_flux_threshold
+    )
+    tile_dict = full_cat.to_tile_catalog(4, 5).get_brightest_sources_per_tile().to_dict()
+    tile_dict = squeeze_tile_dict(tile_dict)
+    tile_dict["star_fluxes"] = tile_dict["star_fluxes"].clamp(min=1e-18)
+    tile_dict["galaxy_fluxes"] = tile_dict["galaxy_fluxes"].clamp(min=1e-18)
+    tile_dict["galaxy_params"][..., 3] = tile_dict["galaxy_params"][..., 3].clamp(min=1e-18)
+    tile_dict["galaxy_params"][..., 5] = tile_dict["galaxy_params"][..., 5].clamp(min=1e-18)
+
+    return image, bg, tile_dict, full_cat, wcs, wcs_header_str, match_id, psf_params  # noqa: WPS227
```
jeff-regier, line 264: Using a dict is one way to bundle together parameters, ideally into meaningful groups, rather than just putting them all together to evade the flake8 check

bliss/encoder/image_normalizer.py
YicunDuanUMich: have removed them

bliss/encoder/image_normalizer.py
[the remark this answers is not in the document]
YicunDuanUMich: add the flag `include_background`

bliss/encoder/image_normalizer.py
YicunDuanUMich: have fixed it

bliss/encoder/image_normalizer.py
YicunDuanUMich: use a new intermediate variable `processed_raw_images`

bliss/encoder/image_normalizer.py
YicunDuanUMich: use intermediate variables to simplify it

bliss/encoder/variational_dist.py
YicunDuanUMich: add flag `include_galaxy_property_loss`

bliss/surveys/dc2.py
YicunDuanUMich: return a dict to address this

bliss/surveys/dc2.py
YicunDuanUMich: return a dict to address this

bliss/surveys/dc2.py
YicunDuanUMich: have fixed it

case_studies/dc2/DC2_bootstrap_generate_data.py
```diff
@@ -0,0 +1,177 @@
...
+                {
+                    "plocs": cur_plocs.clone(),
+                    "source_type": cur_source_type.clone(),
+                    "flux": cur_flux.clone(),
+                },
+                output_file,
+            )
+
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
YicunDuanUMich: update the `run_dc2_exp.sh`

bliss/encoder/variational_dist.py
YicunDuanUMich: delete noqa: E800

bliss/encoder/image_normalizer.py
YicunDuanUMich: plt is not used in this code if we delete the `plot_image` function which is designed for debug

bliss/encoder/image_normalizer.py
YicunDuanUMich: delete noqa: WPS348

bliss/surveys/dc2.py
YicunDuanUMich: have fixed it

case_studies/dc2/DC2_bootstrap_generate_data.py
```diff
@@ -0,0 +1,177 @@
+# pylint: skip-file
+# flake8: noqa
```
jeff-regier, line 2: Better not to disable these checks

bliss/catalog.py
```diff
@@ -722,3 +722,35 @@ def to_astropy_table(self, encoder_survey_bands: Tuple[str]) -> Table:
...
+        d = {}
+        new_max_sources = plocs_mask.sum(dim=1).max()
+        for k, v in self.to_dict().items():
+            if k == "n_sources":
+                d[k] = plocs_mask.sum(dim=1)
+            else:
+                d[k] = v[
+                    plocs_full_mask_sorted_indexes[:, 0].tolist(),
+                    plocs_full_mask_sorted_indexes[:, 1].tolist(),
+                ].view(-1, self.max_sources, v.shape[-1])[:, :new_max_sources, :]
```
jeff-regier, line 752: This code seems cryptic to me. Can you use some intermediate variables to better explain what it's doing? (Is a tensor getting converted to a Python list and then back to a tensor?)

bliss/encoder/encoder.py
```diff
@@ -36,6 +37,7 @@ def __init__(
         sample_image_renders: MetricCollection,
         matcher: CatalogMatcher,
         min_flux_threshold: float = 0,
+        min_flux_threshold_during_test: float = 0,
```
jeff-regier, line 40: Did you mean to change this variable name?

bliss/encoder/encoder.py
```diff
@@ -1,3 +1,4 @@
+# flake8: noqa: WPS348
```
jeff-regier, line 1: WPS348 is a good check, please don't disable it

bliss/encoder/metrics.py
```diff
@@ -174,10 +174,64 @@ def compute(self):
...
+            }
+
+            return {
+                "detection_precision": precision,
+                "detection_recall": recall,
+                "detection_f1": f1,
+                **precision_bin_results,
+                **recall_bin_results,
+                **f1_bin_results,
+            }
```
jeff-regier, line 206: Having a multiple return statements in a function is something of an anti-pattern. Please refactor this function to avoid duplicating code (the logic in the if block and afterwards seems about the same).

bliss/encoder/variational_dist.py
```diff
@@ -11,11 +11,12 @@
 
 
 class VariationalDistSpec(torch.nn.Module):
-    def __init__(self, survey_bands, tile_slen):
+    def __init__(self, survey_bands, tile_slen, include_galaxy_property_loss: bool = True):
         super().__init__()
 
         self.survey_bands = survey_bands
         self.tile_slen = tile_slen
+        self.include_galaxy_property_loss = include_galaxy_property_loss
```
jeff-regier, line 19: If `include_galaxy_property_loss` if false, if would be better to omit galaxy-specific variables from `self.factor_specs`.

bliss/encoder/encoder.py
YicunDuanUMich: `min_flux_threshold` is used during training. `min_flux_threshold_during_test` is used to filter true_cat and our prediction during test. Now, we set both of them to 100.

case_studies/dc2/DC2_bootstrap_generate_data.py
```diff
@@ -0,0 +1,177 @@
...
+            [
+                torch.tensor(flux.values).view(-1, 1)
+                for flux in [
+                    lsst_catalog_df["cModelFlux_g"],
+                    lsst_catalog_df["cModelFlux_i"],
+                    lsst_catalog_df["cModelFlux_r"],
+                    lsst_catalog_df["cModelFlux_u"],
+                    lsst_catalog_df["cModelFlux_y"],
+                    lsst_catalog_df["cModelFlux_z"],
+                ]
```
jeff-regier, line 142: Please use an intermediate variable here, to avoid deep nesting, which is hard to read

case_studies/dc2/DC2_bootstrap_generate_data.py
```diff
@@ -0,0 +1,177 @@
...
+            image_height_index=split_result["image_height_index"],
+            image_width_index=split_result["image_width_index"],
+        )
+
+        with open(output_path / ("lsst_" + split_id), "wb") as output_file:
+            torch.save(
+                {
+                    "plocs": cur_plocs.clone(),
+                    "source_type": cur_source_type.clone(),
+                    "flux": cur_flux.clone(),
```
jeff-regier, line 77: Please use an intermediate variable here, to avoid deep nesting

bliss/catalog.py
YicunDuanUMich: I'm trying to do the following operation.
```
>>> t.shape
torch.Size([4, 5, 2])
>>> t[[0, 1], [3, 4]]
tensor([[0.9178, 0.3564],
        [0.4694, 0.9806]])
>>> t[[1, 2], [3, 4]]
tensor([[0.0147, 0.4497],
        [0.2121, 0.2853]])
>>> 
```

YicunDuanUMich: the tensor `plocs_full_mask_sorted_indexes` contain such indexes, but i find it will throw error if i use `v[plocs_full_mask_sorted_indexes]`.

bliss/encoder/variational_dist.py
YicunDuanUMich: if we remove the galaxy-specific variables, it may require us to modify the neural network architecture. In `_parse_factors` function, we parse the input according to `self.factor_specs`.

bliss/catalog.py
jeff-regier: OK I see

bliss/catalog.py
```diff
@@ -722,3 +722,35 @@ def to_astropy_table(self, encoder_survey_bands: Tuple[str]) -> Table:
         est_cat_table.remove_column("galaxy_params")
 
         return hstack([est_cat_table, galaxy_params_table])
+
+    def filter_full_catalog_by_ploc_box(self, box_origin: torch.Tensor, box_len: float):
```
jeff-regier, line 726: Could you add a unit test of this function to `test_catalogs.py`?

bliss/encoder/encoder.py
jeff-regier: I see. It's so unusual to have different thresholds during training and validation/testing. With different thresholds, you can't even compare training loss and validation loss to get an idea of how much you're overfitting, right?

jeff-regier: Could we filter the catalogs when computing validation and testing metrics, rather than when we compute loss?

YicunDuanUMich: yes. i design this flag according to our previous discussion. we may face the case where we need to perform training using data with flux over 90 and test the model using data of higher flux (for example, 100). if we don't need to test this case in future, we can remove it from this PR.

YicunDuanUMich: > Could we filter the catalogs when computing validation and testing metrics, rather than when we compute loss?

do you mean we need to train the model with unfiltered data?

bliss/surveys/dc2.py
```diff
@@ -293,6 +183,161 @@ def test_dataloader(self):
...
+    return tile_dict_copy
+
+
+def unsqueeze_tile_dict(tile_dict):
+    tile_dict_copy = copy.copy(tile_dict)
+    for k, v in tile_dict_copy.items():
+        if k != "n_sources":
+            tile_dict_copy[k] = rearrange(v, "h w nh nw -> 1 h w nh nw")
+    tile_dict_copy["n_sources"] = rearrange(tile_dict_copy["n_sources"], "h w -> 1 h w")
+    return tile_dict_copy
```
jeff-regier, line 226: Is there anyway to avoid squeezing and then unsqueezing these tensors? i.e., Can we just index them according to their original format?

bliss/surveys/dc2.py
```diff
@@ -293,6 +183,161 @@ def test_dataloader(self):
...
+    )
+
+    image = result_dict["inputs"]["image"]
+    bg = result_dict["inputs"]["bg"]
+    tile_dict = result_dict["tile_dict"]
+    wcs_header_str = result_dict["other_info"]["wcs_header_str"]
+    psf_params = result_dict["inputs"]["psf_params"]
+
+    # split image
+    split_lim = self_copy["image_lim"][0] // self_copy["n_split"]
```
jeff-regier, line 288: Could you add a small test of the new code to `test_dc2.py`? It wouldn't need to assert anything: it could just run the code to check for syntax errors. (A "smoke test".) To make the test work, you may need to add a few representative DC2 files to the `data/tests/dc2` directory.

You can see all the code that currently isn't exercised by pytest here:
https://app.codecov.io/gh/prob-ml/bliss/blob/yd%2Fdc2_new_plots/bliss%2Fsurveys%2Fdc2.py

bliss/encoder/variational_dist.py
jeff-regier: The network architecture should automatically adapt if you remove them. The method `n_params_per_source` counts the number of parameters for whatever factors are present in `self.factor_specs`, and adjusts the output dimension of the network accordingly.

bliss/encoder/encoder.py
jeff-regier: I think we should use the same threshold for computing training loss and validation/testing loss (e.g., 90). But use a potentially different threshold for computing the validation metrics (e.g., 100). Does that make sense? The threshold for metrics could be an argument passed to the metric even rather than an argument of the encoder.

jeff-regier: If you set `exclude_last_bin` to True for `DetectionMetrics`, and the last bin cutoff is 100 (converted to magnitude), then you're effectively doing this.

YicunDuanUMich: yes, this is what we did in our current code, except that we filter the objects in encoder. Conducting this step in encoder may simplify the design logic. if we pass this argument to metric, we may need to modify the interface and code for each metric class.

jeff-regier: OK I get it now :-)

bliss/surveys/dc2.py
jeff-regier: OK, got it

bliss/surveys/dc2.py
YicunDuanUMich: if we can run the `prepare_data` function of DC2, we can cover most of the code. however, in previous PR, i found that (probably) the `multiprocessing` called by `generate_split_file` can't run in our test server. it will lead to dead lock. I'll try to bypass this limit

jeff-regier: Maybe we could set the number of processes to launch in the hydra config file, and we could set it to 1 in the testing config?

bliss/encoder/variational_dist.py
YicunDuanUMich: professor, could we have a face-to-face discussion about this?

bliss/surveys/dc2.py
YicunDuanUMich: i'll try it

YicunDuanUMich:
The test fails because we didn't upload the config file, i'll upload it tonight

jeff-regier (changes requested):
In addition to the items above, would you rename the DC2 case study directory to `dc2_cataloging` as part of this PR? This change will help us to distinguish your case study from Shreyas's case study, which also uses DC2, but not for cataloging.

case_studies/dc2/DC2_bootstrap_generate_data.py
YicunDuanUMich: have fixed it

bliss/encoder/encoder.py
YicunDuanUMich: fixed it

bliss/encoder/metrics.py
YicunDuanUMich: fixed it

case_studies/dc2/DC2_bootstrap_generate_data.py
YicunDuanUMich: fixed it

case_studies/dc2/DC2_bootstrap_generate_data.py
YicunDuanUMich: fixed it

bliss/catalog.py
YicunDuanUMich: I wrote a new test `test_filter_full_catalog_by_ploc_box` for it

jeff-regier (approved):
Looks good, thanks for the revisions

PR #1001, opened 2024-06-04
clean up

jeff-regier:
This PR removes some case studies and files we aren't using any more. It also removes the `data` directory from bliss. Now we have a `tests/data` subdirectory instead, which stores only the data needed for the tests to run. Any other data is stored on the GPU server under `/data/scratch`. `base_config.yaml` and `testing_config.yaml` reflect this new organization.


bliss/conf/base_config.yaml
```diff
@@ -12,14 +12,13 @@ hydra:
...
-    dc2: ${paths.data}/dc2
+
+    sdss: /data/scratch/sdss
+    decals: /data/scratch/decals
+    des: /data/scratch/des
+    dc2: /data/scratch/dc2local
+
     output: ${paths.root}/output
-    pretrained_models: ${paths.data}/pretrained_models
-    dc2_catalog: ${paths.root}/case_studies/dc2_cataloging/merged_catalog
```
jeff-regier, line 22: @YicunDuanUMich I put a copy of the merged catalog under `/data/scratch/dc2local`. I also made a smaller version of this file, for use during testing, which I placed in `tests/data/dc2`.

YicunDuanUMich: get it

PR #1002, opened 2024-06-04
Detect galaxy cluster members in GalSim generated images.

### LINKED ISSUE #992: Detect galaxy clusters in GalSim images using the BLISS encoder

kapnadak:
Wrapped up data generation into a single bash script
Wrote custom variational distribution
Wrote metrics for cluster membership (accuracy, precision, recall, F1)
Modified network to include successive downsampling (to allow for tile sizes larger than 4)
Set-up a complete pipeline from data generation to training and testing
Resolves #992 

@gapatron


bliss/encoder/convnet.py
```diff
@@ -52,26 +52,81 @@ def forward(self, x):
...
+            self.backbone.append(ConvBlock(128, num_features, stride=1))
+        elif tile_slen == 4:
+            self.backbone.append(ConvBlock(nch_hidden, 64, kernel_size=5, padding=2))
+            self.backbone.append(
+                nn.Sequential(*[ConvBlock(64, 64, kernel_size=5, padding=2) for _ in range(4)])
+            )
+            self.backbone.append(ConvBlock(64, 128, stride=2))
+            self.backbone.append(nn.Sequential(*[ConvBlock(128, 128) for _ in range(5)]))
+            self.backbone.append(ConvBlock(128, num_features, stride=2))
+        else:
```
jeff-regier, line 90: @kapnadak @gapatron  Now that I'm see this, I think it'd be better if we left `convnet.py` unmodified for your case study. Instead, we can define a new version of `FeaturesNet` within your case study directory, which can be modified without affecting other case studies. PR #1003 should make it possible to swap in arbitrary CNN architectures by defining a subclass of `Encoder`.

PR #1004, opened 2024-06-05
sample from truncated normal without rejection sampling

jacksonloper:
I couldn't find a TruncatedNormal in torch.distribitutions.  There was a truncnormal in torch.nn.init, but it couldn't handle multiple means and stdevs.  So I just sampled with inverse cdf.


jeff-regier (approved):
Great!

jacksonloper:
It would be nice to somehow test that it is sampling from the right distribution :).  Currently have no tests for that.  I tested myself and looked at histograms and they looked right, but I wonder if there's another way.

PR #1006, opened 2024-06-06
Refactor dc2 cataloging

YicunDuanUMich:
* move all dc2 cataloging related code to case_studies/dc2_cataloging/utils

* update SourceTypeAccuracy to report classification accuracy for each flux bin

* update VariationalDistSpec and VariationalDist to make the override convenient

* update config files and experiment scripts

---------


bliss/encoder/image_normalizer.py
```diff
@@ -94,9 +66,6 @@ def get_input_tensor(self, batch):
         assert batch["images"].size(2) % 16 == 0, "image dims must be multiples of 16"
         assert batch["images"].size(3) % 16 == 0, "image dims must be multiples of 16"
 
-        if self.log_transform_stdevs:
-            assert batch["background"].min() > 1e-6, "background must be positive"
```
jeff-regier, line 98: The assertion still seems like a good one

jeff-regier: (won't the sqrt raise an error if there are negative values?)

bliss/encoder/image_normalizer.py
```diff
@@ -138,13 +108,7 @@ def get_input_tensor(self, batch):
             if self.include_background:
                 inputs[0] = self.clahe(backgrounds, self.clahe_min_stdev)
 
-        if self.asinh_params:
-            processed_raw_images = (raw_images - self.asinh_thresholds_tensor) * torch.exp(
-                self.asinh_scales_tensor
-            )
-            inputs.append(torch.asinh(processed_raw_images))
```
jeff-regier, line 145: It would be nice if everyone had access to the asinh normalization eventually. But perhaps it's still too experimental?

bliss/encoder/metrics.py
```diff
@@ -64,7 +64,43 @@ def match_catalogs(self, true_cat, est_cat):
         return matching
 
 
-class DetectionPerformance(Metric):
+class SourceFilterMetric(Metric):
+    def __init__(self, source_type_filter: str | None):
```
jeff-regier, line 68: How about setting `source_type_filter` to `"total"` by default rather than `None`? Then you wouldn't need a separate variable named `source_type_name`

bliss/encoder/metrics.py
```diff
@@ -322,28 +336,79 @@ def plot(self):
...
+        n_bins = len(cutoffs) + 1
+
+        true_fluxes = true_cat["on_fluxes"][:, :, self.ref_band].contiguous()
+
+        true_filter_bools, _ = self._get_filter_bools(true_cat, est_cat)
+
         for i in range(true_cat.batch_size):
             tcat_matches, ecat_matches = matching[i]
-            self.n_matches += tcat_matches.size(0)
+            assert len(tcat_matches) == len(
```
jeff-regier, line 364: Strange place for a line break

bliss/encoder/metrics.py
```diff
@@ -322,28 +336,79 @@ def plot(self):
...
+            per_bin_elements_count = bin_indexes.bincount(minlength=n_bins)
+
+            true_gal = true_cat.galaxy_bools[i][tcat_matches][to_bin_mapping]
+            est_gal = est_cat.galaxy_bools[i][ecat_matches][to_bin_mapping]
 
-            self.gal_tp += (true_gal * est_gal).sum()
-            self.star_tp += (~true_gal * ~est_gal).sum()
+            gal_tp_bool = list(torch.split(true_gal * est_gal, per_bin_elements_count.tolist()))
+            star_tp_bool = list(torch.split(~true_gal * ~est_gal, per_bin_elements_count.tolist()))
+
```
jeff-regier, line 386: Can you avoid converting from tensors to lists and then back again here?

bliss/encoder/metrics.py
jeff-regier: How about setting
```
error_msg = "tcat_matches and ecat_matches should be of the same size"
```
on it's own line, to improve readability?

bliss/encoder/variational_dist.py
```diff
@@ -55,20 +57,32 @@ def make_dist(self, x_cat):
...
+    def __init__(self, factors, survey_bands, tile_slen):
         super().__init__()
 
         self.factors = factors
         self.survey_bands = survey_bands
         self.tile_slen = tile_slen
-        self.include_galaxy_property_loss = include_galaxy_property_loss
+        self.factors_name_set = set(self.factors.keys())
+
+        self.location_factors_name_list = ["loc"]
```
jeff-regier, line 77: How about computing binary class attributes here instead, e.g.,
```
self.loc_available = self._factors_are_available(["loc"])
```
The code is a bit more concise this way and you don't call `self_factors_are_available()` nearly as often.

case_studies/dc2_cataloging/utils/convnet.py
```diff
@@ -0,0 +1,162 @@
...
+class Detect(nn.Module):
+    def __init__(self, in_channels, out_channels):
+        super().__init__()
+        self.conv = nn.Conv2d(in_channels, out_channels, 1, 1, 0, bias=True)
+
+    def forward(self, x):
+        return self.conv(x).permute([0, 2, 3, 1])
+
+
+class Bottleneck(nn.Module):
```
jeff-regier, line 36: Better not to duplicate code. Please use the versions of these in `bliss.encoder.convnet` if possible. It's only if you're modify them that we'd want to include them here.

case_studies/dc2_cataloging/utils/image_normalizer.py
```diff
@@ -0,0 +1,242 @@
...
+    def get_input_tensor(self, batch):
+        pre_input_tensor = super().get_input_tensor(batch)
+        raw_images = batch["images"][:, self.bands].unsqueeze(2)
+
+        filtered_images = raw_images - self.asinh_thresholds_tensor
+        processed_images = filtered_images * torch.exp(self.asinh_scales_tensor)
+        processed_images = torch.asinh(processed_images)
+
+        return (
+            torch.cat((pre_input_tensor, processed_images), dim=2)
```
jeff-regier, line 62: Avoid complicate logic in return statements

case_studies/dc2_cataloging/utils/convnet.py
```diff
@@ -0,0 +1,162 @@
...
+        )
+
+    def forward(self, x):
+        preprocessed_x = self.preprocess3d(x[0]).squeeze(2)
+        asinh_params = x[1]
+
+        asinh_params_mean = asinh_params.mean(dim=2, keepdim=True)
+        asinh_params_var = asinh_params.var(dim=2, keepdim=True)
+        normalized_asinh_params = (asinh_params - asinh_params_mean) / torch.sqrt(
+            asinh_params_var + 1e-5
```
jeff-regier, line 109: avoid multiline statements when possible (typically by using more intermediate variables). this helps with readability.

case_studies/dc2_cataloging/utils/image_normalizer.py
```diff
@@ -0,0 +1,242 @@
...
+        self.asinh_thresholds_tensor = thresholds
+
+        self.asinh_buffer_size = 1000
+        self.asinh_buffer_ptr = 0
+        self.asinh_quantiles_tensor = torch.tensor([0.01, 0.1, 0.25, 0.5, 0.75, 0.9, 0.99])
+        buffer_shape = (self.asinh_buffer_size, *self.asinh_thresholds_tensor.squeeze().shape)
+        self.asinh_thresholds_buffer = torch.full(buffer_shape, torch.nan)
+        # it's hard to get the device type of torch.nn.Module
+        # https://stackoverflow.com/questions/58926054/how-to-get-the-device-type-of-a-pytorch-module-conveniently
+        self.device_indicator = torch.nn.Parameter(torch.tensor([0]), requires_grad=False)
```
jeff-regier, line 112: Can you do `device = batch["images"].device` in your `get_input_tensor` method instead?

jeff-regier: The batch should already be on the right device, I think, when `get_input_tensor` is called. If it isn't perhaps the problem is upstream?

case_studies/dc2_cataloging/utils/image_normalizer.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def num_channels_per_band(self):
+        pre_nch = super().num_channels_per_band()
+        return pre_nch + len(self.asinh_params["thresholds"])
+
+    def get_input_tensor(self, batch):
+        pre_input_tensor = super().get_input_tensor(batch)
+        raw_images = batch["images"][:, self.bands].unsqueeze(2)
+
+        self.asinh_thresholds_tensor = self.asinh_thresholds_tensor.to(
```
jeff-regier, line 122: I believe calling `to` is discouraged when you're using pytorch-lightning, because lightning is supposed to manage devices for you and calls to `to` prevent it from doing so.

jeff-regier (changes requested):
Looks good! In addition the changes above, would you verify that the runtime (seconds/iteration) is increased by this PR? (I don't think it will be, but it would be good to verify that before merging.)

YicunDuanUMich:
> Looks good! In addition the changes above, would you verify that the runtime (seconds/iteration) isn't increased by this PR? (I don't think it will be, but it would be good to verify that before merging.)

Do you mean the runtime of our new image_normalizer?

YicunDuanUMich:
image
if trained on 2 GPUs, our new implementation is slightly faster

jeff-regier:
Thanks for the revisions

PR #1015, opened 2024-06-17
Yd/test container

YicunDuanUMich:
* write code to run experiment on great lake
* upload a singularity container


case_studies/dc2_cataloging/run_great_lake/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+You can find our pre-built container at [this link](https://drive.google.com/file/d/1cezddVVZAnofFy4PZB_aOUaIOkFDvpQB/view?usp=sharing).
+
+### Run the Container on Great Lake
+
+Transfer the `bliss_machine.sif` to your Great Lake home folder, and run it with singularity. We write two simple scripts to show how to run it interactively (`run_bliss_machine_interactively.sh`) or using `sbatch` (`run_bliss_machine.sbatch` and `run_bliss_machine.sh`). If you want to configure your own container for your experiment, I suggest you to look at the following items before running your code:
+
+* The `runscript` part in `bliss_machine.def`: it can switch to a specific git branch and run self-defined command using poetry. (Probably in most cases, you don't need to rebuild container.)
+* The `paths` override in `configs/full_train_config_great_lake_exp_xxx`: make sure your data is on Great Lake and the output path is what you expect.
+* The `matmul_precision` in `configs/full_train_config_great_lake_exp_xxx`: it may be helpful to set this tag if you are using GPU with Tensor Cores (like NVIDIA A40)
```
jeff-regier, line 25: Nice, I'll be curious to hear about whether this helped

case_studies/dc2_cataloging/run_great_lake/configs/full_train_config_great_lake_exp_06-16-1.yaml
```diff
@@ -0,0 +1,165 @@
...
+            tile_slen: ${simulator.prior.tile_slen}
+
+    do_data_augmentation: true
+    compile_model: false
+    double_detect: false
+    use_checkerboard: true
+
+surveys:
+   dc2:
+      split_results_dir: ${paths.dc2}/dc2_split_results
```
jeff-regier, line 145: Better to treat the survey directories (e.g., `${paths.dc2}`) as "read only", and write output exclusively to within `${paths.output}`. When we're doing large-scale runs, we won't have write access to the survey directories.

case_studies/dc2_cataloging/DC2_exp_record.md
```diff
@@ -360,6 +360,26 @@ processed_raw_images = F.relu(raw_images - self.asinh_thresholds_tensor) * self.
...
+
+1. run fixed thresholds asinh
+
+2. run fixed thresholds asinh
+
+3. run moving avg thresholds asinh
+
+4. run moving avg thresholds asinh
+
+5. run moving avg thresholds asinh
```
jeff-regier, line 385: Ideally hydra would log all the settings used for each run in the correct output directory (the `hparams.yaml` file). I find it hard to keep track of which run is which otherwise, although creating a logging file like this one seems helpful too

case_studies/dc2_cataloging/run_great_lake/run_bliss_machine.sh
```diff
@@ -0,0 +1,4 @@
+sbatch --job-name=great_lake_exp_06-16-1 \
+       --output=great_lake_exp_06-16-1.out \
+       --error=great_lake_exp_06-16-1.err \
+       run_bliss_machine.sbatch exp_06-16-1
```
jeff-regier, line 4: It's great you figured this out! It should be easier to perform replicate runs now, which will give us more confidence in our scores.

bliss/surveys/dc2.py
```diff
@@ -30,16 +31,28 @@ def map_nested_dicts(cur_dict, func):
 
 
 class DC2Dataset(Dataset):
-    def __init__(self, split_files_list) -> None:
+    def __init__(self, split_files_list, in_memory: bool) -> None:
```
jeff-regier, line 34: I'm surprised this in-memory stuff is necessary. Is the DC2 data being staged to a local SSD? Seems like that should provide fast-enough access

YicunDuanUMich: it's really strange that when i ran the exp last night, one epoch takes 30 min, but when i ran the exp with the same setting and refactored code, it takes me about 1h per epoch

YicunDuanUMich: the data are stored at `/scrach` folder

jeff-regier: Is `/scratch` the local SSD?

jeff-regier: On the Great Lakes login nodes, `/scratch` is a massive network filesystem:
```
df -h | grep scratch
gl-scratch                                                          1.9P  1.5P  344T  82% /gpfs
```
This will be relatively slow to access.

YicunDuanUMich: sorry, i make a stupid mistake. i'm now trying to copy data first to `/tmp` and then run the code

jeff-regier: There's supposed to be a 14 TB NVMe SSD on these SPGPU instances, which should be very fast. That would be a great place to stage the DC2 data.

jeff-regier: It appears to be mounted at `/tmp_data`.

YicunDuanUMich: i find it. i'm now copying the data

jeff-regier: I recommend adding the copy command to your (non-interactive) bash login script because the data may not always be in `/tmp_data`; slurm can assign you to any SPGPU node and `/tmp_data` likely gets deleted occasionally. `rsync` may be even better than `cp` here: it copies only what is needed to make both versions of the data consistent.

case_studies/dc2_cataloging/run_great_lake/run_bliss_machine.sbatch
```diff
@@ -0,0 +1,55 @@
...
+mkdir -p ./output
+echo "create overlay image folder if not exists"
+mkdir -p ./overlay_imgs
+
+echo "copy dc2local from remote server to /tmp_data/"
+mkdir -p /tmp_data/pduan/dc2local/
+rsync -av --info=progress2 \
+ -e "ssh -i ~/.ssh/jeffbox_great_lake_side_rsa" \
+ pduan@deeplearning-01.stat.lsa.umich.edu:/data/scratch/dc2local/dc2_split_results.tar \
+ /tmp_data/pduan/dc2local/
```
jeff-regier, line 43: Did this turn out to be faster than copying from `/scratch`? This morning I'm getting about 4 GB/minute when copying from `/scratch` and from `/nfs/turbo/lsa-regier/lsstdesc-public/dc2/`.

jeff-regier: It actually may be slower to copy from deeplearning-01: iirc that computer is connected by a 100 Mbit/sec ethernet cable, which can only transfer 0.75 GB/minute at best.

YicunDuanUMich: when i test this command on great lake, it can achieve at most 110MB/s and download the data within about 12 min. Yesterday, i tested copying data from `/scratch`, but it seemed to be extremely slow (more than 30 min, probably due to rush hour).

jeff-regier: We may be limited by the ethernet speed regardless of where we fetch the data from. 1000 Mb/s ethernet cable, which I think is what deeplearning-01 uses, can transmit at most 7.5 GB/minute.

YicunDuanUMich: yes, but may be this is tolerable

case_studies/dc2_cataloging/run_great_lake/configs/full_train_config_great_lake_exp_06-16-1.yaml
YicunDuanUMich: change it to `${paths.output}`

case_studies/dc2_cataloging/DC2_exp_record.md
YicunDuanUMich: sometimes, i temporarily change the code and run the experiment, so i use this file to also record the code changes

jeff-regier (approved):
Looks good. FWIW, it seems like it's possible to run bliss on Great Lakes without any containers. I call
```
module load python/3.10.4
```
before installing poetry and that seems to be all that's required.

PR #1023, opened 2024-06-24
Refactor DC2

YicunDuanUMich:
* let DC2 inherit CachedSimulatedDataset
* extend CachedSimulatedDataset to support distributed training
* update scripts and configs
* refactor DC2


bliss/cached_dataset.py
```diff
@@ -122,26 +230,42 @@ def parse_slices(self, splits: str, length: int):
     def train_dataloader(self):
         assert self.file_paths[self.slices[0]], "No cached data found"
         transform = transforms.Compose(self.train_transforms)
-        my_dataset = MyIterableDataset(
+        dataset_type = MyIterableDataset if self.use_iterable_dataset else MyDataset
```
jeff-regier, line 233: Is it possible to always use `MyIterableDataset`, rather than having two types of datasets? It seems like `MyIterableDataset` is (or could become) flexible enough to handle all our use cases. If that's right, it'd be better not to have two ways to do it.

jeff-regier: Oh, is the issue distributed training? Can we generalize `MyIterableDataset` to support distributed training somehow? If so, we wouldn't have to maintain two distinct data loading pipelines.

YicunDuanUMich: i'll try it

bliss/cached_dataset.py
```diff
@@ -94,19 +193,28 @@ def __init__(
...
 
-        file_names = [f for f in os.listdir(cached_data_path) if f.endswith(".pt")]
-        self.file_paths = [os.path.join(cached_data_path, f) for f in file_names]
+        self.file_paths = None
+        self.slices = None
 
-        # parse slices from percentages to indices
-        self.slices = self.parse_slices(splits, len(self.file_paths))
+    def setup(self, stage: str) -> None:
+        if stage == "fit":
```
jeff-regier, line 212: Is it necessary to do different things depending on `stage`?

YicunDuanUMich: this is because i expect it to load the file path after 'prepare data'

YicunDuanUMich: of course, if we remove the if statement it can also work well. i add the if statement for clear coding

bliss/surveys/dc2.py
```diff
@@ -30,94 +29,55 @@ def map_nested_dicts(cur_dict, func):
...
-                split = torch.load(split_file)
-        else:
-            split = self.in_memory_data[idx]
-        return split
+def split_list(ori_list, sub_list_len):
+    return [ori_list[i : (i + sub_list_len)] for i in range(0, len(ori_list), sub_list_len)]
 
 
-class DC2(Survey):
+class DC2Dataset(CachedSimulatedDataset):
```
jeff-regier, line 36: Would you rename `CachedSimulatedDataset` and `DC2Dataset` to `CachedSimulatedDataModule` and `DC2DataModule`, respectively? I realize calling them datasets is sort of confusing now that they inherit from `DataModule` and we have other classes that inherit from `Dataset`.

bliss/cached_dataset.py
YicunDuanUMich: in previous implementation, we didn't test the `stage` (in its original interface, `stage` doesn't have a default value)

bliss/surveys/dc2.py
YicunDuanUMich: yeah, i can handle it

bliss/cached_dataset.py
YicunDuanUMich: I found a notebook that describes how to write distributed iterable dataset ([this link](https://colab.research.google.com/drive/1OFLZnX9y5QUFNONuvFsxOizq4M-tFvk-?usp=sharing)). It seems that it will treat iterable dataset as a map dataset, and the random shuffle is hard to realize. Another way is to rewrite our iterable dataset using `rank` information, but I suggest using PyTorch's `DistributedSampler` (in our code, it's `MyDistributedSampler`) so that it can automatically hanle the workers and distributed processes using PyTorch's API.

jeff-regier: This may be worth considering too:
https://pytorch.org/data/main/torchdata.datapipes.iter.html
It appears to be an alternative to an `IterableDataset`.
Not that we should be carried away with optimizing something that already works, but writing out tens of thousands of "split files" seems problematic: hard drives don't provide good throughput if data is accessed as small blocks in a random order.

YicunDuanUMich: Thanks! I'll read this link. By the way, in our `MyDataset`, it reads the data by chunk (equivalent to `MyIterableDataset`) so we are now not using tens of thousands of "split files".

bliss/cached_dataset.py
jeff-regier: It seems better not to call `setup` if `stage != "fit"`, rather have the entire function body within a conditional block. (Please lmk if I'm missing something.)

bliss/cached_dataset.py
YicunDuanUMich: yeah, it's equivalent to `MyIterableDataset`; i'll delete `MyIterableDataset`

jeff-regier: Great!

bliss/cached_dataset.py
```diff
@@ -47,6 +51,101 @@ def __call__(self, ex):
         return ex
 
 
+class MySampler(Sampler):
```
jeff-regier, line 54: Maybe call this `ChunkedSampler` to explain it's function a bit

YicunDuanUMich:
I'm working on ensuring the seed for each subprocess is the same. I seems that even though we use the `pl.seed_everything`, the seeds on different GPUs are still different.

bliss/cached_dataset.py
YicunDuanUMich: do you mean we should write the code like:
```python
if stage == "fit":
    setup(stage)
```

YicunDuanUMich:
I write a new setup function. Now it generates specific dataset for each stage.

jeff-regier (approved):

PR #1030, opened 2024-06-28
Regular update

YicunDuanUMich:
* slightly refactor `cached_dataset.py`
* update the notebooks in my case study
* update the scripts


bliss/old_data_augmentation.py
```diff
@@ -0,0 +1,55 @@
...
+    return plocs_clone
+
+
+def plocs_vflip(plocs, image_size):
+    plocs_clone = plocs.clone()
+    plocs_clone[..., 0] = image_size - plocs[..., 0] - 1
+    return plocs_clone
+
+
+class RotateFlipTransform(torch.nn.Module):
```
jeff-regier, line 20: There appears to be a bug in this version of `RotateFlipTransform`. `fr_transform` refers to the version in `data_augmentation.py` and `old_fr_transform` refers to the version here.
```
datum01["tile_catalog"]["n_sources"]
tensor([[0, 0, 0, 0, 0],
        [1, 1, 1, 1, 1],
        [0, 0, 0, 0, 0],
        [0, 0, 0, 0, 0],
        [0, 0, 0, 0, 0]])
fr_transform(datum01, rotate_id=1, do_flip=False)["tile_catalog"]["n_sources"]
tensor([[0, 1, 0, 0, 0],
        [0, 1, 0, 0, 0],
        [0, 1, 0, 0, 0],
        [0, 1, 0, 0, 0],
        [0, 1, 0, 0, 0]])
old_fr_transform(datum01, rotate_id=1, do_flip=False)["tile_catalog"]["n_sources"]
tensor([[1, 0, 0, 0, 0],
        [0, 0, 0, 0, 0],
        [1, 1, 0, 0, 0],
        [1, 0, 0, 0, 0],
        [0, 0, 0, 0, 1]])
```

jeff-regier: `old_fr_transform` appears to be adding 0.5 one of the location coordinates, which can cause light sources to "switch" tiles. Unfortunately this bug wouldn't seem to help us explain the recent performance regression, but it seems important to fix it nevertheless

jeff-regier: It could be that there's an additional bug in the version of `RotateFlipTransform` in `data_augmentation.py`, and that once this one is fixed, that one will be more apparent.

YicunDuanUMich: Thanks! I'll try to fix it.

YicunDuanUMich:
ready for merge

jeff-regier (approved):
Looks good

PR #1031, opened 2024-07-02
Refactor Generative Prior to use DES sources

kapnadak:
- Separated generative prior into two independent priors for background and cluster
- Background sources are sampled directly from DES catalogs
- Background prior samples sources 40 times (for 40 images) from a single DES catalog before switching to a new DES catalog. This makes catalog generation almost 40x faster since loading dataframe was a bottleneck
- Allow GalSim to determine the correct stamp size to use


case_studies/galaxy_clustering/data_generation/background_prior.py
```diff
@@ -0,0 +1,305 @@
...
+        self.dec_cen = -40.228830895890404
+        self.pixels_per_mpc = 80
+        self.mean_sources = 0.004
+        self.G1_beta = 0.6
+        self.G1_loc = 0
+        self.G1_scale = 0.035
+        self.G2_beta = 0.6
+        self.G2_loc = 0
+        self.G2_scale = 0.032
+        self.pixel_scale = 0.26
```
jeff-regier, line 35: pixel scale for coadds is 0.263, not 0.26, according to https://arxiv.org/pdf/2101.05765

kapnadak: Just fixed this, and the speckled pattern!

jeff-regier (approved):

PR #1034, opened 2024-07-08
Small changes to run script

eshanjain1:
Needed to make some changes to the run script to account for the encoder.


case_studies/strong_lensing/run_sim.sh
```diff
@@ -0,0 +1,182 @@
+#!/bin/bash
+
+# Script to automate running the galsim command with the yaml configuration
+
+# Default Values
+num_files=3
+num_galaxies=3
+img_size=500
+
+
```
jeff-regier, line 10: If I run the script without arguments, it just prints a message and exits cleanly:
```
$ ./run_sim.sh 
Starting data generation...
$
```

Would you change it to run with sensible default if the user provides no arguments? (And perhaps to print a message about what defaults it is using.)

jeff-regier: Ah, I see the output was getting written to a log file. Perhaps send the output to stdout too (or instead), because otherwise a user won't no were to look.

case_studies/strong_lensing/run_sim.sh
```diff
@@ -0,0 +1,182 @@
...
+while getopts "f:s:g:" opt; do
```
jeff-regier, line 11: It seem the problem was that `data/DECam_00154912_12_psfcat.psf` isn't found. Is that a small file? If so, would you add it to the repo so your script can be run by others?

case_studies/strong_lensing/run_sim.sh
```diff
@@ -0,0 +1,182 @@
...
+fi
+
+# Log file
+LOG_FILE="galsim_run.log"
+
+NUM_UNLENSED_IMAGES=$((num_galaxies/2))
+NUM_LENSED_IMAGES=$((num_galaxies-NUM_UNLENSED_IMAGES))
+
+echo "Starting data generation..."
+echo "Data generation beginning..." &> $LOG_FILE
```
jeff-regier, line 69: Would you display error messages in addition to directing them to a log file?

jeff-regier: Actually, how about just displaying all the output to STDOUT? Otherwise the program appears to be hanging

case_studies/strong_lensing/simulate.yaml
```diff
@@ -0,0 +1,118 @@
+
+variables :
+  output_dir : output_yaml
```
jeff-regier, line 3: My `output_yaml` directory doesn't appear contain any yaml. Could we just call it `output`?

jeff-regier: (It contains png and fits files)

case_studies/strong_lensing/run_sim.sh
jeff-regier: (A user can always redirect STDOUT to a file if they want, and then they'll know where to find the output.)

jeff-regier (changes requested):
There seem to be some merge conflicts

jeff-regier (approved):

PR #1047, opened 2024-07-21
Regular update

YicunDuanUMich:
* add ellipticity estimation
* add notebooks to show error wrt blendedness


YicunDuanUMich:
ready for merge

bliss/conf/base_config.yaml
```diff
@@ -129,11 +129,10 @@ variational_factors:
 metrics:
     detection_performance:
         _target_: bliss.encoder.metrics.DetectionPerformance
-        mag_bin_cutoffs: [19, 19.4, 19.8, 20.2, 20.6, 21, 21.4, 21.8]
-        mag_band: 2
+        bin_cutoffs: [200, 400, 600, 800, 1000]
     source_type_accuracy:
         _target_: bliss.encoder.metrics.SourceTypeAccuracy
-        flux_bin_cutoffs: [200, 400, 600, 800, 1000]
+        bin_cutoffs: [200, 400, 600, 800, 1000]
```
jeff-regier, line 135: `bin_cutoffs` is quite vague without the word `flux`, can we avoid changing it?

bliss/encoder/metrics.py
```diff
@@ -149,55 +149,61 @@ def _get_postfix(self):
...
 
-        self.mag_band = mag_band
-        self.mag_bin_cutoffs = mag_bin_cutoffs if mag_bin_cutoffs else []
-        self.bin_unit_is_flux = bin_unit_is_flux
+        self.bin_cutoffs = bin_cutoffs if bin_cutoffs else []
+        self.ref_band = ref_band
+        self.bin_type = bin_type
         self.exclude_last_bin = exclude_last_bin
 
+        assert self.bin_type in {"Flux", "Mag"}, "invalid bin type"
```
jeff-regier, line 167: I'd call this `self.flux_bin_type`. Possibly values could include `nmgy`, `njy`, and `mag`.

jeff-regier: i.e., `nmgy`, `njy`, and `mag` are all measures of flux. Of the three, `mag` is the most vague because there are multiple magnitude scaled

bliss/encoder/metrics.py
```diff
@@ -149,55 +149,61 @@ def _get_postfix(self):
...
         assert isinstance(true_cat, FullCatalog), "true_cat should be FullCatalog"
         assert isinstance(est_cat, FullCatalog), "est_cat should be FullCatalog"
 
-        if self.mag_band is not None:
-            true_mags = true_cat.on_fluxes if self.bin_unit_is_flux else true_cat.magnitudes
-            true_mags = true_mags[:, :, self.mag_band].contiguous()
-            est_mags = est_cat.on_fluxes if self.bin_unit_is_flux else est_cat.magnitudes
-            est_mags = est_mags[:, :, self.mag_band].contiguous()
+        if self.ref_band is not None:
+            if self.bin_type == "Flux":
```
jeff-regier, line 185: It'd be nice if the catalog classes did the conversions among physical units. Based on Camille's comments yesterday, it seems like we should be using magnitude more often

bliss/encoder/metrics.py
```diff
@@ -149,55 +149,61 @@ def _get_postfix(self):
...
+            else:
+                raise NotImplementedError()
+
         else:
             # hack to match regardless of magnitude; intended for
             # catalogs from surveys with incompatible filter bands
-            true_mags = torch.ones_like(true_cat["plocs"][:, :, 0])
-            est_mags = torch.ones_like(est_cat["plocs"][:, :, 0])
+            true_bin_measures = torch.ones_like(true_cat["plocs"][:, :, 0])
+            est_bin_measures = torch.ones_like(est_cat["plocs"][:, :, 0])
```
jeff-regier, line 202: `true_fluxes` and `est_fluxes`, perhaps

bliss/encoder/variational_dist.py
```diff
@@ -79,9 +76,18 @@ def compute_nll(self, params, true_tile_cat):
             gating = rearrange(true_tile_cat.star_bools, "b ht wt 1 1 -> b ht wt")
         elif self.nll_gating == "is_galaxy":
             gating = rearrange(true_tile_cat.galaxy_bools, "b ht wt 1 1 -> b ht wt")
+        elif self.nll_gating == "is_in_cosmodc2":
+            gating = rearrange(true_tile_cat["cosmodc2_mask"], "b ht wt 1 1 -> b ht wt")
```
jeff-regier, line 80: Could you made this change in a subclass of `VariationalDist` instead?

bliss/surveys/dc2.py
```diff
@@ -320,7 +320,7 @@ def read_image_for_bands(self, image_index):
                 self.image_lim[0], self.image_lim[1]
             )
 
-            image += bg
+            # we assume image doesn't contain bg
```
jeff-regier, line 323: No need to append `bg` then, right?

bliss/encoder/metrics.py
```diff
@@ -210,20 +216,20 @@ def update(self, true_cat, est_cat, matching):
...
+            tmi = cur_batch_true_bin_meas[cur_batch_true_filter_bools]
+            emi = cur_batch_est_bin_meas[cur_batch_est_filter_bools]
 
             tcat_matches = tcat_matches[cur_batch_true_filter_bools[tcat_matches]]
             ecat_matches = ecat_matches[cur_batch_est_filter_bools[ecat_matches]]
 
-            tmim = cur_batch_true_mags[tcat_matches]
-            emim = cur_batch_est_mags[ecat_matches]
+            tmim = cur_batch_true_bin_meas[tcat_matches]
+            emim = cur_batch_est_bin_meas[ecat_matches]
```
jeff-regier, line 232: more specific names: `cur_batch_true_fluxes` and `cur_batch_est_fluxes`

bliss/encoder/metrics.py
jeff-regier: In `TileCatalog`, perhaps `on_fluxes` could be converted a method that takes units as an argument? Or we could replace it with three properties: `on_nmgy`, `on_njy`, and `on_mag`.

case_studies/dc2_cataloging/train_config.yaml
```diff
@@ -34,10 +34,87 @@ variational_factors:
...
+  flux_error:
+    _target_: bliss.encoder.metrics.FluxError
+    survey_bands: ${encoder.survey_bands}
+  # shear_error:
+  #   _target_: case_studies.dc2_cataloging.utils.metrics.ShearMSE
+  #   bin_cutoffs: [200, 400, 600, 800, 1000]
+  #   bin_type: "Flux"
+  ellipticity_error:
+    _target_: case_studies.dc2_cataloging.utils.metrics.EllipticityMSE
+    bin_cutoffs: [200, 400, 600, 800, 1000]
```
jeff-regier, line 101: Perhaps we should be specifying everything in terms of magnitude, if those are the units that we'll use for publishing the results

case_studies/dc2_cataloging/utils/encoder.py
```diff
@@ -109,12 +84,16 @@ def _compute_loss(self, batch, logging_name):
...
-
+        x_features = self.get_features(batch)
         x_cat = self.catalog_net(x_features)
         loss = self.var_dist.compute_nll(x_cat, target_cat)
 
+        nan_mask = torch.isnan(loss)
+        if nan_mask.any():
+            loss = loss[~nan_mask]
+            msg = f"NaN detected in loss. Ignored {nan_mask.sum().item()} NaN values."
+            warnings.warn(msg)
```
jeff-regier, line 95: Has this helped? I deleted it from the main encoder because although it seemed like a good idea, it never seemed to catch any nans.

bliss/conf/base_config.yaml
YicunDuanUMich: I rename it to `bin_cutoffs` because we also use `blendedness` as x-axis for some metrics.

bliss/encoder/metrics.py
YicunDuanUMich: Good idea. I'll do it.

bliss/encoder/variational_dist.py
YicunDuanUMich: No problem.

bliss/surveys/dc2.py
YicunDuanUMich: I'll remove it.

case_studies/dc2_cataloging/utils/encoder.py
YicunDuanUMich: In my experiments, it also can't catch any NaN.

bliss/conf/base_config.yaml
jeff-regier: Oh I see. That does seem like a useful generalization then.

bliss/encoder/metrics.py
YicunDuanUMich: I change it to `on_nmgy`, `on_njy` and `on_mag`

bliss/encoder/variational_dist.py
YicunDuanUMich: Write some gating classes to realize this

bliss/surveys/dc2.py
YicunDuanUMich: already removed

case_studies/dc2_cataloging/train_config.yaml
YicunDuanUMich: change to magnitudes

YicunDuanUMich:
Because our bin cutoffs can be `blendedness`, I don't change the `bin_cutoffs`, `bin_type` and some variables like `cur_batch_true_bin_meas`.

.flake8
```diff
@@ -10,6 +10,10 @@ ignore =
     # high Jones complexity isn't always so bad
     WPS221,
 
+    # more than 3 `elif` branches will not be a problem
+    # the solutions found in network (like creating a class to do the `elif`) will make the code complicated
+    WPS223,
```
jeff-regier, line 15: I wonder if `match` and `case` would be more clear. They added those in Python 3.10, apparently

bliss/catalog.py
```diff
@@ -164,13 +159,12 @@ def on_fluxes(self):
...
-    def magnitudes(self):
-        # TODO: we shouldn't assume fluxes are stored in nanomaggies because they aren't for DC2
-        return convert_nmgy_to_mag(self.on_fluxes)
+    def on_mag(self) -> Tensor:
+        return convert_nmgy_to_mag(self.on_nmgy)
 
     @property
-    def magnitudes_njy(self):
-        return convert_nmgy_to_njymag(self.on_fluxes)
+    def on_njy(self) -> Tensor:
```
jeff-regier, line 166: But this conversion is to log njy, not njy

jeff-regier: Is njymag a unit that others use, which is different from mag?  Could you add a citation to the code (in a comment) explaining the units

bliss/catalog.py
```diff
@@ -149,12 +149,7 @@ def galaxy_bools(self) -> Tensor:
         return is_galaxy * self.is_on_mask.unsqueeze(-1)
 
     @property
-    def on_fluxes(self):
-        """Gets fluxes of "on" sources based on whether the source is a star or galaxy.
-
-        Returns:
-            Tensor: a tensor of fluxes of size (b x nth x ntw x max_sources x 1)
-        """
+    def on_nmgy(self):
```
jeff-regier, line 152: So internally are we storing everything in nanomaggies now, even for DC2?

jeff-regier: (Will that change how the encoder operates on DC2?)

jeff-regier (changes requested):

bliss/encoder/metrics.py
```diff
@@ -149,55 +149,61 @@ def _get_postfix(self):
...
-            est_mags = est_cat.on_fluxes if self.bin_unit_is_flux else est_cat.magnitudes
-            est_mags = est_mags[:, :, self.mag_band].contiguous()
+        if self.ref_band is not None:
+            if self.bin_type == "Flux":
+                true_bin_measures = true_cat.on_nmgy
+                true_bin_measures = true_bin_measures[:, :, self.ref_band].contiguous()
+                est_bin_measures = est_cat.on_nmgy
+                est_bin_measures = est_bin_measures[:, :, self.ref_band].contiguous()
+            elif self.bin_type == "Mag":
+                true_bin_measures = true_cat.on_njy
```
jeff-regier, line 191: `true_cat.on_fluxes(self.bin_type)` seems more clear. We could avoid the `elif` statement then

bliss/encoder/variational_dist.py
```diff
@@ -40,6 +43,37 @@ def compute_nll(self, x_cat, true_tile_cat):
...
+class NllGating(ABC):
+    @classmethod
+    @abstractmethod
+    def get_nll_gating(cls, true_tile_cat: TileCatalog):
+        """Get Gating for nll."""
+
+
+class NullGating(NllGating):
+    @classmethod
+    def get_nll_gating(cls, true_tile_cat: TileCatalog):
```
jeff-regier, line 55: Maybe `__call__` would be a better method name? That's the name of a special method that gets invoked when you treat the object as if it were a function.

case_studies/dc2_cataloging/utils/metrics.py
```diff
@@ -100,3 +102,373 @@ def get_cur_filter_bools(self, true_cat, est_cat):
...
+    def update(self, true_cat, est_cat, matching):
+        cutoffs = torch.tensor(self.bin_cutoffs, device=self.device)
+        n_bins = len(cutoffs) + 1
+
+        if self.bin_type == "Flux":
+            true_bin_measures = true_cat.on_nmgy[:, :, self.ref_band].contiguous()
+        elif self.bin_type == "Mag":
+            true_bin_measures = true_cat.on_njy[:, :, self.ref_band].contiguous()
+        else:
+            raise NotImplementedError()
```
jeff-regier, line 206: Perhaps simpler:
```
if self.bin_type in {"flux", "mag"}:
    true_bin_measures = true_cat.on_fluxes(self.bin_type)[:, :, self.ref_band].contiguous()
```

.flake8
YicunDuanUMich: fixed

bliss/catalog.py
YicunDuanUMich: Add an explanation:
```
For the difference between mag (Pogson magnitude) and njymag (AB magnitude), please view
    the "Flux units: maggies and nanomaggies" part of
    https://www.sdss3.org/dr8/algorithms/magnitudes.php#nmgy
    When we change the standard source to AB sources, we need to do the conversion
    described in "2.10 AB magnitudes" at
    https://pstn-001.lsst.io/fluxunits.pdf
```

bliss/catalog.py
YicunDuanUMich: Yes, at the beginning of this project, we stored fluxes in nanomaggies.

bliss/encoder/metrics.py
YicunDuanUMich: fixed

bliss/encoder/variational_dist.py
YicunDuanUMich: fixed

case_studies/dc2_cataloging/utils/metrics.py
YicunDuanUMich: fixed

jeff-regier (approved):
Thanks for the revisions!

PR #1048, opened 2024-07-22
dc2 weak lensing

shreyasc30:


shreyasc30:
waiting on #1047 for image normalizers & potential dc2 changes

bliss/cached_dataset.py
```diff
@@ -149,7 +149,10 @@ def __getitem__(self, index):
             self.buffered_file_index = converted_index
             with open(self.file_paths[converted_index], "rb") as f:
                 self.buffered_data = torch.load(f)
-        output_data = self.buffered_data[converted_sub_index]
+        try:
+            output_data = self.buffered_data[converted_sub_index]
+        except KeyError:
```
jeff-regier, line 154: Why are we getting key errors here?

bliss/catalog.py
```diff
@@ -190,27 +196,44 @@ def to_full_catalog(self):
...
             params[param_name] = param
 
-        params["n_sources"] = reduce(self["n_sources"], "b nth ntw -> b", "sum")
+        if self.locations_sources:
+            params["n_sources"] = reduce(self["n_sources"], "b nth ntw -> b", "sum")
+
+        if "shear" in self:  # pylint: disable=too-many-branches
+            params["shear"] = self.get("shear")
+        if "convergence" in self:  # pylint: disable=too-many-branches
+            params["convergence"] = self.get("convergence")
```
jeff-regier, line 236: I'm reluctant to add special cases to this function that will make it even more complicated; it's so complex already. Is there a way to avoid these changes?

bliss/catalog.py
```diff
@@ -116,7 +116,13 @@ class TileCatalog(BaseTileCatalog):
     galaxy_params_index = {k: i for i, k in enumerate(galaxy_params)}
 
     def __init__(self, tile_slen: int, d: Dict[str, Tensor]):
-        self.max_sources = d["locs"].shape[3]
+        self.locations_sources = True
+
+        if "locs" in d:
+            self.max_sources = d.get("locs").shape[3]
```
jeff-regier, line 122: Could you use `BaseTileCatalog` instead of adding special cases to `TileCatalog`?

bliss/conf/base_config.yaml
```diff
@@ -239,8 +239,8 @@ surveys:
         prepare_data_processes_num: 4
         data_in_one_cached_file: 1250
         splits: 0:80/80:90/90:100
-        batch_size: 64
-        num_workers: 4
+        batch_size: 1
+        num_workers: 1
```
jeff-regier, line 243: Better to make the changes in config file that extends base_config.yaml, rather than directly in base_config.yaml

bliss/conf/base_config.yaml
```diff
@@ -303,7 +302,7 @@ train:
     matmul_precision: high
 
 predict:
-    dataset: ${surveys.sdss}
+    dataset: ${surveys.dc2}
```
jeff-regier, line 305: same here

bliss/surveys/dc2.py
```diff
@@ -320,7 +320,6 @@ def read_image_for_bands(self, image_index):
                 self.image_lim[0], self.image_lim[1]
             )
 
-            image += bg
             image_list.append(image)
             bg_list.append(bg)
```
jeff-regier, line 324: Perhaps we don't need to record `bg` at all any more?

case_studies/weak_lensing/lensing_dc2.py
```diff
@@ -0,0 +1,190 @@
...
+        dc2_cat_path: str,
+        image_slen: int,  # assume square images: image_slen x image_slen
+        tile_slen: int,
+        splits: str,
+        batch_size: int,
+        num_workers: int,
+        cached_data_path: str,
+        **kwargs,
+    ):
+        super().__init__(
```
jeff-regier, line 23: Can you pass `*args` and `**kwargs` to the parent class, instead of explicitly listing all of them?

case_studies/weak_lensing/lensing_encoder.py
```diff
@@ -0,0 +1,172 @@
...
+        var_dist: VariationalDist,
+        sample_image_renders: MetricCollection,
+        mode_metrics: MetricCollection,
+        sample_metrics: Optional[MetricCollection] = None,
+        optimizer_params: Optional[dict] = None,
+        scheduler_params: Optional[dict] = None,
+        reference_band: int = 2,
+        **kwargs,
+    ):
+        super().__init__(
```
jeff-regier, line 31: Same here, how about passing `*args` and `**kwargs` to the parent class? (and removing named arguments in the construct that are just passed on to `Encoder`)

case_studies/weak_lensing/lensing_encoder.py
```diff
@@ -0,0 +1,172 @@
...
+    def predict_step(self, batch, batch_idx, dataloader_idx=0):
+        """Pytorch lightning method."""
+        with torch.no_grad():
+            return {
+                "mode_cat": self.sample(batch, use_mode=True),
+                # we may want multiple samples
+                "sample_cat": self.sample(batch, use_mode=False),
+            }
+
+    def configure_optimizers(self):
```
jeff-regier, line 168: Could you omit some of these methods, which are the same in the parent class?

case_studies/weak_lensing/lensing_metrics.py
```diff
@@ -6,35 +6,62 @@ class LensingMapMSE(Metric):
...
+        baseline_convergence_sq_err = ((true_convergence - baseline_pred_convergence) ** 2).sum()
+
+        self.shear1_sum_squared_err += shear1_sq_err
+        self.baseline_shear1_sum_squared_err += baseline_shear1_sq_err
+        self.shear2_sum_squared_err += shear2_sq_err
+        self.baseline_shear2_sum_squared_err += baseline_shear2_sq_err
+        self.convergence_sum_squared_err += convergence_sq_err
+        self.baseline_convergence_sum_squared_err += baseline_convergence_sq_err
+
+        self.total = torch.tensor(  # pylint: disable=attribute-defined-outside-init
```
jeff-regier, line 48: Would it work to set `self.total = None` in the constructor? (rather than disabling this check)

bliss/catalog.py
shreyasc30: will take a look. The main issue here and in the previous case is that TileCatalog assumes ''locs'' is in the data, but we provide no such parameter

bliss/catalog.py
shreyasc30: this change was made since shear and convergence don't follow the generalized pattern requiring indices_of_on_sources  that the catalog mandates. will take a look at the version after your new PR and see what can be done here.

case_studies/weak_lensing/lensing_dc2.py
shreyasc30: they are listed since some of them are hardcoded as dummy values: i.e. min_flux_for_loss, max_sources_per_tile, image_lim, prepare_data_processes, data_in_one_cached_file, train_transforms, nontrain_transforms. 

I can instead add these hardcoded values to kwargs but this provides more visibility into the subclassing imo

case_studies/weak_lensing/lensing_encoder.py
shreyasc30: will remove named arguments, but same issue as above with hardcoding certain params

bliss/cached_dataset.py
shreyasc30: investigating. was initially done as a workaround for file names, perhaps redundant now. will fix if so

bliss/catalog.py
shreyasc30: resolved in refactor

bliss/catalog.py
shreyasc30: resolved in refactor

bliss/conf/base_config.yaml
shreyasc30: resolved

bliss/conf/base_config.yaml
shreyasc30: resolved

bliss/surveys/dc2.py
shreyasc30: waiting on incoming dc2 change (in the open yd PR) that will do the same

case_studies/weak_lensing/lensing_encoder.py
shreyasc30: resolved

shreyasc30:
refactored the tile catalog and made the small changes.
imo, the explicit stating of input parameters is better than just accepting **kwargs; that is there since the base_config provides arguments that can't be replaced (i.e. without it, there's some cases of "unexpected keyword ___ found" , but in this case we ensure that any missing arguments in lensing_config are noted quickly by the explicit constructor.

bliss/cached_dataset.py
shreyasc30: waiting for data gen to fix this, still outstanding -- will test & fix first thing tomorrow

case_studies/weak_lensing/lensing_encoder.py
jeff-regier: Ok sounds good

jeff-regier (changes requested):
Looking good. Please mark this PR as ready for review once the merge conflicts are resolved.

jeff-regier (approved):

PR #1052, opened 2024-08-01
Improve multidetection and add some calibration plots

YicunDuanUMich:
* improve the multidetection according to YOLOv10
* add some calibration plots to assess posterior distribution


bliss/catalog.py
```diff
@@ -126,7 +126,7 @@ class TileCatalog(BaseTileCatalog):
     galaxy_params_index = {k: i for i, k in enumerate(galaxy_params)}
 
     def __init__(self, d: Dict[str, Tensor]):
-        self.max_sources = d["locs"].shape[3]
+        assert "locs" in d
```
jeff-regier, line 129: Could we also assert that locs is a 5d tensor here?

bliss/catalog.py
```diff
@@ -347,12 +358,24 @@ def filter_by_flux(self, min_flux=0, max_flux=torch.inf, band=2):
...
         d = {}
+        nan_ok_variables = {
+            "shear",
+            "convergence",
+            "ellipticity",
+            "blendedness",
+            "locs_vsbc",
+            "ellipticity_vsbc",
+            "star_fluxes_vsbc",
+            "galaxy_fluxes_vsbc",
```
jeff-regier, line 369: Better not to have a list of case-study specific variable names here

jeff-regier: Is it possible to disallow nans from all variables?

bliss/catalog.py
```diff
@@ -347,12 +358,24 @@ def filter_by_flux(self, min_flux=0, max_flux=torch.inf, band=2):
...
+        }
         for key, val in sorted_self.items():
-            if key in {"shear", "convergence"}:
```
jeff-regier, line 351: Good to remove special logic for shear and convergence here if at all possible

bliss/catalog.py
```diff
@@ -390,6 +413,25 @@ def union(self, other, disjoint=False):
                 d[k] = torch.where(ns11 > 0, d1, d2)
         return TileCatalog(d)
 
+    def stack(self, other):
```
jeff-regier, line 416: Is this different than the merge  method?

bliss/encoder/variational_dist.py
```diff
@@ -304,6 +305,14 @@ def mode(self):
         assert (self.base_dist.mean >= 0).all() and (self.base_dist.mean <= 1).all()
         return self.base_dist.mode
 
+    @property
+    def batch_shape(self):
+        return self.base_dist.batch_shape
+
+    @property
+    def event_shape(self):
+        return self.base_dist.event_shape
```
jeff-regier, line 314: OK makes sense

case_studies/dc2_cataloging/utils/variational_dist.py
```diff
@@ -1,232 +1,117 @@
...
+        # for now, only test the vsbc for locs, ellipticity and flux
+        vsbc_variables = {
+            "locs",
+            "ellipticity",
+            "star_fluxes",
+            "galaxy_fluxes",
+        }
+        for qk, params in fp_pairs:
+            if qk.name in vsbc_variables:
+                sample_result[qk.name + "_vsbc"] = self._get_vsbc(qk, params, true_tile_cat)
```
jeff-regier, line 78: Does the VSBC metric deal with checkerboard structure of the variational distribution correctly? It seems tricky to apply VSBC with the checkerboard because we can't just read off the CDF of the marginal, as we could if the variational distribution were mean field

case_studies/dc2_multidetection/train_config.yaml
```diff
@@ -0,0 +1,169 @@
...
+    _target_: bliss.encoder.metrics.DetectionPerformance
+    bin_cutoffs: [23.9, 24.1, 24.5, 24.9, 25.6]
+    bin_type: "njymag"
+    filter_list:
+      - _target_: case_studies.dc2_cataloging.utils.metrics.SourceCountFilter
+        filter_source_count: "2plus"
+  source_type_accuracy:
+    _target_: bliss.encoder.metrics.SourceTypeAccuracy
+    bin_cutoffs: [23.9, 24.1, 24.5, 24.9, 25.6]
+    bin_type: "njymag"
```
jeff-regier, line 97: I've been reading the SDSS using AB magnitude too. (See https://classic.sdss.org/dr7/algorithms/fluxcal.php) Given that, is njymag vs mag the correct distinction? I do agree that mag has different meanings for different surveys, because the surveys bands have different filter curves, but I'm not that's captured by the njymag/mag distinction.

jeff-regier: Perhaps bliss (the main package) should be agnostic to the units of flux. Then each case study would use whatever units are appropriate for the survey.

YicunDuanUMich:
Thanks for the review. I'll fix them tomorrow.

bliss/catalog.py
YicunDuanUMich: I've removed the nan test here. nans may be necessary. They are different from 0 and play the role of mask.

I also remove the max_flux in this function. It doesn't function correctly.

bliss/catalog.py
YicunDuanUMich: The key of `stack` is to add the `n_sources_mask`. This variable will be very helpful in multidetection. Since the sources in tensor are sparse after using this function, I didn't call it union.

jeff-regier: I'm reluctant to add new methods to the main bliss package that are used solely in one case study. Could you keep this function outside of the TileCatalog class, in the case study where it's used, instead?

bliss/catalog.py
jeff-regier: Sounds good

bliss/catalog.py
YicunDuanUMich: get it

bliss/catalog.py
YicunDuanUMich: have fixed

bliss/catalog.py
YicunDuanUMich: remove the stack

case_studies/dc2_cataloging/utils/variational_dist.py
YicunDuanUMich: I'll try to find a way to bypass it in the next PR.

case_studies/dc2_multidetection/train_config.yaml
YicunDuanUMich: I realize that this may change the code of others' case stuides. Maybe it's better to put it in a specific PR.

jeff-regier (approved):

PR #1053, opened 2024-08-03
Spatially-variant PSF/background reproducability

aakashdp6548:
Mostly this PR is to reproduce the results and figures from the spatially-variant PSF paper.

I've made some changes to metrics, including:
- Changing `SourceTypeAccuracy` to bin by magnitude instead of flux for consistency with other metrics
- Added binning by magnitude for `FluxError` and `GalaxyShapeError`


aakashdp6548:
Hey @jeff-regier, here is the code to reproduce the results/figures for the PSF paper. I haven't pulled from main yet so it's not ready to merge, just wanted to get this up as a draft so I don't forget. 

I made some changes to the metrics (mentioned in the first post). Can you let me know if these are fine to change, or if they may break someone else's results? I can create a new metric instead but didn't want to increase code bloat...

bliss/cached_dataset.py
```diff
@@ -69,7 +69,7 @@ def __call__(self, datum_in):
 
 class ChunkingSampler(Sampler):
     def __init__(self, dataset: Dataset) -> None:
-        super().__init__()
+        super().__init__(dataset)
```
jeff-regier, line 72: I think this fix was made on master too

bliss/encoder/metrics.py
```diff
@@ -372,18 +372,18 @@ def plot(self):
 class SourceTypeAccuracy(FilterMetric):
     def __init__(
         self,
-        flux_bin_cutoffs: list,
+        mag_bin_cutoffs: list,
```
jeff-regier, line 375: In a recent PR, @YicunDuanUMich has changed this variable name to `bin_cutoffs`, and added another variable to specify the type of bin. The type can be either "flux" or "mag". Perhaps that change will help here

bliss/encoder/metrics.py
```diff
@@ -501,61 +501,201 @@ def get_internal_states(self):
...
+        exclude_last_bin=False,
+    ):
         super().__init__()
 
-        gpe_init = torch.zeros(len(self.GALSIM_NAMES))
-        self.add_state("galsim_param_err", default=gpe_init, dist_reduce_fx="sum")
+        self.mag_band = mag_band
+        self.mag_bin_cutoffs = mag_bin_cutoffs
+        self.n_bins = len(self.mag_bin_cutoffs) + 1
+        self.exclude_last_bin = exclude_last_bin  # used to ignore dim objects
```
jeff-regier, line 603: @YicunDuanUMich Will you look over Aakash's change to metrics.py too? I'd like it we could use the same metrics code for both your DC2 paper and Aakash's PSF-variation paper, so that the results are as comparable as possible.

jeff-regier: @aakashdp6548 It would help if you'd merge/rebase this file so we can more easily understand the changes

case_studies/psf_variation/train_models.sh
```diff
@@ -1,27 +1,20 @@
...
 
-# train base model
-bliss -cp ~/bliss/case_studies/psf_variation -cn config mode=train \
-    train.trainer.logger.name=PSF_MODELS \
-    train.trainer.logger.version=base_model \
-    encoder.image_normalizer.concat_psf_params=False \
-    cached_simulator.cached_data_path=/data/scratch/aakash/train_single_94-1-12
+# Generate new data
+bash ~/bliss/scripts/generate_data_in_parallel.sh -n 32 -cp ~/bliss/case_studies/psf_variation/conf -cn psf_aware
+bash ~/bliss/scripts/generate_data_in_parallel.sh -n 32 -cp ~/bliss/case_studies/psf_variation/conf -cn single_field
```
jeff-regier, line 7: probably good to rename the script since it includes data generation now

jeff-regier: (which is great, thanks for the change. now it's fully reproducible)

case_studies/psf_variation/train_models.sh
```diff
@@ -1,27 +1,20 @@
...
+bliss -cp ~/bliss/case_studies/psf_variation/conf -cn psf_unaware mode=train
 
-# run the evaluation notebook
-jupyter nbconvert --execute evaluate_models.ipynb --to html
+# Train psf-aware model
+bliss -cp ~/bliss/case_studies/psf_variation/conf -cn psf_aware mode=train
+
+# Run evaluation and generate figures
+python evaluate_models.py --run_eval --plot_eval --run_calibration --plot_calibration --data_path=/data/scratch/aakash/multi_field
+python evaluate_models.py --run_eval --plot_eval --run_calibration --plot_calibration --data_path=/data/scratch/aakash/single_field
```
jeff-regier, line 20: Could you call the ipynb files in `case_studies/psf_variation` too? (Or if we no longer need those notebooks, could we remove them?)

jeff-regier: Oh, there's actually just one left: `find_param.bins.ipynb`.

case_studies/psf_variation/conf/config.yaml
```diff
@@ -0,0 +1,321 @@
...
+    galaxy_a_bd_ratio: 2.0
+
+surveys:
+    sdss:
+        _target_: bliss.surveys.sdss.SloanDigitalSkySurvey
+        dir_path: ${paths.sdss}
+        fields:
+            - run: 94
+              camcol: 1
+              fields: ${range:12,482,20}
```
jeff-regier, line 58: Would you change `base_config.yaml` to use multiple fields by default? (That will also cause pytest to use multiple fields by default, so maybe two is enough.) Going forward, because of the success of your project, we basically always want to train with multiple PSFs.

case_studies/psf_variation/train_models.sh
jeff-regier: Oh and I see: this notebooks doesn't do much except find the thresholds that are included elsewhere

bliss/encoder/metrics.py
aakashdp6548: Updated all the metrics to take `bin_cutoffs`, `ref_band`, and `bin_type`

bliss/encoder/metrics.py
aakashdp6548: Merged from master, should be up to date now

case_studies/psf_variation/train_models.sh
aakashdp6548: Yeah it's just finding magnitude quantiles to use for binning in metrics and the calibration plots. I kept it in for reference but it's not really relevant to the main results

case_studies/psf_variation/train_models.sh
aakashdp6548: Renamed to `run_experiments.sh`

bliss/encoder/metrics.py
```diff
@@ -501,59 +501,203 @@ def get_internal_states(self):
...
+        self.add_state(
+            "flux_abs_pct_err",
+            default=torch.zeros((len(self.survey_bands), self.n_bins)),  # n_bins per band
+            dist_reduce_fx="sum",
+        )
+        self.add_state("n_matches", default=torch.zeros(self.n_bins), dist_reduce_fx="sum")
 
     def update(self, true_cat, est_cat, matching):
+        cutoffs = torch.tensor(self.bin_cutoffs, device=self.device)
+        true_bin_measures = true_cat.on_fluxes(self.bin_type)[:, :, self.ref_band].contiguous()
```
jeff-regier, line 534: Do we need a `ref_band` in this error metric? Wouldn't we just compute flux error for whatever bands are available? @aakashdp6548 Because you're using the `OneBandTransform`, I think your data will only have one band.

aakashdp6548: This does compute flux error for each band, but we still need a reference band to choose which band to bin by. We could possibly bin separately for each band, but then we'd need multiple `bin_cutoffs` instead, which imo is more functionality than is needed right now.

jeff-regier: Got it, that makes sense

case_studies/psf_variation/conf/config.yaml
aakashdp6548: I added `3900/6/269` as we have it saved in `tests/data/sdss`.

jeff-regier: Great!

jeff-regier (approved):
Thanks Aakash!

PR #1056, opened 2024-08-09
Refactor metrics and to_tile_catalog

YicunDuanUMich:
* optimize `to_tile_catalog`: achieve 30x speedup (on a full catalog of 4000x4000 image, the processing time decreases from about 21s to 0.7s)
* refactor metrics to be agnostic to flux unit
* update notebooks
* delete old notebooks


YicunDuanUMich:
The test coverage decreases because the test doesn't go throught some functions in metrics that are designed for notebook. This is not a significant issue.

jeff-regier (approved):
Nice!

PR #1057, opened 2024-08-10
Batched to_tile_catalog

YicunDuanUMich:
* remove the `for` loop over `batch_size` in `to_tile_catalog`
* about 70x speedup for prcessing batch of size 64 (time decreases from about 0.7s to 0.01s)


jeff-regier (approved):
Excellent

PR #1060, opened 2024-08-19
Fix flux unit

YicunDuanUMich:
* the flux unit defined in `catalog.py` is problematic; fix it in this PR
* add some comments to `to_tile_catalog`
* consider the cases where corner points in `to_tile_catalog` have invalid indices


bliss/encoder/metrics.py
```diff
@@ -138,36 +138,31 @@ def get_report_bins(self):
...
+        self.base_njy_bin_cutoffs = self.bin_cutoffs
         self.report_bin_unit = report_bin_unit
-        assert self.report_bin_unit in {"mag", "nmgy", "njymag"}, "invalid bin type"
+        assert self.report_bin_unit in {"njy", "ab_mag"}, "invalid bin type"
 
     def get_state_for_report(self, state_name):
         state = getattr(self, state_name)
         match self.report_bin_unit:
-            case "nmgy":
+            case "njy":
```
jeff-regier, line 151: I don't use njy for SDSS, only nmgy

jeff-regier: I wish there was a generic term that could mean either

bliss/catalog.py
```diff
@@ -10,32 +10,33 @@
...
+        flux: the flux in any linear unit
+        c: you may change this according to your survey dataset;
 
     Returns:
-        Tensor indicating fluxes in AB magnitude
+        Tensor indicating fluxes in magnitude
     """
 
-    return 22.5 - 2.5 * torch.log10(nmgy / 3631)
+    return 22.5 - 2.5 * torch.log10(flux / c)
```
jeff-regier, line 39: Better to subtract log10(c) for stability

bliss/conf/base_config.yaml
```diff
@@ -143,17 +143,17 @@ variational_factors:
 metrics:
     detection_performance:
         _target_: bliss.encoder.metrics.DetectionPerformance
-        base_nmgy_bin_cutoffs: [200, 400, 600, 800, 1000]
-        report_bin_unit: njymag
+        base_njy_bin_cutoffs: [200, 400, 600, 800, 1000]
```
jeff-regier, line 146: Seems better to always specify cutoffs in magnitude, and to always report them in the same units

jeff-regier (changes requested):

bliss/catalog.py
```diff
@@ -451,33 +436,17 @@ def galaxy_bools(self) -> Tensor:
...
 
-    @property
-    def on_mag(self) -> Tensor:
-        return convert_nmgy_to_mag(self.on_nmgy)
-
-    @property
-    def on_njymag(self) -> Tensor:
-        return convert_nmgy_to_njymag(self.on_nmgy)
+    def on_magnitudes(self, c) -> Tensor:
+        return convert_flux_to_magnitude(self.on_fluxes, c)
```
jeff-regier, line 449: Where is `c` defined? I believe `c` is survey specific.

bliss/conf/base_config.yaml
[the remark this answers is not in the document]
jeff-regier: Alternatively, we could call these `flux_bin_cutoffs`, and use the default units for the survey (without specifying what they are). I'm not sure which way is better.

bliss/encoder/metrics.py
```diff
@@ -138,36 +138,31 @@ def get_report_bins(self):
...
                 pass
-            case "mag":
-                report_bins = convert_nmgy_to_mag(report_bins)
-                report_bins = torch.flip(report_bins, dims=(-1,))
-            case "njymag":
-                report_bins = convert_nmgy_to_njymag(report_bins)
+            case "ab_mag":
```
jeff-regier, line 164: Maybe call this "mag" rather than "ab_mag"? Apparently SDSS magnitudes aren't exactly ab magnitudes

bliss/encoder/metrics.py
YicunDuanUMich: change it to `flux`

bliss/catalog.py
YicunDuanUMich: remove the 22.5 and rename `c` to `zero_point`

bliss/catalog.py
YicunDuanUMich: I tried to modify the `Survey` class but find that we also need to change the `Encoder` and some notebooks. So I move the zero point specification to our metrics by adding one argument `mag_zero_point` to our metric classes.

bliss/conf/base_config.yaml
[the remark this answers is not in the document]
YicunDuanUMich: change it to `base_flux_bin_cutoffs`

bliss/encoder/metrics.py
YicunDuanUMich: change it to `mag`

jeff-regier (approved):
Looks good

PR #1062, opened 2024-08-20
Weak lensing DC2 catalog generation updates

timwhite0:
This PR contributes scripts and notebooks for generating an alternative version of our DC2 weak lensing catalog that includes all galaxies from the truth table with an r-band flux ≥ 50. This is a different approach from a previous PR in which we included all galaxies from the object-with-truth-match table (their true properties, not their measured properties). Both approaches are reasonable, and the resulting catalogs are not very different. The new, alternative version has ~1 million additional galaxies that were not in the object-with-truth-match table (e.g., because they were faint or blended).


jeff-regier (approved):
50 nJy may be too faint to allow for useful ellipticity estimates

PR #1067, opened 2024-08-26
Weak lensing DC2 updates

timwhite0:
- Updated `load_image_and_catalog` and `generate_cached_data`. More efficient splitting of cached .pt files, and also more similar to the parent DC2 class than it was before.
- New baseline shear estimator that uses `conv2d` to compute a weighted average of ellipticities from neighboring tiles. The previous approach, which was to estimate shear in each tile by taking the average of ellipticities in that tile, can be recovered by setting the size of the kernel in conv2d to 1. After a bit of exploration, it seems like a kernel size of 5 or 7 provides the best approximation of shear. This likely has implications for the "optimal" receptive field of our lensing encoder.
- Renamed and reorganized some notebooks and other files.
- Updated catalog so that it only includes ~3.4 million galaxies with r-band flux ≥ 200 (r-band mag ≤ ~26). Previously our catalog contained ~10-11 million galaxies with r-band flux ≥ 50. With the new catalog, average ellipticity is not quite as good an estimator of shear as it was before, and definitely not as good as it was when we were (erroneously) including all ~128 million CosmoDC2 galaxies. See the bottom of `ellipticity.ipynb` for an example.


jeff-regier (approved):
Looks good. Please try to remove the duplicated code before merging.

PR #1072, opened 2024-09-17
weak lensing encoder shear updates

shreyasc30:
This PR makes the following weak-lensing and DC2-related changes:
- allows users of the dc2 class to pull the entire set of PSF parameters for the truth match galaxies while maintaining flexibility to just use the median
- subsequent updates to lensing dc2 to support the above workflow and expose PSFs in the tile catalog for use in image normalizers
- Novel ResNet-like network structure for weak lensing, along with ResNet and ResNeXt blocks
- Notably, the ability to have a residual connection via 1x1 convolution even when changing dimensions, which previous blocks did not have
- conversion of shear_1 and shear_2 to Normal Factors and temporary removal of convergence from being predicted
- subsequent updates to the encoder, plots, metrics, and config files to support the above


bliss/cached_dataset.py
shreyasc30, line 1: unfortunately, this needs to be wrapped in a try/except block for the case where converted_sub_index is incorrect (i.e. negative) due to a misunderstanding in sizes and our use of a single file per image as opposed to multiple. Despite adding the _size directive as requested, this error still propagates silently. I can make a more elegant fix if necessary but this gets the job done with minimal impact to others' work

jeff-regier: I'm concerned about suppressing this key error silently. Can the error be avoided upstream somehow? Why would `converted_sub_index` be negative?

jeff-regier (changes requested):
Looks good otherwise

bliss/cached_dataset.py
shreyasc30: in our case, every split file corresponds to a single image - hence self.file_paths[converted_index] is correctly read in as the data. Unfortunately, converted_sub_index is set to (index - file_size[converted_index]) - so when index=0 (i.e. the first file) and our files are of size 1, converted_sub_index takes on a value of -1. In our case, we actually just want to load in the data at self.file_paths[converted_index] as it represents a single image. Have been trying to solve this upstream but to no avail thus far

shreyasc30: perhaps I'm misunderstanding something about how this method is written / what the point of converted_sub_index is altogether

shreyasc30: I agree, there should be a better way to do this that doesn't allow silent errors to propagate, but it's unclear to me exactly how to fix the issue. Will keep investigating for now

shreyasc30: turns out I was overthinking the solution, solved in 0ffa3126de19229cf605d68155bbcb163e6efb80 (simple fix)

jeff-regier (approved):
Thanks for the revision

PR #1078, opened 2024-10-21
Add JIF results; Estimate ellipticity using second moments

YicunDuanUMich:
* update the code and notebook according to the master branch
* add JIF results: you can find the results in `jif_related` folder
* estimate ellipticity using second moments


jeff-regier (approved):
Nice

PR #1079, opened 2025-01-07
Weak lensing: updated baseline shear estimator + manuscript figures

timwhite0:
- Regenerated catalog after DC2 files were re-synced. Added `ellipticity_1_true_dc2` and `ellipticity_2_true_dc2` to the catalog.
- Estimated galaxy ellipticities using LSST-measured second moments `Ixx_pixel`, `Ixy_pixel`, `Iyy_pixel`. Switched to using these LSST estimates in our baseline shear estimator instead of the true ellipticities.
- Updated tuning procedure for the baseline shear estimator. We now use only the training set to tune (I was previously using all DC2 images by mistake), and we use R-squared to select the optimal weighted-average kernel size and standard deviation.
- Created some figures for the manuscript.

See `evaluate_encoder.ipynb` and `manuscript/figure*.ipynb` for recent results. Here's a summary from a fairly recent training run:
```
BLISS:
test loss: -10.34
shear 1 test MSE: 5.8e-05
shear 2 test MSE: 5.9e-05
convergence test MSE: 6.5e-05

Baseline estimator based on LSST ellipticity estimates:
shear 1 test MSE: 8.5e-05 
shear 2 test MSE: 6.5e-05
```


jeff-regier (approved):
Looks good!

PR #1089, opened 2025-06-06
Weak lensing: descwl, DC2 redshift bins, refactored encoder

timwhite0:
- Script to convert images and catalogs from descwl-shear-sims to BLISS input. Configs to train on the resulting images.
- Reorganized the weak lensing case study directory; separate subdirectories for DC2, descwl, and preliminary implementation of a BLISS weak lensing simulator.
- Rewrote the weak lensing subclass of surveys.dc2 to bin shear and convergence maps by redshift bins. Updated weak lensing plots and metrics to accommodate this change. We find that our encoder can infer shear and convergence at 3 redshift bins and at 5 redshift bins. In a future PR, we will create/update results notebooks to demonstrate this capability.
- Refactored our encoder. Combined WeakLensingFeaturesNet and WeakLensingCatalogNet into WeakLensingNet and added a few extra ResNet blocks. As before, the encoder comprises a sequence of many ResNet blocks, and each ResNet block comprises Conv2d, GroupNorm, and SiLU layers.
- Fixed preprocessing of PSF parameters in the weak lensing subclass of surveys.dc2.

### DIFF
--- a/bliss/encoder/image_normalizer.py
+++ b/bliss/encoder/image_normalizer.py
@@ -99,7 +99,4 @@ class NullNormalizer(torch.nn.Module):
         return 1
 
     def get_input_tensor(self, batch):
-        return rearrange(
-            torch.nan_to_num(batch["images"], nan=batch["images"].nanmedian()),
-            "b bands h w -> b bands 1 h w",
-        )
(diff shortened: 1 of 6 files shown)


jeff-regier (approved):
Please merge when ready

